PR #1, opened 2025-03-10
Open-source release of GREASE

RyanGlScott:
GREASE is a CLI tool that checks properties about binaries and LLVM bitcode
using under-constrained symbolic execution.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -19,7 +19,11 @@ jobs:
       fail-fast: false
       matrix:
         # Cover both x86-64 and ARM64.
-        os: [ubuntu-22.04, ubuntu-22.04-arm]
+        #
+        # For now, we disable the ARM64 job, as ARM64 Ubuntu runners are only
+        # available in public repos. Once we make this repo public, we can
+        # re-enable it. (#55)
+        os: [ubuntu-22.04] #, ubuntu-22.04-arm]
     steps:
       - uses: actions/checkout@v4
         with:


elf-edit-core-dump/LICENSE
```diff
@@ -0,0 +1,30 @@
...
+
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+
+    * Redistributions in binary form must reproduce the above
+      copyright notice, this list of conditions and the following
+      disclaimer in the documentation and/or other materials provided
+      with the distribution.
+
+    * Neither the name of Ryan Scott nor the names of other
```
thebendavis: update for consistency with rest of files

```suggestion
    * Neither the name of Galois, Inc. nor the names of other
```

README.md
```diff
@@ -1 +1,17 @@
...
\ No newline at end of file
+# grease
+
+`grease` is a CLI tool that checks properties about binaries using under-constrained symbolic execution.
+
+For more information, see [the documentation](./doc).
+
+## Acknowledgements
+
+This material is based upon work supported by the Defense Advanced Research Projects Agency under Contract No. W31P4Q-22-C-0017 and W31P4Q-23-C-0020
```
thebendavis:
```suggestion
This material is based upon work supported by the Defense Advanced Research Projects Agency under Contract No. W31P4Q-22-C-0017 and W31P4Q-23-C-0020.
```

.github/ci.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
langston-barrett, line 1: Any reason in particular not to just inline `install_system_deps` into `ci.yaml`? Doing so would make it easier to understand `SOLVER_PKG_VERSION`.

Installing SMT solvers from `what4-solvers` seems like a good candidate for another Galois-specific action in [GaloisInc/.github](https://github.com/GaloisInc/.github).

.github/ci.sh
```diff
@@ -0,0 +1,24 @@
...
+set -xEeuo pipefail
+
+[[ "$RUNNER_OS" == 'Windows' ]] && IS_WIN=true || IS_WIN=false
+BIN=${PWD}/bin
+EXT=""
+$IS_WIN && EXT=".exe"
```
langston-barrett: Perhaps:
```suggestion
EXT=""
if [[ "$RUNNER_OS" == 'Windows' ]]; then
  EXT=".exe"
fi
BIN=${PWD}/bin
```

.github/ci.sh
```diff
@@ -0,0 +1,24 @@
...
+mkdir -p "$BIN"
```
langston-barrett: Tiny nit, but this seems to be the only effectful line in the top-level of this script - maybe move it into `install_system_deps`?

.github/ci.sh
```diff
@@ -0,0 +1,24 @@
...
+
+is_exe() { [[ -x "$1/$2$EXT" ]] || command -v "$2" > /dev/null 2>&1; }
+
+install_system_deps() {
+  (cd $BIN && curl -o bins.zip -sL "https://github.com/GaloisInc/what4-solvers/releases/download/$SOLVER_PKG_VERSION/$BIN_ZIP_FILE" && unzip -o bins.zip && rm bins.zip)
```
langston-barrett: It's a bit confusing to have all this on one line. Maybe:
```suggestion
  pushd $BIN
  curl -o bins.zip -fsSL "https://github.com/GaloisInc/what4-solvers/releases/download/$SOLVER_PKG_VERSION/$BIN_ZIP_FILE"
  unzip -o bins.zip
  rm bins.zip
  popd

```
(Also added some `curl` flags for better behavior on error)

.github/workflows/ci.yaml
```diff
@@ -6,17 +6,101 @@ on:
     - main
   pull_request:
 
+env:
+  # The solver package snapshot date. If you update this, make sure to also
+  # update it in the Dockerfile.
+  SOLVER_PKG_VERSION: "snapshot-20241119"
```
langston-barrett: Could this be local to the job that uses it, like `BIN_ZIP_FILE`? That would make it clear that it isn't used in other steps

.github/workflows/ci.yaml
```diff
@@ -6,17 +6,101 @@ on:
...
+  # a valid version of the "happy" tool).
+  CACHE_VERSION: 1
+
 jobs:
   build:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
+        ghc: ["9.4.8", "9.6.6", "9.8.4"]
+        cabal: [ '3.10.3.0' ]
```
langston-barrett: Tiny nit: why the defferent quoting/list styles?
```suggestion
        ghc: ["9.4.8", "9.6.6", "9.8.4"]
        cabal: ["3.10.3.0"]
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,54 @@
...
+      pages: write
+      id-token: write
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: false
+        fetch-depth: 0
```
langston-barrett: The docs say:

> Set fetch-depth: 0 to fetch all history for all branches and tags.

I'd be surprised if this is necessary for the docs.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  lint:
+    runs-on: ubuntu-22.04
+    name: lint
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: false
+          fetch-depth: 0
```
langston-barrett: Same q: re `fetch-depth`

.github/workflows/lint.yml
```diff
@@ -0,0 +1,25 @@
...
+
+      - shell: bash
+        run: |
+          curl --location -o hlint.tar.gz \
```
langston-barrett: Consider the standard battery of "do it more reasonably" `curl` flags
```suggestion
          curl -fsS --location -o hlint.tar.gz \
```

Dockerfile
```diff
@@ -0,0 +1,104 @@
...
+ENV LANGUAGE=en_US:en
+ENV LC_ALL=en_US.UTF-8
+
+ARG TARGETPLATFORM
+# If you update the values below, make sure to also update them in
+# .github/workflows/ci.yaml.
+ENV WHAT4_SOLVERS_VER=snapshot-20241119
+ENV GHCUP_VER=0.1.40.0
+ENV CABAL_VER=3.10.3.0
+ENV GHC_VER=9.4.8
```
langston-barrett: Perhaps we should be using GHC 9.8 here?

README.md
```diff
@@ -1 +1,17 @@
-# GREASE
\ No newline at end of file
+# grease
+
+`grease` is a CLI tool that checks properties about binaries using under-constrained symbolic execution.
+
+For more information, see [the documentation](./doc).
```
langston-barrett: Consider linking to Github pages

.github/workflows/ci.yaml
RyanGlScott: Hah, this is copy-pasta from a different project—no deeper reason that that. I've switched to use the same style throughout.

.github/ci.sh
RyanGlScott: Sorry. More copy-pasta.

.github/ci.sh
RyanGlScott: More copy-pasta. (In the script where I copied this from, `$BIN` is referenced from multiple functions, so it made a tiny bit more sense to do this upfront. Not so much here, where `$BIN` is only used in a single function.)

.github/ci.sh
RyanGlScott: Mmm, delicious copy-pasta.

.github/ci.sh
RyanGlScott: I have a slight preference towards keeping this in a script, as reproducing what a given step in a CI workflow does locally becomes annoying when the step has more than, say, two or three lines of bash (as is the case with `install_system_deps`). I can say from experience on other projects that I have needed to run `install_system_deps` locally to debug things, so putting it in a script definitely paid its weight then.

I agree that we probably should move this logic (and lots of other logic that is shared among all of our CI workflows) into `GaloisInc/.github`. Are you OK if I defer this to an issue?

.github/workflows/ci.yaml
RyanGlScott: Yes, I'll do that. (In the workflow I copied this from, `SOLVER_PKG_VERSION` was referenced in multiple locations, but not so here.)

.github/workflows/docs.yml
RyanGlScott: Yikes, I completely misunderstood what `fetch-depth` did. I _meant_ to say `fetch-depth: 1` here—I'll change this.

.github/workflows/lint.yml
RyanGlScott: Similarly, this was a misunderstanding on my part.

.github/workflows/lint.yml
RyanGlScott: Good idea.

Dockerfile
RyanGlScott: No strong opinions here. We have to pick _some_ version in our support window, and GHC 9.8 is indeed one such version.

README.md
RyanGlScott: Good idea. I've replaced the `doc` link with https://galoisinc.github.io/grease.

.github/ci.sh
langston-barrett: > I have a slight preference towards keeping this in a script, as reproducing what a given step in a CI workflow does locally becomes annoying when the step has more than, say, two or three lines of bash (as is the case with install_system_deps). I can say from experience on other projects that I have needed to run install_system_deps locally to debug things, so putting it in a script definitely paid its weight then.

Fair, I trust your experience here.

> Are you OK if I defer this to an issue?

Yes of course! I don't even think we necessarily need a GREASE issue for this. Sorry, didn't mean to suggest that you do it right now.

RyanGlScott: I meant "issue" as in a [GaloisInc/.github](https://github.com/GaloisInc/.github) issue.

langston-barrett: Oh! Then yes, let's :)

RyanGlScott: I've opened https://github.com/GaloisInc/.github/issues/9 for this.

Issue #55, opened 2025-03-12
CI: Re-enable `ubuntu-22.04-arm` Docker job


RyanGlScott:
We aspire to have both x86-64 and ARM64 Docker images built in CI. Unfortunately, the only GitHub Actions runners that can run ARM64 at the moment are its ARM64 Ubuntu runners, which are still in a public beta. This means that we cannot run CI jobs on these runners while the `grease` repo is still private.

For now, I have disabled the part of the Docker CI configuration that runs a job on a  `ubuntu-22.04-arm` runner for the sake of getting the CI to pass. When we flip the repo from private to public, we should re-enable this job.

PR #63, opened 2025-03-21
Revert "CI: Disable ubuntu-22.04-arm Docker job for now"

RyanGlScott:
Now that the `grease` repo is public, we can use Ubuntu ARM runners on GitHub Actions (which are still in public beta as of the time of writing). Fixes https://github.com/GaloisInc/grease/issues/55.


langston-barrett (approved):
LGTM if CI approves

Issue #22, opened 2025-03-11
Handle dynamic relocations that reference code in other shared libraries (e.g., `COPY` and `GLOB_DAT`)


RyanGlScott:
Several dynamic relocation types (`COPY`, `GLOB_DAT`, etc.) involve referencing data found in a shared library. `ambient-verifier` handles these, but in order to use the same approach as `ambient-verifier`, we'd first need to support simulating shared libraries (#21). Perhaps we don't need to do all of #21 first, however. As a workaround, we could treat a read from one of these relocation types similarly to how we treat a read from a mutable global (whose behavior is controlled by the `--globals` flag).

There are also dynamic relocation types that reference data within the _same_ binary (e.g., `ABS32` and `ABS64`), but these would likely be easier to support, given that the corresponding data should always be relatively straightforward to look up.

RyanGlScott:
Another way that this can arise is with `R_X86_64_64` relocations. Here is a small example of a program that will emit such relocations:

```c
#include <stddef.h>
#include <stdlib.h>

struct foo {
  void* (*my_malloc)(size_t);
  void (*my_free)(void*);
};

static struct foo f = { malloc, free };

int main(void) {
  int* x = f.my_malloc(sizeof(int));
  *x = 0;
  int y = *x;
  f.my_free(x);
  return y;
}
```
```
$ clang test.c -o test.exe
$ objdump -R test.exe 

test.exe:     file format elf64-x86-64

DYNAMIC RELOCATION RECORDS
OFFSET           TYPE              VALUE 
0000000000003e18 R_X86_64_RELATIVE  *ABS*+0x0000000000001120
0000000000003e20 R_X86_64_RELATIVE  *ABS*+0x00000000000010e0
0000000000004020 R_X86_64_RELATIVE  *ABS*+0x0000000000004020
0000000000003fd8 R_X86_64_GLOB_DAT  __libc_start_main@GLIBC_2.34
0000000000003fe0 R_X86_64_GLOB_DAT  _ITM_deregisterTMCloneTable@Base
0000000000003fe8 R_X86_64_GLOB_DAT  __gmon_start__@Base
0000000000003ff0 R_X86_64_GLOB_DAT  _ITM_registerTMCloneTable@Base
0000000000003ff8 R_X86_64_GLOB_DAT  __cxa_finalize@GLIBC_2.2.5
0000000000004028 R_X86_64_64       malloc@GLIBC_2.2.5
[...]

PR #68, opened 2025-03-24
Support more `SymbolReloc`s (e.g., `R_X86_64_64`)

RyanGlScott:
In addition to `GLOB_DAT` relocs, all of our supported architectures contain at least one other symbol-related relocation (e.g., `R_X86_64_64` on x86-64). What's more, some programs may actually call into these relocations directly under certain circumstances, such as in this program:

```c

struct foo {
  void* (*my_malloc)(size_t);
  void (*my_free)(void*);
};

static struct foo f = { malloc, free };

int main(void) {
  int* x = f.my_malloc(sizeof(int));
  *x = 0;
  int y = *x;
  f.my_free(x);
  return y;
}
```

(Annoyingly, the `musl`-based compiled that the test suite's Nix flake provides do not exhibit the same behavior, so it is difficult to add a test case for this property.)

This patch adds `R_X86_64_64` _et al._ as `SymbolReloc`s such that `grease` will be able to characterize them as external function calls when necessary. As a result, this patch resolves one part of https://github.com/GaloisInc/grease/issues/22.

### DIFF
--- a/src/Grease/Macaw.hs
+++ b/src/Grease/Macaw.hs
@@ -26,9 +26,10 @@ import Control.Lens ((^.), to)
 import Control.Monad (return)
 import Control.Monad.IO.Class (MonadIO(..))
 import Data.Bool (otherwise)
+import qualified Data.ByteString.Builder as Builder
+import qualified Data.ByteString.Lazy as BSL
 import Data.Function (($), (.))
-import Data.Functor ((<$>))
-import Data.Int (Int)
+import Data.Int (Int, Int64)
 import Data.Kind (Type)
 import Data.List ((++))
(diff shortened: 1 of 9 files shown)


langston-barrett (approved):
Nice :+1:

Just want to tag https://github.com/GaloisInc/macaw/issues/326 here: We are currently going the `ambient-verifier` route and creating our own custom architecture-neutral abstraction over relocations that is more fine-grained than Macaw's. We should continue to consider whether this is the right abstraction (and so, should presumably be upstreamed into Macaw), or whether there *is no* right abstraction (and we should eventually migrate to having each architecture handle its own relocations, though likely with some shared code for common categories).

Issue #29, opened 2025-03-11
Add golden output files for each test case


RyanGlScott:
Currently, each test case passes or fails on the basis of a particular check (e.g., not triggering the `no-mprotect` requirement, not causing a heuristics failure, etc.) However, the test suite doesn't check _why_ each test case fulfills its check. For example, it's possible that a test case might trigger a `grease` bug for one reason in one commit, but then trigger an entirely different bug in another commit. With the way the test suite is currently set up, however, we wouldn't be notified of this change. The test suite would simply see that a bug was triggered, and it would be none the wiser.

In pursuit of being able to track these sorts of changes, I think it would be nice if each test case had accompanying golden files that contain the expected output when running `grease` on that test case. (At the moment, `grease`'s output is the most reliable metric we have for tracking changes in its behavior.) This would mean that we would need to ensure that each test case gives deterministic output (e.g., no timestamps or nondeterministic random values).

langston-barrett:
Let's close this in favor of #79, #80, and #81. Porting tests to FileCheck addresses the concern of the specificity of their assertions (the "why"), while remaining insensitive to unrelated, innocuous changes in GREASE's output. 

Feel free to reopen if you disagree!

Issue #19, opened 2025-03-11
UX tests


RyanGlScott:
@langston-barrett says:

-----

It would be helpful to have UX tests that compare GREASE's log output to a file. This probably wouldn't be too big of a lift, because much of GREASE is already parameterized by a "log action" that could be used to capture said output.

langston-barrett:
Now that all of our tests are ported to Oughta, this should be quite easy to accomplish. We should create a new directory `tests/ux` and stash such tests in there.

PR #71, opened 2025-03-28
FileCheck: A library for testing programs that output text

langston-barrett:
The included README (and linked LLVM docs) might be a good place to reviewing.

Towards #19 and #29.

Opening a draft PR in hopes of getting feedback on the overall approach. It's obviously not totally complete, but I figured it's worth someone taking a look before I convert a bunch of tests.


.github/workflows/ci.yaml
```diff
@@ -72,6 +72,14 @@ jobs:
       shell: bash
       run: cabal configure --enable-tests
 
+    - name: Build FileCheck
+      shell: bash
+      run: cabal build pkg:filecheck
+
+    - name: Test FileCheck
+      shell: bash
+      run: cabal test pkg:filecheck
```
RyanGlScott, line 82: We might as well `cabal check` this library and run it through `hlint` too.

cabal.project
```diff
@@ -1,4 +1,5 @@
 packages:
+  filecheck/filecheck.cabal
```
RyanGlScott: A minor thing, but I'd prefer to list `grease`-related libraries first in this `cabal.project` file, since they're the main focus of the repo.

filecheck/src/FileCheck.hs
```diff
@@ -0,0 +1,117 @@
...
+import FileCheck.Command qualified as FCC
+import FileCheck.Directive qualified as FCD
+import FileCheck.Directive (Prefix)
+import GHC.Stack (HasCallStack)
+import Prelude hiding (lines)
+
+-- | Output of the program under test
+newtype Output = Output Text
+
+data FileCheckFailure = FileCheckFailure Command Text
```
RyanGlScott: Haddocks?

filecheck/src/FileCheck.hs
```diff
@@ -0,0 +1,117 @@
...
+      ]
+
+instance X.Exception FileCheckFailure
+
+-- | Match text against a sequence of 'Command's.
+check ::
+  HasCallStack =>
+  [Command] ->
+  Output ->
+  Either FileCheckFailure ()
```
RyanGlScott: This function doesn't throw any exceptions. Does it really need a `HasCallStack` constraint?

filecheck/src/FileCheck.hs
```diff
@@ -0,0 +1,117 @@
...
+-- Throws 'CommandParseFailure' if the commands cannot be parsed.
+parseCommentsAndCheck ::
+  HasCallStack =>
+  Maybe Prefix ->
+  -- | Start of line comment
+  Text ->
+  -- | Text containing comments with embedded commands
+  Text ->
+  Output ->
+  Either FileCheckFailure ()
```
RyanGlScott: Similar questions about the `HasCallStack` constraints on these two functions.

filecheck/src/FileCheck/Command.hs
```diff
@@ -0,0 +1,74 @@
...
+import Data.Text qualified as Text
+import Data.Text (Text)
+import FileCheck.Directive (Directive, Prefix)
+import FileCheck.Directive qualified as FCD
+
+data Command
+  = Command
+    { cmdDirective :: Directive
+    , cmdContent :: Text
+    }
```
RyanGlScott: Haddocks?

filecheck/src/FileCheck/Command.hs
```diff
@@ -0,0 +1,74 @@
...
+
+splitOnOne :: Text -> Text -> (Text, Text)
+splitOnOne needle haystack =
+  let (hd, tl) = Text.breakOn needle haystack in
+  (hd, Text.drop (Text.length needle) tl)
+
+
+parse :: Maybe Prefix -> Text -> Maybe Command
```
RyanGlScott: Haddocks?

filecheck/src/FileCheck/Command.hs
```diff
@@ -0,0 +1,74 @@
...
+  let split = ":"
+  let (hd, tl) = splitOnOne split t
+  d <- FCD.parse pfx hd
+  let tl' = Maybe.fromMaybe tl (Text.stripPrefix " " tl)
+  pure (Command d tl')
+
+-- | Match a 'Command' against a chunk of 'Text'.
+--
+-- Returns 'Just' the remaining text after the match, or 'Nothing' if matching
+-- fails.
```
RyanGlScott: Does it have to be an exact match, or can you match wildcard patterns? Worth clarifying somewhere (either in these Haddocks or elsewhere in the documentation).

filecheck/src/FileCheck/Directive.hs
```diff
@@ -0,0 +1,50 @@
...
+  , Directive(..)
+  , parse
+  , print
+  ) where
+
+import Data.Text (Text)
+import qualified Data.Maybe as Maybe
+import Prelude hiding (print)
+
+-- | @-check-prefix@
```
RyanGlScott: I'm not sure what this Haddock comment means.

filecheck/src/FileCheck/Directive.hs
```diff
@@ -0,0 +1,50 @@
...
+data Directive
+  = -- | @CHECK@
+    Check
+  | -- | @CHECK-EMPTY@
+    CheckEmpty
+  | -- | @CHECK-NEXT@
+    CheckNext
+  | -- | @CHECK-NOT@
+    CheckNot
+  | -- | @COM@ (comment)
```
RyanGlScott, line 29: Could you elaborate a bit more on what each of these directives checks for? For instance, I'm not entirely sure what `CHECK-EMPTY` means. Does it require the program not to have any output at all? At least one empty line? Something else?

filecheck/src/FileCheck/Directive.hs
```diff
@@ -0,0 +1,50 @@
...
+     | t == pfx <> "-NEXT" -> Just CheckNext
+     | t == pfx <> "-NOT" -> Just CheckNot
+     | t == "COM" -> Just Com
+     | otherwise -> Nothing
+
+parse :: Maybe Prefix -> Text -> Maybe Directive
+parse pfx = parseWithPrefix (Maybe.fromMaybe (Prefix "CHECK") pfx)
+
+print :: Directive -> Text
+print =
```
RyanGlScott: Haddocks?

filecheck/test/Test.hs
```diff
@@ -0,0 +1,41 @@
...
+discover pass dir = do
+  entries <- map (dir </>) <$> Dir.getDirectoryContents dir
+  files <- Monad.filterM Dir.doesFileExist entries
+  Monad.forM files $ \file -> do
+    content <- Text.IO.readFile file
+    pure $
+      TTH.testCase file $ do
+        let prefix = Nothing
+        let comment = "# "
+        let isComment = (comment `Text.isPrefixOf`)
```
RyanGlScott: Does this mean that you can't write `#CHECK` (i.e., that you need at least one space in between the `#` and the directive)?

filecheck/test/Test.hs
```diff
@@ -0,0 +1,41 @@
...
+import FileCheck (Output(..), parseCommentsAndCheck)
+import Prelude hiding (lines)
+import System.Directory qualified as Dir
+import System.FilePath ((</>))
+import Test.Tasty.HUnit qualified as TTH
+import Test.Tasty qualified as TT
+
+discover :: Bool -> FilePath -> IO [TT.TestTree]
+discover pass dir = do
+  entries <- map (dir </>) <$> Dir.getDirectoryContents dir
```
RyanGlScott: Consider using [`listDirectory`](https://hackage.haskell.org/package/directory-1.3.9.0/docs/System-Directory.html#v:listDirectory) here instead of `getDirectoryContents` so that you don't include the `.` and `..` entries, which are almost surely not going to be useful.

langston-barrett:
> Opening a draft PR in hopes of getting feedback on the overall approach.

@RyanGlScott and I discussed offline, and we are agreed that this is a good direction forwards.

filecheck/test/Test.hs
langston-barrett: Yeah, just for the test suite.

filecheck/src/FileCheck.hs
langston-barrett: Nope! Good call.

filecheck/src/FileCheck.hs
langston-barrett: These are both documented as throwing exceptions?

filecheck/src/FileCheck/Directive.hs
langston-barrett: I've noted that upstream docs should be consulted for the semantics (for now).

filecheck/src/FileCheck/Directive.hs
langston-barrett: I've noted that upstream docs should be consulted for the semantics (for now).

filecheck/src/FileCheck.hs
RyanGlScott: Oh! It is really strange that a function which returns `Either` (which usually indicates purity) can also throw an exception under the hood. Is there a way to remove the exception-throwing in favor of encoding the error in a `Left` value?

filecheck/src/FileCheck/Command.hs
langston-barrett: Right now it has to be an exact match. This is indirectly covered in the README:

> The simplest command is `CHECK: <str>`, which checks that the output contains `<str>`.

It's not incredibly explicit, but I think it's reasonable to assume that "contains `<str>`" does *not* indicate regex/wildcards/globs unless otherwise noted.

This will also become more explicit when regex support is added.

filecheck/src/FileCheck.hs
langston-barrett: This part has been considerably changed, these now return `Result`s instead of `Either`; they still throw exceptions.

langston-barrett:
Here is what the output looks like from Tasty on a failing test:
```
Tests
  LLVM
    abort:                                  FAIL
      Exception:
      Check failed! Passing checks:
      ✔️  CHECK at tests/llvm/abort.llvm.cbl:3:1-3:39 matched text at <out>:408:27-408:56
      pattern:
        Likely bug: unavoidable error
      match:
        Likely bug: unavoidable error



      Failing check:
      ❌ CHECK at tests/llvm/abort.llvm.cbl:6:1-6:27 did not match text at <out>:408:56
      pattern:
        Call to aborf
      text:
         (safety condition is unsatisfiable) at tests/llvm/abort.llvm.cbl:9:5
        tests/llvm/abort.llvm.cbl:9:5: error: in test
        Call to abort


      Use -p '/abort/&&/LLVM.abort/' to rerun this test only.
```

filecheck/src/FileCheck.hs
```diff
@@ -0,0 +1,205 @@
...
+-- | Failure to parse a t'Command'
+data CommandParseFailure = CommandParseFailure Text
+  deriving Show
+
+instance X.Exception CommandParseFailure
+
+-- | Parse t'Command's from comments embedded in a file, and use them to check
+-- an t'Output'.
+--
+-- Throws t'CommandParseFailure' if the commands cannot be parsed.
```
RyanGlScott: I tried looking to see where `CommandParseFailure` is thrown, but I couldn't find it. Am I missing something?

filecheck/src/FileCheck/Command.hs
```diff
@@ -0,0 +1,114 @@
...
+import FileCheck.Directive (Directive, Prefix)
+import FileCheck.Directive qualified as FCD
+import FileCheck.Pos qualified as FCP
+import FileCheck.Pos (Loc, Span)
+import Prelude hiding (span)
+
+-- | A t'Command' is a pair of a 'Directive' and its argument.
+--
+-- It represents a single line written by the user, e.g., @CHECK: foo@ is parsed
+-- to @v'Command' 'FCD.Check' "foo"@.
```
RyanGlScott: This `Command` value in the Haddocks has two fields, but the actual `Command` data constructor has three fields.

filecheck/src/FileCheck/Directive.hs
```diff
@@ -0,0 +1,54 @@
...
+     | t == "COM" -> Just Com
+     | otherwise -> Nothing
+
+-- | Parse a 'Directive'. Returns 'Nothing' on parse failure.
+parse :: Maybe Prefix -> Text -> Maybe Directive
+parse pfx = parseWithPrefix (Maybe.fromMaybe (Prefix "CHECK") pfx)
+
+-- | Print a 'Directive'. Conceptually the inverse of 'parse'.
+print :: Directive -> Text
+print =
```
RyanGlScott: Perhaps this should be named `printDirective` instead of just `print`? That would mirror other functions like `printResult`, `printLoc`, etc., and it would avoid clashing with `Prelude.print`.

filecheck/src/FileCheck/Pos.hs
```diff
@@ -0,0 +1,76 @@
...
+tshow = Text.pack . show
+
+-- | A source position
+data Pos
+  = Pos
+    { line :: {-# UNPACK #-} !Int
+    , col :: {-# UNPACK #-} !Int
+    }
+
+-- | Move a position forward past a of text.
```
RyanGlScott: I think there is a word missing here ("past a of text").

filecheck/test/Test.hs
```diff
@@ -0,0 +1,58 @@
...
+
+{-
+Test FileCheck, using FileCheck.
+
+The test suite reads text files from @test-data/{fail,pass}/@ and runs
+FileCheck on them /twice/. All of the commands are embedded in the text files,
+on lines that start with @#@. It first matches against the entire text file
+(specifically, the non-@#@ lines) using the unprefixed commands (e.g., @CHECK@).
+It then serializes the t'FC.Result' and matches /that/ against the commands
+using @v'FC.Prefix "OUT"@. This double-checking ensures that directives match
```
RyanGlScott:
```suggestion
using @v'FC.Prefix \"OUT\"@. This double-checking ensures that directives match
```

grease-cli/tests/Main.hs
```diff
@@ -457,52 +463,50 @@ makeSanityTests arch d =
...
-    , testCase "trap" "trap.llvm.cbl" $ assertSpecificBug Bug.MustFail Nothing
-    , testCase "user override" "user-override.llvm.cbl" assertSuccess
-    , testCase "user override defun" "user-override-defun.llvm.cbl" assertSuccess
-    , testCase "struct override" "struct-override-caller.llvm.cbl" assertSuccess
-    , testCase "declare in override" "declare-in-override/declare-in-override.llvm.cbl" assertSuccess
-    , testCase "startup override" "startup-override/test.llvm.cbl" assertSuccess
-    , testCase "Rust enum" "rust-enum.llvm.cbl" assertSuccess
-    ]
+llvmTests :: IO T.TestTree
+llvmTests = do
```
RyanGlScott, line 467: Do you think it is worth applying the same FileCheck treatment to the Macaw-based S-expression tests? (No need to do so as part of this PR, but it seems like a logical next step.)

grease-cli/tests/llvm/declare-in-override/declare-in-override.config
```diff
@@ -1 +0,0 @@
---overrides tests/llvm/declare-in-override/my-malloc.llvm.cbl
```
RyanGlScott, line 1: Why was this file moved?

grease-cli/tests/Main.hs
langston-barrett: Yes, definitely. We should probably do it for the tests involving bitcode and binaries too.

https://github.com/GaloisInc/grease/issues/79
#80 
#81

grease-cli/tests/llvm/declare-in-override/declare-in-override.config
langston-barrett: The `.config` file moved to be next to the `.llvm.cbl` file, which also moved. The `.llvm.cbl` file moved up a directory so that it was next to the other `.llvm.cbl` files, and so could be discovered by the `Dir.listDirectory` in the harness. Similar treatment was applied to the test in `startup-override/`.

filecheck/src/FileCheck.hs
langston-barrett: Indeed, this is outdated. Thanks for the catch!

RyanGlScott (approved):
Fantastic work! I am definitely looking forward to using this in the future.

langston-barrett:
Aw, thanks! I'm happy to hear that.

PR #77, opened 2025-04-02
`load{,Concrete}String`: Add an argument for maximum characters to read

### LINKED ISSUE #74: `load{,Concrete}String`: Allow specifying a maximum number of characters to read
The comments in `Grease.Macaw.Memory` say:

https://github.com/GaloisInc/grease/blob/db58869d7096da9814e18a3c8f8c4955ddea304b/grease/src/Grease/Macaw/Memory.hs#L10-L14

I now want to support a `strncpy`-like function, so I would like to perform this API change. I will take inspiration from how `ambient-verifier` does it, i.e., by adding [`Maybe Int`](https://github.com/GaloisInc/ambient-verifier/blob/eab04abb9750825a25ec0cbe0379add63f05f6c6/src/Ambient/Extensions.hs#L322-L323) arguments.

RyanGlScott:
Fixes #74.


grease/src/Grease/FunctionOverride.hs
```diff
@@ -371,7 +371,7 @@ callAssert bak mvar mmConf archCtx _pfn _pfile _pline ptxt = do
   let sym = C.backendGetSym bak
   st0 <- get
   let endian = Symbolic.fromCrucibleEndian (archCtx ^. archEndianness)
-  (txt, st1) <- liftIO $ loadConcreteString bak mvar mmConf endian ptxt st0
+  (txt, st1) <- liftIO $ loadConcreteString bak mvar mmConf endian ptxt Nothing st0
```
langston-barrett: No need to change this necessarily, but I sometimes like giving things names to make it clear what they represent
```suggestion
  let maxSize = Nothing
  (txt, st1) <- liftIO $ loadConcreteString bak mvar mmConf endian ptxt maxSize st0
```
An alternative would be to make a new `data` type, but that feels perhaps excessive in this case.

grease/src/Grease/Macaw/Memory.hs
```diff
@@ -3,19 +3,12 @@ Copyright        : (c) Galois, Inc. 2024
...
-
-* We check if the last character is a null terminator by consuling an SMT
-  solver. This is useful (and necessary) for some use cases, but it does require
-  adding 'OnlineSolverAndBackend' constraints to support it.
+about this functionality is @grease@-specific, but we make a somewhat
+opinionated design choice: when checking if the last character is a null
+terminator, we consult an SMT solver. This is useful (and necessary) for some
+use cases, but it does require adding 'OnlineSolverAndBackend' constraints to
+support it. This design choice warrants a closer look before being considered
+for inclusion upstream in @macaw@:
```
langston-barrett:
```suggestion
for inclusion upstream in @macaw@.
```

langston-barrett (approved):

PR #86, opened 2025-04-10
doc: Describe which C undefined behaviors GREASE supports

### LINKED ISSUE #85: doc: Describe which C undefined behaviors GREASE checks for
There is a big list of C11 UBs [here](https://wiki.sei.cmu.edu/confluence/display/c/CC.+Undefined+Behavior).

This is a subtle question, because GREASE analyzes machine code. Accordingly, many UBs are out-of-scope. Additionally, GREASE reports certain behaviors that are technically well-defined at the machine code level because it assumes a C-like memory model. We should discuss all of this further in the docs.

We should link to this discussion from existing places in the docs such as the following:

> GREASE operates on compiled code (binaries and LLVM). There may be bugs (e.g., undefined behavior) in the source program that the compiler has optimized away; GREASE has no means of finding such bugs.

(In limitations.md)

> GREASE features a [memory model] that is capable of detecting many types of common software-related issues, such as double-frees or uninitialized stack reads. 

(In requirements.md)

langston-barrett:
Progress towards #85.

### DIFF
--- a/doc/undefined-behavior.md
+++ b/doc/undefined-behavior.md
@@ -1,0 +1,67 @@
+# Undefined behavior
+
+GREASE aims to find or verify the absence of *bugs* in the program under analysis.
+It supports checking both for specific [requirements](requirements.md) and for generally suspicious behaviors.
+This page describes the latter.
+
+What constitutes a bug?
+When analyzing C, C++, or Rust code, a conservative answer is that any form of reachable undefined behavior in a complete program constitutes a bug.
+For machine code, things aren't so clear.
+
+Many behaviors that might be considered undesirable or undefined at the source level have well-defined semantics at the machine code level.
+For example, consider a C program with a stack buffer overflow into an adjacent variable.
+In the C abstract machine, such behavior is clearly undefined.
+However, if the program is compiled in a straightforward way into machine code, the distinction between different stack variables is lost.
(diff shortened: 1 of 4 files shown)


doc/undefined-behavior.md
```diff
@@ -0,0 +1,67 @@
...
+| 34 | A side effect on a scalar object is unsequenced relative to either a different side effect on the same scalar object or a value computation using the value of the same scalar object (6.5.1). | ? | |
+| 35 | An exceptional condition occurs during the evaluation of an expression (6.5.1). | ? | |
+| 37 | A function is defined with a type that is not compatible with the type (of the expression) pointed to by the expression that denotes the called function (6.5.3.3). | No | Types are not available at the binary level. |
+| 38 | A member of an atomic structure or union is accessed (6.5.3.4).	 | ? | |
+
+## Exceptions
+
+Certain operations generate traps or exceptions at the processor level. GREASE attempts to treat these as errors. These include:
+
+- Division by zero
```
langston-barrett, line 67: @RyanGlScott Is there anything else we should list here? Floating point exceptions, perhaps?

RyanGlScott: Macaw is not very well equipped to reason about properties of floating-point numbers, so I'm doubtful that `grease` would be able to find much (if anything) there.

RyanGlScott (approved):

PR #89, opened 2025-04-10
ci: Upgrade to the latest Cabal version

langston-barrett:
Just keeping up to date.


.github/workflows/ci.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         ghc: ["9.4.8", "9.6.6", "9.8.4"]
-        cabal: ["3.10.3.0"]
+        cabal: ["3.14.1.1"]
```
RyanGlScott, line 23: If we update this here, then we ought to update it in the Dockerfile as well.

langston-barrett: Good point! Done!

Issue #83, opened 2025-04-04
filecheck: Variables


langston-barrett:
Many of our FileCheck-based tests check for the same strings, e.g.
```
$ rg '; CHECK: All goals passed!' grease-cli/tests | wc -l

23
```
It would be nice if FileCheck had some support for DRYing up these assertions, in order to avoid mass updates when these messages change and also to support brevity in the tests themselves.

My proposal is to add *variables*, e.g.,
```
; CHECK: $ok
```
where the call to FileCheck's API would specify an environment mapping variable names to strings, perhaps:
```haskell
let env = FileCheck.Environment (Map.fromList [("ok", "All goals passed!")])
```
I believe that this would be similar to LLVM FileCheck's `-D` flag, but that flag has very vague docs so I can't be sure: https://llvm.org/docs/CommandGuide/FileCheck.html#cmdoption-FileCheck-D-VAR

langston-barrett:
Fixed by #92

PR #92, opened 2025-04-17
filecheck: Lua integration

### LINKED ISSUE #84: filecheck: Use Lua as a DSL
After writing #82 and #83 and reading some more LLVM FileCheck documentation, I've come (yet again) to the realization that any sufficiently expressive tool is best expressed as (or rather, *is*) a DSL.

LLVM's FileCheck has:

- [Conditionals](https://llvm.org/docs/CommandGuide/FileCheck.html#the-filecheck-check-prefix-option)
- [Comments](https://llvm.org/docs/CommandGuide/FileCheck.html#the-com-directive)
- [Negation](https://llvm.org/docs/CommandGuide/FileCheck.html#the-check-not-directive)
- [Iteration](https://llvm.org/docs/CommandGuide/FileCheck.html#the-check-count-directive)
- [Regular expressions](https://llvm.org/docs/CommandGuide/FileCheck.html#filecheck-regex-matching-syntax)
- [Variables](https://llvm.org/docs/CommandGuide/FileCheck.html#filecheck-string-substitution-blocks)
- [Numeric types and operations](https://llvm.org/docs/CommandGuide/FileCheck.html#filecheck-numeric-substitution-blocks)

That's a programming language, right there!

Two follow-up questions naturally arise:

- Is the DSL well-suited to the task at hand?

  - Is it easy to learn? Is it phrased in terms from the domain?
  - It is familiar, does it draw from existing languages that the relevant developers know?
  - Does it allow concise and readable expression of the relevant tasks?

- Is it well-suited *enough* to justify its existence? Would reusing a general-purpose language be easier, both in the implementation and for users?

Upon reflection, in my opinion, the answer to the above questions for LLVM FileCheck is "not really". Rather than try to (re-)design and (re-)implement common PL features like variables and regular expressions, it might be worth thinking about alternatives. Specifically, we might want to consider using Lua as a DSL, as it is battle-tested and easily embeddable.

What would the API look like? Consider the semantics of a FileCheck program. In Haskell parlance, it is basically a function that consumes `Text`, and produces:

1. A list of matches (which consist of the span of text that was matched, along with metadata such as which command matched it)
2. A failed match (if any)
3. Leftover text after all succesful matches

i.e.,
```haskell
Text -> ([Match], Maybe FailedMatch, Text)
```
and if I'm getting my monad nesting right,
```haskell
StateT Text (WriterT [Match] (Except FailedMatch)) a
```

How can we express this in a convenient Lua API? I think we would have a Haskell variable that keeps track of the state, and expose functions (Lua wrappers around Haskell implementations) that perform the FileCheck operations that manipulate it, e.g.

- `check(String)`: Check the current text against the provided string, abort execution and throw `FailedMatch` if it doesn't match (a la `CHECK`)
- `next(String)`: Same, but like `CHECK-NEXT`
- `empty(String)`: like `CHECK-EMPTY`

It's easy to come up with useful extensions:

- `try_check(String) -> bool`: Like `check`, but returns a boolean rather than aborting execution on failure
- `try_{next,empty}(String) -> bool`: Same
- `{try_,}rx(String)`: Match a regular expression
- `{line,col}() -> int`: get the current source position
- so many more...

Variables would be taken care of by Lua's native concept of variables. Same for LLVM FileCheck's numeric variables. Iteration would be Lua loops.

I think it's worth exploring this before diving into #82 and #83.

langston-barrett:
Fixes #84, #83.


filecheck/src/FileCheck/Exception.hs
```diff
@@ -0,0 +1,80 @@
+{-# LANGUAGE LambdaCase #-}
```
langston-barrett, line 1: This could use your eyes, @RyanGlScott. The exception type has to be pushable to the Lua stack. Thus, you either have to marshal it to native Lua types, or use "light userdata", i.e., an opaque pointer. I've done the latter here. HsLua takes Haskell `Ptr`s as userdata, and the only `Storable` thing I found that I could easily get my hands on was a `StablePtr`. Thus, the exceptions are stashed in a `Ptr (StablePtr ...))`. This is inelegant. I'm very open to ideas and suggestions, though I wouldn't be unduly troubled by keeping the current approach.

filecheck/README.md
```diff
@@ -47,25 +47,25 @@ run tests:
 module Main (main) where
 
 import Control.Monad (filterM, forM)
-import Data.Text.IO qualified as Text.IO
+import ByteString qualified as BS
```
RyanGlScott: Should this be:

```suggestion
import Data.ByteString qualified as BS
```

filecheck/README.md
```diff
@@ -81,37 +81,141 @@ main = do
...
+Generated 37106428!
+Printing 37106428...
+```
+```lua
+check "Generated "
+num=string.find(text, "^%d+")
+check(string.format("Printing %d...", num))
+```
+
+The above example borders on *too much* logic for a test-case. Use discretion
```
RyanGlScott:
```suggestion
The above example borders on *too much* logic for a test case. Use discretion
```

filecheck/README.md
```diff
@@ -121,12 +225,18 @@ post for considerable motivation and discussion.
...
 
 In comparison to golden testing, FileCheck-style tests are *coarser*. They only
-check particular parts of the program's output. This can cause less churn in the
-test-suite when the program output changes in ways that are not relevant to the
-properties being tested. However, it is more complex. For example, it requires
-learning the FileCheck command language. It can also cause unexpected successes,
-e.g., if the program output contains the pattern being checked, but not in the
-proper place.
+check particular parts of the program's output. This can cause less churn in
+the test-suite when the program output changes in ways that are not relevant
```
RyanGlScott:
```suggestion
the test suite when the program output changes in ways that are not relevant
```

filecheck/src/FileCheck/Traceback.hs
```diff
@@ -0,0 +1,99 @@
...
+getTraceback sm =
+  Traceback . filter ((== FCE.sourceMapFile sm) . frameSource) . reverse <$> go 3 []
+  where
+    go level frames = do
+      mf <- getFrame sm level
+      case mf of
+        Nothing -> pure frames
+        Just f -> go (level + 1) (f : frames)
+
+printTraceback :: Traceback -> Text
```
RyanGlScott, line 97: General comment: make sure this (and any other exported functions) have Haddocks.

filecheck/src/FileCheck/Exception.hs
RyanGlScott: I don't think that there's anything inherently inelegant about using a `StablePtr` to marshal Haskell values to C code. That being said, a `StablePtr` can be safety casted to and from a `Ptr` anyway, so I think the code be simplified to something like this:

<details>

```diff
diff --git a/filecheck/src/FileCheck/Exception.hs b/filecheck/src/FileCheck/Exception.hs
index 56aa2ca..d540e4c 100644
--- a/filecheck/src/FileCheck/Exception.hs
+++ b/filecheck/src/FileCheck/Exception.hs
@@ -13,11 +13,8 @@ import Control.Monad.IO.Class (liftIO)
 import Data.Text qualified as Text
 import Data.Text.Encoding qualified as Text
 import FileCheck.Result qualified as FCR
-import Foreign.Marshal qualified as Foreign
-import Foreign.Ptr as Foreign
 import Foreign.StablePtr (StablePtr)
 import Foreign.StablePtr qualified as Foreign
-import Foreign.Storable as Foreign
 import HsLua qualified as Lua
 import HsLua.Core.Utf8 qualified as Lua.Utf8

@@ -31,7 +28,7 @@ data Exception
   | Failure NoMatch

 -- | Wrapper for 'FCR.Failure'
-newtype NoMatch = NoMatch (Ptr (StablePtr FCR.Failure))
+newtype NoMatch = NoMatch (StablePtr FCR.Failure)

 instance Show NoMatch where
   -- can't do IO here, but this Show instance won't be used anyway
@@ -46,16 +43,12 @@ instance Show Exception where
 instance X.Exception Exception

 noMatch :: NoMatch -> IO FCR.Failure
-noMatch (NoMatch ptr) = do
-  sp <- Foreign.peek ptr
-  Foreign.deRefStablePtr sp
+noMatch (NoMatch sp) = Foreign.deRefStablePtr sp

 throwNoMatch :: FCR.Failure -> Lua.LuaE Exception a
 throwNoMatch failure = do
   sp <- liftIO (Foreign.newStablePtr failure)
-  ptr <- liftIO Foreign.malloc
-  liftIO (Foreign.poke ptr sp)
-  Catch.throwM (Failure (NoMatch ptr))
+  Catch.throwM (Failure (NoMatch sp))

 instance Lua.LuaError Exception where
   popException = do
@@ -69,12 +62,12 @@ instance Lua.LuaError Exception where
         case top' of
           Just ptr -> do
             Lua.pop 1
-            pure (Failure (NoMatch ptr))
+            pure (Failure (NoMatch (Foreign.castPtrToStablePtr ptr)))
           Nothing -> Lua.failLua "Bad exception!"

   pushException =
     \case
       LuaException (Lua.Exception msg) -> Lua.pushstring (Lua.Utf8.fromString msg)
-      Failure (NoMatch ptr) -> Lua.pushlightuserdata ptr
+      Failure (NoMatch sp) -> Lua.pushlightuserdata (Foreign.castStablePtrToPtr sp)

   luaException s = LuaException (Lua.luaException s)
```
</details>

RyanGlScott: There is still a question of how we should manage the memory that the `StablePtr` holds on to, as the underlying `Failure` won't be garbage-collected due to being stored in a `StablePtr`. I suppose we could call `freeStablePtr` somewhere, although it's not clear to me where we would do so.

filecheck/src/FileCheck/Traceback.hs
langston-barrett: The only `exposed-modules:` is `FileCheck`, so this is only exported for use by other FileCheck. Would you like to see Haddocks for all of those, too?

RyanGlScott: Oops, I didn't realize that none of these modules were exposed. Never mind!

filecheck/src/FileCheck/Exception.hs
RyanGlScott: Then again, we probably don't really _need_ to manage the memory here, since this is fated to just throw an exception anyway, so the memory will be "freed" (in the [_Wing Commander_ sense](https://en.wikipedia.org/wiki/Wing_Commander_(video_game)#Development) of the word).

RyanGlScott (approved):
Hopefully this gives us enough power to write any FileCheck-based test we could ever conceive of. Thanks for working on this!

grease-cli/tests/llvm/abort.llvm.cbl
```diff
@@ -4,7 +4,7 @@
   (start start:
     (let g (resolve-global "abort"))
     (let h (load-handle Unit () g))
+;; next_line_must_fail()
```
langston-barrett, line 7: @RyanGlScott: I think the ability to write this helper feels especially cool, so I just wanted to brag about it 😁 

It checks that the output contains `Likely bug: unavoidable error`, followed by a string of the form `{file}:{line+1}` where `{file}` is the name of the test file in question (here `abort.llvm.cbl`) and `{line+1}` is the line in that file following the line that this function is called on.

PR #97, opened 2025-04-18
oughta: Rename from FileCheck

### LINKED ISSUE #96: filecheck: Rename to "Oughta"
This library bears less and less resemblance to LLVM's FileCheck. Plus, we probably shouldn't pun names anyway if we can avoid it. Let's rename it to "oughta" as in "ought to".

langston-barrett:
Fixes #96.


oughta/README.md
```diff
@@ -1,16 +1,19 @@
...
-More precisely, FileCheck allows you to build parsers for the output of your
-program. The inputs to FileCheck are the output of the program under test and a
-separate, generally quite short program written in the FileCheck parsing DSL.
+[FileCheck]: https://llvm.org/docs/CommandGuide/FileCheck.html
+[compiletest]: https://rustc-dev-guide.rust-lang.org/tests/compiletest.html 
+
+More precisely, Oughta provides a DSL to build *recognizers* (i.e., parsers that
+simply accept or reject an string). The inputs to Oughta are a string (usually,
+the output of the program under test) and a separate, generally quite short
+program written in the Oughta DSL.
```
langston-barrett, line 16: Note that I also did a little README editing here.

oughta/README.md
```diff
@@ -48,7 +51,7 @@ module Main (main) where
 
 import Control.Monad (filterM, forM)
 import Data.ByteString qualified as BS
-import FileCheck qualified as FC
+import Oughta qualified as Ota
```
RyanGlScott: `Oughta` is short enough where I wonder if we should just use `Oughta.foo` instead of `Ota.foo`.

oughta/oughta.cabal
```diff
@@ -1,26 +1,26 @@
...
 author: Galois, Inc.
 maintainer: grease@galois.com
 copyright: Galois, Inc. 2025
 license: BSD-3-Clause
 license-file: LICENSE
-synopsis: Like LLVM's FileCheck tool, but as a Haskell library.
+synopsis: A tool to test programs that output text.
 description:
-  Like LLVM's FileCheck tool, but as a Haskell library.
+   A tool to test programs that output text.
```
RyanGlScott: While we're at it, should we change "tool" to "library" here? As I understand it, Oughta isn't a standalone tool like LLVM's FileCheck.

RyanGlScott (approved):
There are some lingering `filecheck` references that ought to be updated as well:

* https://github.com/GaloisInc/grease/blob/4fbbb9f7a84048392f9b30897cb3c0613bba362f/oughta/src/Oughta/LuaApi.hs#L159
* https://github.com/GaloisInc/grease/blob/4fbbb9f7a84048392f9b30897cb3c0613bba362f/oughta/src/Oughta/Exception.hs#L35
* https://github.com/GaloisInc/grease/blob/4fbbb9f7a84048392f9b30897cb3c0613bba362f/doc/dev.md#L71

PR #101, opened 2025-04-18
Port Macaw S-expression tests to Oughta

### LINKED ISSUE #79: tests: Port Macaw S-expression tests to Oughta
#71 ported the LLVM S-expression tests to FileCheck. We should do the same for the Macaw S-expression tests. This has the following advantages:

- The tests may have richer assertions. For example, they can assert not only that there *was* a bug, but *what kind* of bug it was.
- They may test aspects of the UX of GREASE's output.
- They can be discovered from a directory, rather than [listed explicitly](https://github.com/GaloisInc/grease/blob/main/grease-cli/tests/Main.hs#L530-L595) in the test harness. This (1) avoids duplication, (2) makes it easier and faster to add a test, and (3) avoids recompiling the harness when new tests are added.

This change should be reflected [in the documentation](https://github.com/GaloisInc/grease/blob/main/doc/dev.md?plain=1#L174).

langston-barrett:
Fixes #79.


grease-cli/tests/Main.hs
```diff
@@ -465,64 +465,74 @@ makeSanityTests arch d =
...
+capture logRef =
+  LJ.LogAction (\msg -> liftIO (IORef.modifyIORef logRef (msg :)))
+
+withCapturedLogs ::
+  (GreaseLogAction -> IO ()) ->
+  IO Text.Text
+withCapturedLogs withLogAction = do
+  logRef <- IORef.newIORef []
+  withLogAction (capture logRef)
+  logs <- List.reverse <$> IORef.readIORef logRef
```
RyanGlScott, line 482: Leave a comment here saying why this `reverse` is necessary.

RyanGlScott (approved):
Hooray!

PR #108, opened 2025-04-25
Port binary tests to Oughta

### LINKED ISSUE #81: tests: Create a harness for Oughta-based tests targeting binaries
Our test-cases involving binaries are organized into [an elaborate directory hierarchy](https://github.com/GaloisInc/grease/blob/b0474a228dcf1a574eecce9a893a3d63e47f522b/doc/dev.md#binary-test-cases) which determines the expected outcome of the test. This has a few disadvantages:

- The expected outcomes are quite coarse-grained. See #29.
- Each architecture is expected to produce the same outcome. When they don't, they are added to [a list of expected failures](https://github.com/GaloisInc/grease/blob/b0474a228dcf1a574eecce9a893a3d63e47f522b/grease-cli/tests/Main.hs#L257), but the mere fact of being an expected failure is *also* incredibly coarse-grained.
- Each new type of outcome requires a new directory and changes to the test harness to support it. `grease-cli/tests/Main.hs` is now ~600 lines long, if all of our tests were FileCheck based I think we could probably cut this in half.
- Tests never examine *intrinsic* properties (e.g., was a function call skipped, or was its defintion available and used)

We should create a new part of the test harness that uses FileCheck to run tests involving binaries. The FileCheck commands (`CHECK:`) would live in the C file. We would use prefixes to distinguish between the output on different architectures (when necessary).

Eventually, it would be nice to port our existing tests to FileCheck. However, there are quite a lot of them, so let's just start by porting a few over (perhaps especially the ones with an xfail architecture) and adding new ones.

langston-barrett:
Fixes #81.


langston-barrett:
Some points to note:

- Requirements are now parsed from `.config` files, just like most other options
- Expected failures are now just marked with TODO comments

grease-cli/tests/Main.hs
```diff
@@ -460,28 +223,81 @@ withCapturedLogs withLogAction = do
...
+    let drop2 = FilePath.dropExtension drop1
+    let path = dir </> FilePath.addExtension drop2 "c"
+    content <- Text.IO.readFile path
+    let isArchComment =
+          Text.stripPrefix $
+            case arch of
+              Armv7 -> "// arm: "
+              PPC32 -> "// ppc32: "
+              X64 -> "// x64: "
+    let isGenericComment = Text.stripPrefix "// all: "
```
RyanGlScott, line 273: Do you think we should use triple slashes (`///`) instead of double (`//`) to match the existing conventions in `tests/llvm-bc`?

RyanGlScott (approved):

grease-cli/tests/Main.hs
langston-barrett: I don't have a strong opinion. The triple slashes in the bitcode tests were to disambiguate Oughta comments from other comments, but that isn't necessary here because we have these additional prefixes that name different architectures. I opted for maximal brevity rather than maximal consistency, I think either is reasonable.

RyanGlScott: Fair enough. I don't have a strong opinion either, so I'm content to use the existing conventions.

Issue #93, opened 2025-04-18
Embed flags in test files


langston-barrett:
Our test suite currently support a convention where GREASE flags can be specified in `.config` files next to the test files. These `.config` files are usually quite short:

https://github.com/GaloisInc/grease/blob/95ad235c674d4b3da7de545580550619ffc3b152/grease-cli/tests/llvm/declare-in-override.config

The tests are usually also pretty brief:

https://github.com/GaloisInc/grease/blob/95ad235c674d4b3da7de545580550619ffc3b152/grease-cli/tests/llvm/declare-in-override.llvm.cbl

I think it would be nice to embed the former into the latter inside a specially-formatted comment, e.g.
```
; flags: --overrides tests/llvm/extra/my-malloc.llvm.cbl
```
That way, you would get more context on a single screen. More test-cases (though not the linked one) would consist of just a single file. This is parsimonious for all kinds of reasons, including (theoretically) opening them up to automation such as fuzzing and test-case reduction.

langston-barrett:
Done in #111

PR #111, opened 2025-04-25
Discover tests by recursive directory traversal, embed flags in tests

### LINKED ISSUE #110: Make Tasty `TestTree` structure match the filesystem structure
Right now, our Tasty tests are organized into a tree like so (taking just the `div` test as an example):
```
ARM -> refine -> pos -> div
PPC32 -> refine -> pos -> div
x86_64 -> ...
```
However, this is contrary to how the tests are organized on disk. I propose we instead change the Tasty tree to match the filesystem tree, like so:
```
refine -> pos -> div -> ARM
refine -> pos -> div -> PPC32
...
```
This will probably let us simplify the test harness considerably.

langston-barrett:
Fixes #110.

I'd like to go further and discover *all* the tests by walking the filesystem from a few defined roots and discovering files and simulating them based in their file extension. But that can be a later PR.


RyanGlScott (approved):

langston-barrett:
Alright, I added a few more commits. Specifically, I also fixed #93.

doc/dev.md
```diff
@@ -156,25 +156,24 @@ following command-line options:
...
 
-If you wish to override these options, you can do so by adding one of the
-following files alongside the executables:
+If you wish to override these options, you can do so by adding a line that
+begins with `// flag: ` to the corresponding C program, like so:
 
-* `test.config`: This file should contain a list of command-line options, with
-  each one separated by whitespace, e.g.,
+```c
+// flag: --symbol foo --stack-argument-slots 5
```
RyanGlScott: Do you think this should be `flags` instead of `flag`? At first glance, I thought that you would have to specify one `// flag:` per command-line flag, but that isn't the case.

RyanGlScott (approved):
One question, but otherwise LGTM.

doc/dev.md
langston-barrett: Yeah, good call! Will switch it up.

PR #121, opened 2025-05-02
Move CLI parsing to `grease-cli` package

### LINKED ISSUE #119: Move CLI parsing code into `grease-cli`
The test-suite needs access to the CLI to support the `flags:` directive. However, it is unfortunate to have `optparse-applicative` as a dependency of `grease` (the library). We can improve the siutation by creating a [private sublibrary](https://cabal.readthedocs.io/en/latest/cabal-package-description-file.html#pkg-section-library-library) (AKA an "internal library") that contains the parser. The executable and test-suite can depend on both the `grease` library and this new internal library.

This is done in Cryptol: https://github.com/GaloisInc/cryptol/blob/6653adb71146c65c8863be4f5bd0b3ae61b4ee3c/cryptol.cabal#L276

langston-barrett:
The test-suite needs access to the CLI to support the `flags:` directive. However, it is unfortunate to have optparse-applicative as a dependency of grease (the library). To improve the situation, move all parsing-related code into the `grease-cli` package.

Fixes #119.

### DIFF
--- a/grease-cli/src/Grease/Cli.hs
+++ b/grease-cli/src/Grease/Cli.hs
@@ -54,10 +54,6 @@ boundedEnumMetavar _ = Opt.metavar $ varShowS ""
       List.intersperse (showChar '|') $
       List.map shows [minBound @a .. maxBound @a]
 
--- | 30 seconds in milliseconds
-defaultTimeout :: Int
-defaultTimeout = 30000
-
 entrypointParser :: Opt.Parser Entrypoint
 entrypointParser =
     addressParser Opt.<|> symbolParser Opt.<|> coreDumpParser Opt.<|>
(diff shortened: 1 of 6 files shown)


grease-cli/src/Grease/Cli.hs
```diff
@@ -0,0 +1,241 @@
...
+    varShowS :: ShowS
+    varShowS =
+      showParen True $
+      List.foldr (.) id $
+      List.intersperse (showChar '|') $
+      List.map shows [minBound @a .. maxBound @a]
+
+-- | 30 seconds in milliseconds
+defaultTimeout :: Int
+defaultTimeout = 30000
```
thebendavis: Looks like this PR adds a `defaultTimeout` here but doesn't remove the one in `grease/src/Grease/Options.hs` - can we deduplicate?

grease-cli/src/Grease/Cli.hs
```diff
@@ -0,0 +1,241 @@
...
+  optsJSON <- Opt.switch (Opt.long "json" <> Opt.help "output JSON")
+  optsIterations <- optional $ Opt.option Opt.auto (Opt.long "iters" <> Opt.help "maximum number of iterations of the refinement loop" <> Opt.metavar "N")
+  optsLoopBound <- GO.LoopBound <$> Opt.option Opt.auto (Opt.long "loop-bound" <> Opt.help "maximum number of executions of each loop in the program" <> Opt.metavar "N" <> Opt.value GO.defaultLoopBound)
+  optsNoHeuristics <- Opt.switch (Opt.long "no-heuristics" <> Opt.help "disable heuristics")
+  optsOverrides <-
+    Opt.many (Opt.strOption ( Opt.long "overrides"
+                              <> Opt.metavar "FILE"
+                              <> Opt.help "function overrides, in Crucible S-expression syntax"
+                              ))
+  optsTimeout <- GO.Milliseconds <$> Opt.option Opt.auto (Opt.long "timeout" <> Opt.help "timeout (in milliseconds, default 30000)" <> Opt.metavar "MILLIS" <> Opt.value defaultTimeout)
```
thebendavis: Prefer `<> Opt.showDefault` instead of hardcoding default timeout value in help text string

grease-cli/src/Grease/Cli.hs
```diff
@@ -0,0 +1,241 @@
...
+                          <> Opt.completeWith allMutableGlobalStateStrs)
+  optsRequirement <-
+    Opt.many (Opt.option (megaparsecReader reqParser)
+             (Opt.long "req"
+              <> Opt.help ("names of requirements to test " List.++ describeOptions allRequirementStrs)
+              <> Opt.metavar "REQS"
+              <> Opt.completeWith allRequirementStrs))
+  optsJSON <- Opt.switch (Opt.long "json" <> Opt.help "output JSON")
+  optsIterations <- optional $ Opt.option Opt.auto (Opt.long "iters" <> Opt.help "maximum number of iterations of the refinement loop" <> Opt.metavar "N")
+  optsLoopBound <- GO.LoopBound <$> Opt.option Opt.auto (Opt.long "loop-bound" <> Opt.help "maximum number of executions of each loop in the program" <> Opt.metavar "N" <> Opt.value GO.defaultLoopBound)
```
thebendavis: While we're here, consider including `Opt.showDefault` for this option?

grease-cli/src/Grease/Cli.hs
```diff
@@ -0,0 +1,241 @@
...
+                          <> Opt.showDefault
+                          <> Opt.completeWith allMutableGlobalStateStrs)
+  optsRequirement <-
+    Opt.many (Opt.option (megaparsecReader reqParser)
+             (Opt.long "req"
+              <> Opt.help ("names of requirements to test " List.++ describeOptions allRequirementStrs)
+              <> Opt.metavar "REQS"
+              <> Opt.completeWith allRequirementStrs))
+  optsJSON <- Opt.switch (Opt.long "json" <> Opt.help "output JSON")
+  optsIterations <- optional $ Opt.option Opt.auto (Opt.long "iters" <> Opt.help "maximum number of iterations of the refinement loop" <> Opt.metavar "N")
```
thebendavis: Can we make it clearer the default is unbounded (rather than some default bound), maybe something like:

```suggestion
  optsIterations <- optional $ Opt.option Opt.auto (Opt.long "iters" <> Opt.help "limit maximum number of iterations of the refinement loop" <> Opt.metavar "N")
```

thebendavis (approved):
LGTM - I left a few suggestions for improving the option handling while we're here. Thanks!

langston-barrett:
Good catches and suggestions! Thanks 😄

PR #118, opened 2025-05-01
Remove special handling of command-line flags from the test suite

langston-barrett:
Fixes #117.

PR #124, opened 2025-05-04
Allow Lua code to run GREASE in the test suite

### LINKED ISSUE #120: Allow Lua code to execute GREASE in the test-suite
Currently, our test-suite operates roughly like so:

1. Parse CLI flags from test files
2. Run GREASE with those flags (plus the test binary / S-expression file / LLVM bitcode module as an extra argument)
3. Pass the output of GREASE to Oughta, which runs some Lua code

I propose we instead use Oughta hooks (https://github.com/GaloisInc/oughta/issues/3) to install a Lua function (perhaps named `go`) that takes a list of strings, runs GREASE with those strings as arguments, and sets Oughta's `text` variable to the log output. (This last bit may require another Oughta function, e.g., `reset(String)`.)

This would make it possible to:

- have a single test-case that runs GREASE in multiple configurations on a single program and makes assertions about the various results
- have multiple test-cases that use a single binary (e.g., a Lua script embedded in comments in a C file and also a separate Lua script that just passes the path to the binary as an argument to `go`)
- have a single test-case that runs multiple programs (e.g., comparing the output of GREASE on binaries vs. LLVM bitcode from the same C program)

This would also enable us to remove the special-handling of the file path (item (2) above), probably by making another Lua variable for the binary path.

langston-barrett:
Fixes #120. Much like #118, this results in one additional line of boilerplate for the common case, but allows considerably more flexibility (see use-cases mentioned in the ticket, and the second commit on this branch).


doc/dev.md
```diff
@@ -191,10 +166,26 @@ organized into different subdirectories:
...
+
+The test harness provides the following Lua API, which enriches that of Oughta
+with GREASE-specific functionality. The API is presented with Haskell-style
+type signatures:
+
+- `prog :: String`: The name of the program under test, e.g., `test.llvm.cbl`.
+- `go :: String -> ()`: Run GREASE with the given string as the name of the
+  program. Frequently, `prog` is passed as the argument, e.g., `go(prog)`.
+- `flags :: [String] -> ()`: Append the given flags to the arguments passed to
+  GREASE when `go` is invoked.
```
RyanGlScott: It's a bit weird that we are describing Lua code with Haskell type signatures here. There's nothing _wrong_ with doing so per se, but the mixture of the two languages is a bit confusing. I wonder if we should instead present these like Lua API docs, i.e., something like:

> ```lua
> go(prog_name)
> ```
>
> Run GREASE with the given string as the name of the program [...]

(Inspired by the [`lua-cjson` docs](https://kyne.au/~mark/software/lua-cjson-manual.html))

grease-cli/tests/llvm/func-ptr.llvm.cbl
```diff
@@ -1,12 +1,24 @@
...
+;; check "Skipped call to a symbolic function handle"
+;; ok()
+
 ;; flags {"--symbol", "test"}
+;; flags {"--error-symbolic-fun-calls"}
 ;; go(prog)
+;; check "Goal failed:"
+;; check "Failed to load function handle"
+;; check "Cannot resolve a symbolic pointer to a function handle"
+;; no_heuristic()
```
RyanGlScott: If I understand correctly, this is running the same test case twice, but with different arguments (and a different set of checks). I assume this works because the two sets of semicolon-delimited lines have a newline in between them? If so, is this convention documented somewhere?

doc/dev.md
langston-barrett: Sure, will do!

grease-cli/tests/llvm/func-ptr.llvm.cbl
langston-barrett: > If I understand correctly, this is running the same test case twice, but with different arguments (and a different set of checks).

:+1:

> I assume this works because the two sets of semicolon-delimited lines have a newline in between them?

There's nothing special in the formatting - it could all be one big block. All of the `;;` comments are a single Lua program. This Lua program runs GREASE twice via `go`, resetting the built-in state of Oughta (e.g., `text`, output line / column tracking, matches) each time.

Let me know if that makes sense, and/or if there's a way to make it more clear!

RyanGlScott: Ah. Perhaps it would be worth clarifying that `go` can be run multiple times within a single file in its documentation, then?

RyanGlScott (approved):

PR #125, opened 2025-05-04
Write a literate Haskell/Markdown reference for function calls

langston-barrett:
This is an unfinished experiment, I'm hoping to gather early feedback to avoid wasted effort if we don't want to move forward with this approach.

This PR enables the creation of hybrid/literate Haskell/Markdown source files, and provides a half-completed demonstration on the `ResolveCall` module.

The motivation is as follows. There are GREASE modules that implement behavior that should probably be (or already is) documented as part of our user-facing reference documentation. Currently, we manage the documentation and the code separately as two files: user-facing docs in `doc/` (e.g., `shape-dsl.md`) and developer-facing implementations in `src/` (e.g., `Shape/Parse.hs`). This system works reasonably well. However, it has a few downsides:

- It is easy to update the code and forget to update the documentation.
- The documentation is often somewhat *duplicated* between the user-facing docs and Haddocks that describe intent (as a small example, consider [this Haddock](https://github.com/GaloisInc/grease/blob/3e85804b12a144aaddadee0981b592f9a272be78/grease/src/Grease/Macaw/FunctionOverride.hs#L173) vs our documentation on overrides)
- If the documentation is not duplicated, then the code is generally missing all description of intent, often with a pointer to the user-facing docs. The reader / developer is then left to (1) context-switch to open another document, and (2) search for the relevant bit of docs for the code they care about. (For an example, see `Parse.hs`).

To address these difficulties, this PR uses `markdown-unlit` together with GHC's native support for literate programming to enable the use of Markdown-formatted `.lhs` source files. The notion is to merge documentation with the code that implements the behaviors being described (for some relatively small number of GREASE modules). Ideally, these documents could be included in our mdBook.

However, this approach would also come with a number of notable downsides:

1. It requires a custom build tool (`markdown-unlit`). This isn't *too* bad, because it can be managed by Cabal.
2. It might degrate editor integrations, see

  - https://github.com/haskell/haskell-language-server/discussions/3949
  - https://github.com/haskell/haskell-language-server/issues/689
  - https://github.com/haskell/hie-bios/issues/187
3. It could degrade other tool support, e.g., `hlint`
3. The code becomes subjectively more "cluttered", albiet with hopefully-helpful explanations of what it does and why
4. The current approach uses symlinks from `.lhs` files to `.md` files so that the latter can be read e.g., on GitHub but GHC still picks them up. I haven't experienced this, but I think there's some folklore that symlinks are not very reliable or uniform cross-platform. I don't necessarily think this is a show-stopper - we can probably avoid symlinks if necessary. We might be able to just include `lhs` files directly into mdBook regardless of their extension, or there might be some hack to get GitHub to render `lhs` files as Markdown.

EDIT: See [here](https://github.com/GaloisInc/grease/blob/7d3db24b5b23908c9f79789297b02e9b224cf34d/grease/src/Grease/Macaw/ResolveCall.md) for what the rendered Markdown looks like. The results from mdBook would likely be quite similar.

### DIFF
--- a/grease/src/Grease/Macaw/ResolveCall.md
+++ b/grease/src/Grease/Macaw/ResolveCall.md
@@ -1,3 +1,16 @@
+# Function calls
+
+This document provides a detailed description of how GREASE treats function
+calls in binaries. It is serves as a reference for both users who wish to
+understand and customize GREASE's behavior, and for developers modifying and
+extending the implementation. To that end, it is structured as a literate
+Haskell file, but all the code blocks are hidden by default in collapsable
+sections like the following:
+
+<details>
+<summary>Imports, etc.</summary>
+
+```haskell
 {-|
 Copyright        : (c) Galois, Inc. 2024
 Maintainer       : GREASE Maintainers <grease@galois.com>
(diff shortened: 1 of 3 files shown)


grease/src/Grease/Macaw/ResolveCall.md
```diff
@@ -162,101 +169,180 @@ lookupFunctionHandle bak la halloc arch memory symMap pltStubs dynFunMap funOvs
...
+<details>
+<summary>Skipping calls to invalid addresses</summary>
+
+```haskell
+    Just bv -> do
+      -- This conversion is safe iff MC.ArchAddrWidth arch <= 64
+      let bvWord64 = fromIntegral @Integer @Word64 (BV.asUnsigned bv)
+      let bvMemWord = EL.memWord bvWord64
+      case Loader.resolveAbsoluteAddress memory bvMemWord of
+        -- TODO: This should probably be a hard error (at least by default)?
```
langston-barrett, line 209: Regardless of what we do with this PR, I do think this should likely be a hard error...

grease/src/Grease/Macaw/ResolveCall.md
```diff
@@ -162,101 +169,180 @@ lookupFunctionHandle bak la halloc arch memory symMap pltStubs dynFunMap funOvs
...
+  let override = C.mkOverride' funcName (regStructRepr arch) $ do
+        args <- C.getOverrideArgs
+        let regs' = Ctx.last $ C.regMap args
+        (arch ^. archOffsetStackPointerPostCall) (C.regValue regs')
+  pure $ useFnHandleAndState handle (C.UseOverride override) st
+```
+
+</details>
 
+```haskell
```
langston-barrett, line 345: If we want to proceed with this approach, I would make a new section here describing system calls.

grease/src/Grease/Macaw/ResolveCall.md
```diff
@@ -162,101 +169,180 @@ lookupFunctionHandle bak la halloc arch memory symMap pltStubs dynFunMap funOvs
...
+        Nothing -> skip (InvalidAddress (ppMemWord bvMemWord))
+        Just funcAddrOff ->
+          callConcreteAddr bak la halloc arch memory symMap pltStubs dynFunMap funOvs st mem regs bvMemWord funcAddrOff
+  where
+    ppMemWord = BV.ppHex knownNat . BV.word64 . EL.memWordValue
+```
+</details>
+  
+<details>
+<summary>TODO</summary>
```
langston-barrett, line 219: If we want to proceed with this approach, I'd document the body of this function similarly to how `lookupFunctionHandle` is described above.

RyanGlScott (commented):
I am ambivalent about the use of Literate Haskell (i.e., `.lhs` files). I think that they can make it a bit easier to remember to keep the documentation and code in sync, but I also find that the add a lot of clutter when simply browsing through the code, which is the mode of operation I find myself in more often than not.

I don't use HLS myself, but I do worry that complicating the build system might raise the barrier to entry for those who do. It sounds like if we do pursue this direction, we should ensure that we document how to load the code into HLS properly (perhaps by adding a custom `hie.yaml` file that works around any HLS-related issues).

langston-barrett:
Yeah, I'm also ambivalent! Do you have a *preference*?

RyanGlScott:
I have a slight preference towards not using Literate Haskell, but I don't feel strongly about this.

langston-barrett:
> But I think my experience of [the rendered markdown](https://github.com/GaloisInc/grease/blob/7d3db24b5b23908c9f79789297b02e9b224cf34d/grease/src/Grease/Macaw/ResolveCall.md) is that interspersing text snippets with code snippets (which include comments) makes it harder to follow (e.g. "What function am I in/what were the args again? what indentation level are we at?" etc.).

Yeah, after thinking a bit more about this I think there's really no reason to make the code viewable in the rendered Markdown - anyone interested in the code should just look at the code. One could likely exclude the code from the Markdown rendered by mdBook with a simple [custom preprocessor](https://rust-lang.github.io/mdBook/format/configuration/preprocessors.html#provide-your-own-command) script.

As for "What indentation level are we at?", I think the ideal would be to structure the code in such a way that these questions are uncommon (because the indentiation level is quite low) but realize that that's a tall order.

> I also worry about the editor/tooling issues you identified. Opening the current grease repo in VSCode with Haskell extensions builds fine (pending https://github.com/GaloisInc/grease/pull/135), but with this branch there are errors related to the markdown-unlit use. I expect we could resolve these but I think there is value in keeping things simple to maximize the chance that someone new could check out grease and have things "just work" in their setting without us having to check and update any more of these workflow-support things than we need to.

Yep, agreed!

Thanks for thinking through this with me, you two! I think we're all agreed about the benefits and drawbacks and it's worth maintaining the status quo.

PR #127, opened 2025-05-04
`initState`: Accept additional Crucible intrinsic types

### LINKED ISSUE #126: Allow initializing `grease` state with additional Crucible intrinsics
This applies to both the LLVM and Macaw frontends to `grease`, but as a specific example, consider [`Grease.Macaw.initState`](https://github.com/GaloisInc/grease/blob/db58869d7096da9814e18a3c8f8c4955ddea304b/grease/src/Grease/Macaw.hs#L512-L547). Currently, this assumes that it will always use precisely the set of intrinsics needed for `crucible-llvm`:

https://github.com/GaloisInc/grease/blob/db58869d7096da9814e18a3c8f8c4955ddea304b/grease/src/Grease/Macaw.hs#L557-L559

This is not an ideal for downstream applications, which may want to initialize `grease`-specific state using additional intrinsics. For instance, it is not straightforward to combine `Grease.Macaw.initState` with a `crucible-symio` filesystem, which also requires executing Crucible with [`llvmSymIOIntrinsicTypes`](https://github.com/GaloisInc/crucible/blob/9cb4f93d17c48890a1643a005120b365e6873c56/crucible-llvm/src/Lang/Crucible/LLVM/SymIO.hs#L269-L276).

At a first approximation, fixing this will require adding an additional `IntrinsicTypes sym` argument to `initState`, which will be combined with `llvmIntrinsicTypes` before executing Crucible.

RyanGlScott:
Fixes #126.

### DIFF
--- a/grease-cli/grease-cli.cabal
+++ b/grease-cli/grease-cli.cabal
@@ -211,7 +211,7 @@ test-suite grease-tests
     , hedgehog
     , lumberjack
     , macaw-base
-    , oughta
+    , oughta >= 0.1 && < 0.2
     , parameterized-utils
     , prettyprinter
     , tasty
--- a/grease-cli/src/Grease/Main.hs
+++ b/grease-cli/src/Grease/Main.hs
@@ -84,6 +84,7 @@ import qualified Text.LLVM as L
 -- parameterized-utils
 import Data.Parameterized.Classes (IxedF'(ixF'))
 import qualified Data.Parameterized.Context as Ctx
+import qualified Data.Parameterized.Map as MapF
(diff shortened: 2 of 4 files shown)


langston-barrett (approved):

RyanGlScott:
After some discussion in https://github.com/GaloisInc/grease/issues/131#issuecomment-2851363117, I am going to close this. I think we actually want to wire in the `crucible-symio` overrides unconditionally into `grease`, which will require us to install the `crucible-symio`-related intrinsic types anyway. As such, there isn't a pressing need to make this part of `grease` extensible.

Issue #128, opened 2025-05-05
CI: Test suite fails to build on `main`


RyanGlScott:
As seen in [this failing CI job](https://github.com/GaloisInc/grease/actions/runs/14801523047/job/41561152479#step:8:90):

```
[2 of 2] Compiling Main             ( tests/Main.hs, /home/runner/work/grease/grease/dist-newstyle/build/x86_64-linux/ghc-9.8.4/grease-cli-0.1.0.0/t/grease-tests/build/grease-tests/grease-tests-tmp/Main.o ) [Source file changed]
[2 of 2] Linking /home/runner/work/grease/grease/dist-newstyle/build/x86_64-linux/ghc-9.8.4/grease-cli-0.1.0.0/x/grease/build/grease/grease [Objects changed]

tests/Main.hs:135:22: error: [GHC-83865]
Error:     • Couldn't match expected type: IO Oughta.Result
                  with actual type: Oughta.Output -> IO Oughta.Result
    • Probable cause: ‘Oughta.check’ is applied to too few arguments
      In a stmt of a 'do' block:
        Oughta.Result r <- Oughta.check prog output
      In the expression:
        do let action = withCapturedLogs $ ...
           logTxt <- X.try @X.SomeException action
                       <&>
                         \case
                           Left e -> Text.pack ("Exception: " ++ show e)
                           Right t -> t
           Text.IO.writeFile (FilePath.replaceExtension path "out") logTxt
           let output = Oughta.Output (Text.encodeUtf8 logTxt)
           ....
      In an equation for ‘oughta’:
          oughta opts path prog0
            = do let action = ...
                 logTxt <- X.try @X.SomeException action
                             <&>
                               \case
                                 Left e -> Text.pack ("Exception: " ++ show e)
                                 Right t -> t
                 Text.IO.writeFile (FilePath.replaceExtension path "out") logTxt
                 ....
    |
135 |   Oughta.Result r <- Oughta.check prog output
    |                      ^^^^^^^^^^^^^^^^^^^^^^^^

tests/Main.hs:135:35: error: [GHC-83865]
Error:     • Couldn't match expected type ‘Oughta.Hooks’
                  with actual type ‘Oughta.LuaProgram’
    • In the first argument of ‘Oughta.check’, namely ‘prog’
      In a stmt of a 'do' block:
        Oughta.Result r <- Oughta.check prog output
      In the expression:
        do let action = withCapturedLogs $ ...
           logTxt <- X.try @X.SomeException action
[...]

PR #132, opened 2025-05-05
ghidra: use built-in rust detection

thebendavis:


RyanGlScott (commented):
Nice!

Do you know if this requires a certain minimum version of Ghidra to use? If so, it would be worth documenting this requirement in [the manual](https://github.com/GaloisInc/grease/blob/3e85804b12a144aaddadee0981b592f9a272be78/doc/ghidra-plugin.md).

RyanGlScott:
Also, the CI is failing for unrelated reasons (#128). Cherry-picking [this commit](https://github.com/GaloisInc/grease/pull/127/commits/879cff10f37969e51579017a8f6052e5af9e70a0) from #127 should fix it.

langston-barrett (approved):

RyanGlScott (approved):

Issue #130, opened 2025-05-05
Inconsistent behavior of `htons` override between LLVM and Macaw frontends


RyanGlScott:
Consider the following program, which uses the `htons` function to turn a number in host byte-order (5000) to network byte-order (34835):

```c
#include <arpa/inet.h>
#include <stdlib.h>

int main(void) {
    if (htons(5000) != 34835) {
        abort();
    }
}
```

If you compile this to LLVM bitcode and analyze it with `grease`, then it detects no issues, as expected:

```
$ cabal -v0 run exe:grease -- endianness.bc --symbol main
[2025-05-05 12:39:10] Analyzing from entrypoint 'main'
[2025-05-05 12:39:10] All goals passed!
[2025-05-05 12:39:10] Finished analyzing 'main'.
```

If you compile this to an x86-64 binary and analyze it with `grease`, on the other hand, then `grease` will instead abort:

```
$ cabal -v0 run exe:grease -- endianness.exe --symbol main
[2025-05-05 12:39:43] Analyzing from entrypoint 'main'
[2025-05-05 12:39:43] Warning: no annotation on predicate!
[2025-05-05 12:39:43] Finished analyzing 'main'. Likely bug: unavoidable error (safety condition is unsatisfiable) at 0x1000116c
0x1000116c: error: in main
Call to abort
```

This is weirdly inconsistent. The culprit ultimately lies in how `grease` is configuring the endianness between the LLVM and Macaw frontends. In the LLVM frontend, `grease` will parse the bitcode file's data layout to figure out what the host endianness is (on my machine, this is little-endian). In the Macaw frontend, however, `grease` simply uses `defaultDataLayout`, which [defaults to big-endian](https://github.com/GaloisInc/crucible/blob/9cb4f93d17c48890a1643a005120b365e6873c56/crucible-llvm/src/Lang/Crucible/LLVM/DataLayout.hs#L239). This means that when analyzing an x86-64 binary, `grease` will interpret `htons(5000)` to return `5000` instead of `34835`, even though x86-64 is a little-endian architecture!

At the very minimum, we should update the `defaultDataLayout` to use the same endianness as the binary's underlying architecture, which would fix the test case above.

Issue #129, opened 2025-05-05
`ArchContext`'s `_archEndianness` field is redundant


RyanGlScott:
Currently, the `ArchContext` data type stores an `_archEndianness :: EndianForm` field to record the endianness of the architecture. This is redundant, however, as the `_archInfo :: ArchitectureInfo` field already stores the endianness through its own [`archEndianness`](https://github.com/GaloisInc/macaw/blob/2835b95ec912c682e077744f2f95130799935628/base/src/Data/Macaw/Architecture/Info.hs#L241-L242) field. We should remove `ArchContext`'s `_archEndianness` to avoid duplication.

PR #134, opened 2025-05-05
Respect architecture endianness in Macaw frontend

RyanGlScott:
We now override the `defaultDataLayout` with a sensible endianness value based on the architecture being used. As a result, functions like `htons` now do different things on different architectures, as expected. This fixes https://github.com/GaloisInc/grease/issues/130.

While I was in the neighborhood, I also cleaned up the handling of endianness in `ArchContext`. `ArchContext` was redundantly storing a separate endianness value that could already be obtained from the `ArchitectureInfo`. As such, I removed the redundant `_archEndianness` field in `ArchContext`, thereby fixing #129.

### DIFF
--- a/grease-cli/tests/arm/htons.armv7l.cbl
+++ b/grease-cli/tests/arm/htons.armv7l.cbl
@@ -1,0 +1,18 @@
+; Copyright (c) Galois, Inc. 2025
+
+;; flags {"--symbol", "test"}
+;; go(prog)
+
+(declare @htons ((host-short (Bitvector 16))) (Bitvector 16))
+
+(defun @test ((regs AArch32Regs)) AArch32Regs
+  (start start:
+    (let host-short (bv 16 5000))
+    ; AArch32 is little-endian, so `htons` should swap the byte order
+    (let expected-network-short (bv 16 34835))
+    (let actual-network-short (funcall @htons host-short))
+    (assert!
+      (equal? expected-network-short actual-network-short)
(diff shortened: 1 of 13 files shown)


grease-cli/tests/arm/htons.armv7l.cbl
```diff
@@ -0,0 +1,23 @@
...
+    (let actual-network-short (funcall @htons host-short))
+    (let res (equal? expected-network-short actual-network-short))
+    (branch res success: failure:))
+
+  (defblock success:
+    (return regs))
+
+  (defblock failure:
+    (funcall @abort)
+    (return regs)))
```
langston-barrett: Could this be shortened using `assert!`?

grease-cli/tests/ppc32/htons.ppc32.cbl
```diff
@@ -0,0 +1,23 @@
...
+
+; flags: --symbol test
+
+(declare @abort () Unit)
+(declare @htons ((host-short (Bitvector 16))) (Bitvector 16))
+
+(defun @test ((regs PPC32Regs)) PPC32Regs
+  (start start:
+    (let host-short (bv 16 5000))
+    ; x86-64 is big-endian, so `htons` should return the input unchanged
```
langston-barrett:
```suggestion
    ; PPC is big-endian, so `htons` should return the input unchanged
```

langston-barrett (approved):
Looks great, thanks!

PR #135, opened 2025-05-05
Use package-relative path to buildinfo

thebendavis:
embedFileRelative treats the path as relative to the package root, so this change removes requirements/assumptions about the current path during the build process.

This change is motivated by opening the grease repo in VSCode and getting the error: 

```
Exception when trying to run compile-time code: typecheck(GHC-87897) [Ln 91, Col 13]
^ grease.buildinfo.json: withBinaryFile: does not exist (No such file or directory)
Code: (embedFile "grease.buildinfo.json")
• In the untyped splice: $(embedFile "grease.buildinfo.json")
```

Presumably the paths used by the VSCode Haskell integration don't match those of our command-line developer workflows. This change resolves the issue.


RyanGlScott:
Nice!

Per [the Haddocks](https://hackage.haskell.org/package/file-embed-0.0.16.0/docs/src/Data.FileEmbed.html#embedFileRelative), `embedFileRelative` was introduced in `file-embed-0.0.16`, so I'd recommend adding `file-embed >= 0.0.16` lower version bounds to this line in the `grease.cabal` file:

https://github.com/GaloisInc/grease/blob/3e85804b12a144aaddadee0981b592f9a272be78/grease-cli/grease-cli.cabal#L139

langston-barrett:
(CI is failing due to #128)

thebendavis:
> Per [the Haddocks](https://hackage.haskell.org/package/file-embed-0.0.16.0/docs/src/Data.FileEmbed.html#embedFileRelative), `embedFileRelative` was introduced in `file-embed-0.0.16`, so I'd recommend adding `file-embed >= 0.0.16` lower version bounds 

Good call - done. Thanks!

RyanGlScott (approved):

PR #140, opened 2025-05-06
Add bounds to Cabal files

### LINKED ISSUE #136: Add bounds to Cabal files
Our CI recently started failing on `main` due to missing bounds on a dependency (#128). We should add such bounds, as well as builds against Stackage LTS package sets, `--prefer-oldest`, and `cabal-force-upper-bound` a la [haskell-ci](https://github.com/GaloisInc/.github?tab=readme-ov-file#example-for-library-authors).

langston-barrett:
Fixes #136. Skips the architecture-specific ones, as the dependencies of these are either (1) covered by `grease{,-cli}` or (2) unversioned Macaw architecture-specific packages.

### DIFF
--- a/grease-cli/grease-cli.cabal
+++ b/grease-cli/grease-cli.cabal
@@ -1,4 +1,4 @@
-cabal-version: 2.4
+cabal-version: 3.0
 name: grease-cli
 version: 0.1.0.0
 author: Galois, Inc.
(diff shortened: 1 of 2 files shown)


RyanGlScott (commented):
I like the idea of adding bounds to Hackage dependencies (e.g., `oughta`), but I am less convinced that this is worthwhile for packages that are vendored in as submodules (e.g., `crucible`), as their version bounds are fixed by whatever the submodules use. My recommendation would be to leave version bounds off of these submodule dependencies to avoid churn whenever we need to bump submodules, although I could be convinced to use this approach if you prefer it.

langston-barrett:
I was worried that it might be possible for the Cabal solver to pick the Hackage version of our vendored libraries if we didn't include a lower bound, especially with flags such as `--prefer-oldest`. Do you happen to know if that's (im)possible?

One other (very small) advantage to including the versions is that `cabal outdated` can tell us when our submodules are behind the Hackage-released versions.

RyanGlScott:
> I was worried that it might be possible for the Cabal solver to pick the Hackage version of our vendored libraries if we didn't include a lower bound, especially with flags such as `--prefer-oldest`. Do you happen to know if that's (im)possible?

I believe that if you include a package explicitly in your `cabal.project` file's `packages:` stanza, then it will always pick that version over a Hackage package of the same name. (I haven't tested this with `--prefer-oldest` specifically, but I don't see why that flag in particular would change this property.)

> One other (very small) advantage to including the versions is that `cabal outdated` can tell us when our submodules are behind the Hackage-released versions.

A fair point, although we could just as well look at the `git log` in the submodules to check this property.

In any case, I'll let you make the call—I'm happy with either approach.

langston-barrett:
Okay, in that case I'll remove the bounds. Can always add them if we add a CI build with `--prefer-oldest` and it fails for that reason.

I think the churn outweighs the advantages of `cabal outdated`.

RyanGlScott (approved):

PR #141, opened 2025-05-06
ci: Build against additional dependency versions

langston-barrett:
Currently based on #140. Meant to ensure that we (1) maintain compatibility with a sufficiently-wide range of versions of our dependencies (by building against Stackage LTS) and (2) that our version bounds are accurate (by building with `--prefer-oldest`).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -28,25 +28,6 @@ jobs:
         haddock: [true]  # see `include` below
         os: [ubuntu-22.04]
         pre-hook: [""]  # see `include` below
-        include:
-        # Include a build (with latest Linux OS) against Stackage LTS package
-        # sets for each supported version of GHC that has one to verify that we
-        # support building against these widely-used package sets.
-        - cabal: "3.14.1.1"
-          cache-key-prefix: stackage
-          ghc: 9.8.4
-          haddock: false
-          os: ubuntu-22.04
-          pre-hook: |
-            echo 'import: https://www.stackage.org/lts-23.19/cabal.config' >> cabal.project
-        - cabal: "3.14.1.1"
-          cache-key-prefix: stackage
-          ghc: 9.4.8
-          haddock: false
-          os: ubuntu-22.04
-          pre-hook: |
-            echo 'import: https://www.stackage.org/lts-21.25/cabal.config' >> cabal.project
-    name: GREASE - GHC v${{ matrix.ghc }} - ${{ cache-key-prefix }}
     steps:
     - name: Checkout
       uses: actions/checkout@v4
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -14,22 +14,48 @@ env:
   # a valid version of the "happy" tool).
   CACHE_VERSION: 1
 
+
 jobs:
   build:
     runs-on: ${{ matrix.os }}
     strategy:
-      matrix:
-        ghc: ["9.4.8", "9.6.6", "9.8.4"]
-        cabal: ["3.14.1.1"]
-        os: [ubuntu-22.04]
       # complete all jobs
       fail-fast: false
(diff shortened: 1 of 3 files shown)


langston-barrett:
Huh, `--prefer-oldest` fails with
```
Error: [Cabal-7107]
Could not resolve dependencies:
[__0] trying: asl-parser-0.1.0.0 (user goal)
[__1] trying: text-0.1 (dependency of asl-parser)
[__2] next goal: base (dependency of asl-parser)
[__2] rejecting: base; 3.0.3.1, 3.0.3.2, 4.0.0.0, 4.1.0.0, 4.2.0.0, 4.2.0.1, 4.2.0.2, 4.3.0.0, 4.3.1.0, 4.4.0.0, 4.4.1.0, 4.5.0.0, 4.5.1.0, 4.6.0.0, 4.6.0.1, 4.7.0.0, 4.7.0.1, 4.7.0.2, 4.8.0.0, 4.8.1.0, 4.8.2.0, 4.9.0.0, 4.9.1.0, 4.10.0.0, 4.10.1.0, 4.11.0.0, 4.11.1.0, 4.12.0.0, 4.13.0.0, 4.14.0.0, 4.14.1.0, 4.14.2.0, 4.14.3.0, 4.15.0.0, 4.15.1.0, 4.16.0.0, 4.16.1.0, 4.16.2.0, 4.16.3.0, 4.16.4.0, 4.17.0.0, 4.17.1.0, 4.17.2.0 (constraint from non-reinstallable package requires installed instance)
[__2] rejecting: base-4.17.2.1/installed-4.17.2.1 (conflict: text => base>=4 && <4.3)
[__2] skipping: base; 4.17.2.1, 4.18.0.0, 4.18.1.0, 4.18.2.0, 4.18.2.1, 4.18.3.0, 4.19.0.0, 4.19.1.0, 4.19.2.0, 4.20.0.0, 4.20.0.1, 4.20.1.0, 4.21.0.0 (has the same characteristics that caused the previous version to fail: excluded by constraint '>=4 && <4.3' from 'text')
[__2] fail (backjumping, conflict set: asl-parser, base, text)
Backjump limit reached (currently 4000, change with --max-backjumps or try to run with --reorder-goals).
```
I don't really get this, as `grease{,-cli}.cabal` both explicitly bound `text`.

RyanGlScott:
Wow, that is confusing indeed. It appears that it tries to solve for a build plan containing [`text-0.1`](https://hackage.haskell.org/package/text-0.1) (which, to be fair, _is_ the oldest `text` version that might be allowed based solely on the version bounds on `text` alone), thinks for a bit, realizes that `text-0.1` is incompatible with the installed version of `base`, and then... just gives up? I suppose we could try increasing `--max-backjumps`, but it's not really obvious to me what the right number to use should be.

langston-barrett:
I wonder if `--prefer-oldest` chokes when a lot of dependencies themselves don't have any bounds. I've opened https://github.com/GaloisInc/arm-asl-parser/issues/5 to track adding them to `arm-asl-parser`.

langston-barrett:
Okay, I've reverted commits adding Stackage and lower bounds builds, but at least the upper bounds build works.

.github/workflows/ci.yaml
```diff
@@ -14,23 +14,55 @@ env:
...
+              --location \
+              --proto '=https' \
+              --show-error \
+              --silent \
+              --tlsv1.2 \
+              https://github.com/nomeata/cabal-force-upper-bound/releases/download/0.1/cabal-force-upper-bound.linux.gz | \
+              gunzip > \
+              /usr/local/bin/cabal-force-upper-bound
+            chmod +x /usr/local/bin/cabal-force-upper-bound
+            for f in ./grease*/grease*.cabal; do
```
RyanGlScott: Should this also include `doc/grease-diagrams` and `elf-edit-core-dump`?

RyanGlScott (approved):

PR #151, opened 2025-05-08
Bundle `crucible-llvm` symio-related overrides (`open`, `close`, `read`, `write`)

### LINKED ISSUE #139: Build in `crucible-symio`-related overrides for `open`, `close`, `read`, and `write`
`crux-llvm` includes overrides for the `open`, `close`, `read`, and `write` functions that are built on top of the machinery in `crucible-symio`. These are not included in the default set of overrides found in `crucible-llvm`, as they require some [additional setup](https://github.com/GaloisInc/crucible/blob/9cb4f93d17c48890a1643a005120b365e6873c56/crucible-llvm/src/Lang/Crucible/LLVM/SymIO.hs#L162-L197) to install. As such, `grease` currently does not include these `crucible-symio`-based overrides. We should perform the additional setup and install them for consistency with `crux-llvm` and for the benefit of `grease` users.

RyanGlScott:
This requires some additional plumbing to thread out the various bits of data that these overrides require (e.g., an `LLVMFileSystem` value), but otherwise this proves straightforward.

Fixes #139.

### DIFF
--- a/doc/overrides.md
+++ b/doc/overrides.md
@@ -246,6 +246,10 @@ signatures:
 - `i64 @labs( i64 )`
 - `i64 @llabs( i64 )`
 - `i32 @__cxa_atexit( void (i8*)*, i8*, i8* )`
+- `i32 @open( i8*, i32 )`
+- `i32 @close( i32 )`
+- `ssize_t @read( i32, i8*, size_t )`
+- `ssize_t @write( i32, i8*, size_t )`
 
 For LLVM programs (but not binaries), the following built-in overrides are also
 available:
@@ -338,6 +342,7 @@ schema:
 - `<<LANES> x i<N>>`: (there is no syntax for this Crucible type)
 - `ptr`: `(Ptr <word-size>)`
 - `size_t`: `(Ptr <word-size>)`
+- `ssize_t`: `(Ptr <word-size>)`
 - `<T>*`: `(Ptr <word-size>)`
 - `void`: `Unit`
 
@@ -368,6 +373,16 @@ The following overrides merit a bit of discussion:
   Allocate a fresh pointer of the given size in the underlying
   [memory model](./memory-model.md). This pointer is assumed not to alias with
   any pointers allocated by previous calls to `malloc`.
+
+- `open`, `close`, `read`, and `write`
+
+  These overrides leverage Crucible's experimental symbolic I/O capabilities.
+  In particular, these overrides require the use of a symbolic filesystem,
+  which must be specified with `--fs-root <path-to-filesystem-root>` when
+  invoking `grease`. See [symbolic I/O] for a more detailed description of what
+  the contents of the symbolic filesystem should look like.
+
+  [symbolic I/O]: https://github.com/GaloisInc/crucible/tree/master/crux-llvm#symbolic-io-experimental
 
 ## Startup overrides
 
--- a/grease-cli/src/Grease/Cli.hs
+++ b/grease-cli/src/Grease/Cli.hs
@@ -189,6 +189,13 @@ simOpts = do
     Opt.switch ( Opt.long "rust"
                  <> Opt.help "Use simulator settings that are more likely to work for Rust programs"
                )
+  simFsRoot <-
+    Opt.optional
+          ( Opt.strOption
+            (  Opt.long "fs-root"
+            <> Opt.metavar "PATH"
+            <> Opt.help "The path to the symbolic filesystem"
+            ))
   pure GO.SimOpts{..}
     where
       allMutableGlobalStateStrs :: [String]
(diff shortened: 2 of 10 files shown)


RyanGlScott:
It turns out that porting `ambient-verifier`'s existing test cases involving `open`/`read`/`write` _et al._ is rather annoying, and for a pretty silly reason: in order to create the file path that you pass to `open`, you have to turn a `crucible-syntax` `String` into a C-style pointer of bytes, which is rather tedious. So tedious, in fact, that `ambient-verifier` includes a built-in [`write-c-string :: Pointer -> String Unicode -> Unit`](https://github.com/GaloisInc/ambient-verifier/blob/eab04abb9750825a25ec0cbe0379add63f05f6c6/README.rst#built-in-overrides) function for this purpose. I stopped short of porting over `write-c-string`, but let me know if you'd want this before landing the changes here.

langston-barrett (commented):
Might be worth adding some information about symbolic file support to the user-facing docs, even if it's just a link to the symio docs in the Crucible repo.

langston-barrett:
I'm fine adding these without tests for the moment, let's perhaps file an upstream issue about adding C-string related helpers to Crucible-LLVM-Syntax, and file an issue on this repo saying we should port the ambient tests when that upstream functionality is merged.

RyanGlScott:
See https://github.com/GaloisInc/grease/issues/152.

doc/overrides.md
```diff
@@ -246,6 +246,10 @@ signatures:
 - `i64 @labs( i64 )`
 - `i64 @llabs( i64 )`
 - `i32 @__cxa_atexit( void (i8*)*, i8*, i8* )`
+- `i32 @open( i8*, i32 )`
+- `i32 @close( i32 )`
+- `ssize_t @read( i32, i8*, size_t )`
+- `ssize_t @write( i32, i8*, size_t )`
```
langston-barrett, line 252: I think these [merit additional discussion](https://github.com/GaloisInc/grease/pull/151/files#diff-7dd685a8abdc7af96594c6f1cc945ea58675e9238a0392d269c3256fa50ad329L354), perhaps just a link [here](https://github.com/GaloisInc/crucible/tree/fb06f32e70de358c4affc5f12e54b0cbb248b5ce/crucible-symio) and/or [here](https://github.com/GaloisInc/crucible/tree/master/crux-llvm#symbolic-io-experimental).

grease-cli/src/Grease/Cli.hs
```diff
@@ -189,6 +189,13 @@ simOpts = do
     Opt.switch ( Opt.long "rust"
                  <> Opt.help "Use simulator settings that are more likely to work for Rust programs"
                )
+  simFsRoot <-
+    Opt.optional
+          ( Opt.strOption
+            (  Opt.long "fsroot"
```
langston-barrett: These read as two separate words to me
```suggestion
            (  Opt.long "fs-root"
```

grease-cli/src/Grease/Main.hs
```diff
@@ -487,6 +492,28 @@ interestingConcretizedShapes names initArgs (ConcArgs cArgs) =
...
+-- | Helper, not exported
+--
+-- Initialize the symbolic file system.
+initialLlvmFileSystem ::
+  ( C.IsSymInterface sym
+  , Mem.HasPtrWidth ptrW
+  ) =>
+  C.HandleAllocator ->
+  sym ->
+  C.NatRepr ptrW ->
```
langston-barrett: Why not use `?ptrWidth` from the `HasPtrWidth` constraint?

grease-cli/src/Grease/Main.hs
```diff
@@ -487,6 +492,28 @@ interestingConcretizedShapes names initArgs (ConcArgs cArgs) =
...
+  IO ( CLLVM.SymIO.LLVMFileSystem ptrW
+     , C.SymGlobalState sym
+     , CLLVM.SymIO.SomeOverrideSim sym ()
+     )
+initialLlvmFileSystem halloc sym ptrW simOpts = do
+  fileContents <-
+    case simFsRoot simOpts of
+      Nothing -> pure SymIO.emptyInitialFileSystemContents
+      Just fsRoot -> SymIO.Loader.loadInitialFiles sym fsRoot
+  CLLVM.SymIO.initialLLVMFileSystem halloc sym ptrW fileContents [] C.emptyGlobals
```
langston-barrett: Consider naming the `[]` so it's more obvious what it represents

langston-barrett (approved):
LGTM with a little additional documentation

grease-cli/src/Grease/Main.hs
RyanGlScott: This was a leftover from the type signature for `crucible-llvm`'s [`initialLLVMFileSystem`](https://github.com/GaloisInc/crucible/blob/9cb4f93d17c48890a1643a005120b365e6873c56/crucible-llvm/src/Lang/Crucible/LLVM/SymIO.hs#L180-L183) function, which has both a `HasPtrWidth` constraint and an explicit `NatRepr` argument. You're absolute right, though–it's redundant to have both, and I'm not sure why `crucible-llvm` does things this way. I've removed the `NatRepr` argument from the `grease-cli` function that calls `initialLLVMFileSystem`.

langston-barrett: https://github.com/GaloisInc/crucible/issues/1394

PR #156, opened 2025-05-08
doc: Split section on the test-suite into its own page

langston-barrett:
`dev.md` was getting a bit long, and this section had plenty of sub-sections and content. Enough for its own page, I'd say!


doc/dev.md
```diff
@@ -116,115 +116,4 @@ residing in its own top-level directory:
 * `grease-cli`: This defines a command-line application on top of the library
   code in `grease`.
 
-## Test suite
```
RyanGlScott, line 119: Is it worth including a forward reference to `doc/tests.md` here?

RyanGlScott (approved):

doc/dev.md
langston-barrett: Added under "building", as the normal sequence is build -> test!

PR #159, opened 2025-05-13
Swap gitrev with githash

### LINKED ISSUE #116: Consider migrating away from `gitrev`
[`gitrev` is unmaintained](https://github.com/acfoltzer/gitrev), and doesn't work in a variety of situations. We might want to consider migrating to [`githash`](https://github.com/snoyberg/githash#readme) or  [`gitrev-typed`](https://hackage.haskell.org/package/gitrev-typed-0.1/candidate) when it is released (see also [this discussion](https://discourse.haskell.org/t/resurrecting-gitrev/11933)).

willbasky:
Resolve #116 

I picked up `githash` to avoid raising bound for dependencies
( `gitrev-typed` draws new `filepath`).

### DIFF
--- a/grease/src/Grease/GitRev.hs
+++ b/grease/src/Grease/GitRev.hs
@@ -1,24 +1,29 @@
-{-|
-Copyright        : (c) Galois, Inc. 2024
-Maintainer       : GREASE Maintainers <grease@galois.com>
--}
-
 {-# LANGUAGE TemplateHaskell #-}
 
--- | These are placed in their own module to minimize the cost of recompilation
+-- \| These are placed in their own module to minimize the cost of recompilation
 -- due to Template Haskell.
+
+-- |
+-- Copyright        : (c) Galois, Inc. 2024
+-- Maintainer       : GREASE Maintainers <grease@galois.com>
 module Grease.GitRev (hash, branch, dirty) where
 
-import GitHash (GitInfo, tGitInfoCwd, giHash, giBranch, giDirty)
+import GitHash (GitInfo, giBranch, giDirty, giHash, tGitInfoCwdTry)
 
-gitInfo :: GitInfo
-gitInfo = $$tGitInfoCwd
+gitInfo :: Either String GitInfo
+gitInfo = $$tGitInfoCwdTry
 
 hash :: String
-hash = giHash gitInfo
+hash = case gitInfo of
+    Left _ -> "UNKNOWN"
+    Right gi -> giHash gi
 
 branch :: String
-branch = giBranch gitInfo
+branch = case gitInfo of
+    Left _ -> "UNKNOWN"
+    Right gi -> giBranch gi
 
 dirty :: Bool
-dirty = giDirty gitInfo
+dirty = case gitInfo of
+  Left _ -> False
+  Right gi -> giDirty gi
--- a/grease/src/Grease/GitRev.hs
+++ b/grease/src/Grease/GitRev.hs
@@ -1,10 +1,11 @@
+{-|
+Copyright        : (c) Galois, Inc. 2024
+Maintainer       : GREASE Maintainers <grease@galois.com>
+-}
+
 {-# LANGUAGE TemplateHaskell #-}
 
--- |
--- Copyright        : (c) Galois, Inc. 2024
--- Maintainer       : GREASE Maintainers <grease@galois.com>
---
--- These are placed in their own module to minimize the cost of recompilation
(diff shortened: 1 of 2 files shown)


grease/src/Grease/GitRev.hs
```diff
@@ -9,13 +9,16 @@ Maintainer       : GREASE Maintainers <grease@galois.com>
 -- due to Template Haskell.
 module Grease.GitRev (hash, branch, dirty) where
 
-import Development.GitRev (gitBranch, gitDirty, gitHash)
+import GitHash (GitInfo, tGitInfoCwd, giHash, giBranch, giDirty)
+
+gitInfo :: GitInfo
+gitInfo = $$tGitInfoCwd
```
RyanGlScott: There is a subtle change in behavior between `gitrev` and `githash` here that we will need to address before we can merge this. Namely, we need to be able to build `grease` in contexts where the `.git` information is not available (e.g., the Docker image). `gitrev`'s behavior is to gracefully fail when `.git` information is not available, e.g.,

```
λ> import Development.GitRev
λ> :set -XTemplateHaskell
λ> putStrLn $gitHash
UNKNOWN
```

The behavior of `githash`'s `tGitInfoCwd` function, on the other hand, is to throw an exception when there is no `.git` information:

```
λ> import GitHash
λ> :set -XTemplateHaskell
λ> putStrLn (giHash $$tGitInfoCwd)

<interactive>:6:20: error:
    • Exception when trying to run compile-time code:
        GHEGitRunFailed "." ["rev-parse","--show-toplevel"] (ExitFailure 128) "" "fatal: not a git repository (or any of the parent directories): .git\n"
      Code: tGitInfoCwd
    • In the Template Haskell splice $$tGitInfoCwd
      In the first argument of ‘giHash’, namely ‘$$tGitInfoCwd’
      In the first argument of ‘putStrLn’, namely
        ‘(giHash $$tGitInfoCwd)’
```

I haven't test this, but I believe this means that the Dockerfile wouldn't build with these changes.

-----

To fix this, I think we will want to use [`tGitInfoCwdTry`](https://hackage.haskell.org/package/githash-0.1.7.0/docs/GitHash.html#v:tGitInfoCwdTry) instead. This returns an `Either` instead of throwing an exception upon failure. That we, we can do something like this:

```hs
gitInfo :: Either String GitInfo
gitInfo = $$tGitInfoCwdTry

hash :: String
hash =
  case gitInfo of
    Left _ -> "UNKNOWN"
    Right gi -> giHash gi
```

And similarly for `branch` and `dirty`.

willbasky: In case of `giDirty` we can't use `String` in Left
```haskell
dirty :: Bool
dirty = case gitInfo of
  Left _ -> "UNKNOWN"
  Right gi -> giDirty gi
```
True or False then?

RyanGlScott: `githash`'s [`gitDirty`](https://hackage.haskell.org/package/gitrev-1.3.1/docs/Development-GitRev.html#v:gitDirty) function conservatively returns `False` if there is no `.git` information, so I think we should follow suit and use `False` as well.

willbasky:
btw, we could make 'dirty' to return `Maybe Bool` and `verStr`  handles `commitDirty` 
```haskell 
dirtyLab = case commitDirty of
      Nothing -> "UNKNOWN" 
      Just isDirty -> if isDataFamily
        then " (non-committed files present during build)"
        else ""
```

grease/src/Grease/GitRev.hs
```diff
@@ -1,21 +1,29 @@
...
-
 {-# LANGUAGE TemplateHaskell #-}
 
--- | These are placed in their own module to minimize the cost of recompilation
+-- \| These are placed in their own module to minimize the cost of recompilation
 -- due to Template Haskell.
+
+-- |
+-- Copyright        : (c) Galois, Inc. 2024
+-- Maintainer       : GREASE Maintainers <grease@galois.com>
```
RyanGlScott: The formatting here is a bit weird. How about:

```suggestion
-- |
-- Copyright        : (c) Galois, Inc. 2024
-- Maintainer       : GREASE Maintainers <grease@galois.com>
--
-- These are placed in their own module to minimize the cost of recompilation
-- due to Template Haskell.
```

willbasky: Looks like I accidentally touch comments. I revert them to original state.

RyanGlScott (approved):
Thanks again!

PR #172, opened 2025-05-16
Concretize arguments for bugs, concretization for LLVM

### LINKED ISSUE #114: llvm: Concretization
GREASE provides concrete register values when heuristics fail for binaries:
```
cabal run exe:grease -- --symbol test tests/refine/neg/write_const_global/test.armv7l.elf

[2025-04-30 15:21:46] Analyzing from entrypoint 'test' 
[2025-04-30 15:21:46] Warning: no annotation on predicate!
[2025-04-30 15:21:46] Unable to find a heuristic for any goal
[2025-04-30 15:21:46] Finished analyzing 'test'. Possible bug(s):

At 0x1008c:
PointerWrite outside of static memory range (known BlockID 0): 0x100d4:[32]
Concretized arguments:

R0: 00000000
R1: 00000000
R2: 00000000
R3: 00000000
```
However, it does not do so for LLVM bitcode nor S-expression programs.
```
cabal run exe:grease -- --symbol test --error-symbolic-fun-calls tests/llvm/func-ptr-error.llvm.cbl 

[2025-04-30 15:27:12] Analyzing from entrypoint 'test' 
[2025-04-30 15:27:12] Heuristic: grow and initialize memory referenced by: %0
[2025-04-30 15:27:12] Unable to find a heuristic for any goal
[2025-04-30 15:27:12] Finished analyzing 'test'. Possible bug(s):

At tests/llvm/func-ptr-error.llvm.cbl:14:12:
Failed to load function handle
Details:
  Cannot resolve a symbolic pointer to a function handle
```
Fixing this would require supplying a non-empty string to `toFailedPredicate` here:

https://github.com/GaloisInc/grease/blob/84d8858f1c37ab3e5e3c0b108208f2be5b7e4286/grease-cli/src/Grease/Main.hs#L1204

as is done for Macaw CFGs here:

https://github.com/GaloisInc/grease/blob/84d8858f1c37ab3e5e3c0b108208f2be5b7e4286/grease-cli/src/Grease/Main.hs#L732

langston-barrett:
Fixes #114.

### DIFF
--- a/grease-cli/src/Grease/Main.hs
+++ b/grease-cli/src/Grease/Main.hs
@@ -99,7 +99,7 @@ import Grease.AssertProperty
 import Grease.BranchTracer (greaseBranchTracerFeature)
 import Grease.Bug qualified as Bug
 import Grease.Cli (optsFromArgs)
-import Grease.Concretize (ConcArgs(..), concArgs, printConcArgs)
+import Grease.Concretize (ConcretizedData, ConcArgs(..), concArgs, printConcArgs)
 import Grease.Concretize.JSON (concArgsToJson)
 import Grease.Cursor.Pointer ()
 import Grease.Diagnostic
@@ -164,6 +164,8 @@ import Lang.Crucible.LLVM.MemModel.Partial qualified as Mem
 import Lang.Crucible.LLVM.SymIO qualified as CLLVM.SymIO
 import Lang.Crucible.LLVM.Syntax (llvmParserHooks, emptyParserHooks)
 import Lang.Crucible.LLVM.Translation qualified as Trans
+import Lang.Crucible.SymIO qualified as SymIO
+import Lang.Crucible.SymIO.Loader qualified as SymIO.Loader
 import Lang.Crucible.LLVM.TypeContext qualified as TCtx
 import Lang.Crucible.Simulator qualified as C
 import Lang.Crucible.Simulator.SimError qualified as C
@@ -172,6 +174,7 @@ import Lang.Crucible.Syntax.Prog qualified as CSyn
 import Lumberjack qualified as LJ
 import Prelude (Num(..), fromIntegral, Integral)
 import Prettyprinter qualified as PP
+import Prettyprinter.Render.Text qualified as PP
 import Stubs.FunctionOverride qualified as Stubs
 import System.Directory (Permissions, getPermissions)
 import System.FilePath (FilePath)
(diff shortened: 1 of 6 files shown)


grease-cli/src/Grease/Main.hs
```diff
@@ -342,7 +342,7 @@ checkMustFail bak errs = do
         , Bug.bugUb = Nothing  -- TODO: check if they are all the same?
         }
   if mustFail
-  then pure (Just (MkBatchBug bug []))  -- TODO: concretized args
+  then pure (Just (MkBatchBug bug [] ""))  -- TODO: concretized args
```
RyanGlScott: Consider using `MkMatchBug`'s field selectors here to make it more obvious what each argument is.

grease-cli/src/Grease/Main.hs
```diff
@@ -429,9 +429,10 @@ interestingConcretizedShapes ::
 interestingConcretizedShapes names initArgs (ConcArgs cArgs) =
   let cShapes = fmapFC (Shape.tagWithType . concShape) cArgs in
   let initArgs' = fmapFC Shape.tagWithType initArgs in
+  let isLlvmArg name = "%" `List.isPrefixOf` name in
   Ctx.zipWith
     (\(Const name) (Const isDefault) ->
-      Const (name `List.elem` interestingRegs && not isDefault))
+      Const ((name `List.elem` interestingRegs || isLlvmArg name) && not isDefault))
```
RyanGlScott, line 469: We have now changed the meaning of what it means for an argument to be "interesting", so please update the relevant part of the Haddocks above to reflect this.

grease-cli/src/Grease/Main.hs
```diff
@@ -1102,9 +1106,11 @@ simulateLlvmCfg la simOpts bak fm halloc llvmCtx initMem setupHook mbStartupOvCf
     execAndRefine bak (simSolver simOpts) fm la setupAnns initMem heuristics argNames argShapes args bbMapRef (simLoopBound simOpts) execFeats st
 
   res <- case result of
-    RefinementBug b _ -> do
-      let argsJson = []  -- TODO
-      pure (BatchBug (MkBatchBug b argsJson))
+    RefinementBug b cData -> do
+      let argsJson = concArgsToJson fm argNames (concArgs cData) argTys
+      let interestingShapes = interestingConcretizedShapes argNames (initArgs ^. Shape.argShapes) (concArgs cData)
+      let prettyArgs = printConcArgs MM.Addr64 argNames interestingShapes (concArgs cData)
```
RyanGlScott: Now that we are calling `printConcArgs` from both the LLVM and Macaw frontends, it is a little strange that we have to pass `Addr64` (a `macaw`-specific value) in an LLVM-specific context. I propose that we change `printConcArgs` to take a `NatRepr` (or some other data type not from `macaw`) instead of an `AddrWidthRepr`.

langston-barrett: The difficulty here is that this value gets passed to the shape DSL printer, which only supports specifically 32- or 64-bit addresses. I agree it's a bit strange, but `AddrWidthRepr` is a convenient type for expressing this constraint. Do you think it's worth making our own version of it?

RyanGlScott: Fair enough. Given that we're already depending on `macaw` anyway, I suppose it's not the worst thing in the world to use `AddrWidthRepr` for this purpose. I retract my suggestion.

Issue #176, opened 2025-05-19
`grease` fails to resolve auxiliary function defined in Macaw S-expression program


RyanGlScott:
Consider this Macaw S-expression program:

```lisp
; id.x64.cbl
(declare @abort () Unit)

(defun @vuln () Unit
  (start start:
    (funcall @abort)
    (return ())))

(defun @test ((regs X86Regs)) X86Regs
  (start start:
    (funcall @vuln)
    (return regs)))
```

I would expect `grease` to report the call to `abort` in `vuln` as a bug. Instead, `grease` fails in an unexpected way:

```
$ cabal -v0 run exe:grease -- id.x64.cbl --symbol test
[2025-05-19 13:22:38] Analyzing from entrypoint 'test'
grease: Could not resolve function: vuln
Called at: id.x64.cbl:11:5
In test at id.x64.cbl:11:5
```

What's more, this `Could not resolve function` failure only occurs with Macaw S-expression programs. This does _not_ occur with LLVM S-expression programs such as this one:

```lisp
; id.llvm.cbl
(declare @abort () Unit)

(defun @vuln () Unit
  (start start:
    (funcall @abort)
    (return ())))

(defun @test () Unit
  (start start:
    (funcall @vuln)
    (return ())))
```
```
$ cabal -v0 run exe:grease -- id.llvm.cbl --symbol test
[2025-05-19 13:23:53] Analyzing from entrypoint 'test'
[2025-05-19 13:23:53] Warning: no annotation on predicate!
[2025-05-19 13:23:53] Finished analyzing 'test'. Likely bug: unavoidable error (safety condition is unsatisfiable) at id.llvm.cbl:6:5
id.llvm.cbl:6:5: error: in vuln
Call to abort
```

The discrepancy can be explained in the `SetupHook`s used for each frontend. For LLVM S-expression programs, we have:

https://github.com/GaloisInc/grease/blob/8ae41c3507e090f23135634514c28de43eddb0f1/grease-cli/src/Grease/Main.hs#L1273-L1286

This iterates over every function CFG defined in the LLVM S-expression program and binds the CFG's handle to its implementation. Nothing comparable happens in the `SetupHook` for Macaw S-expression programs, which means that the only CFG that gets bound is the entrypoint function (`test`).

PR #177, opened 2025-05-19
`grease-cli`: Resolve auxiliary functions in Macaw S-expression programs

RyanGlScott:
This ensures that every function in a Macaw S-expression program (both the entrypoint functions and the auxiliary functions) have their function handles properly registered in the `SetupHook` for `simulateMacawSyntax`.

Fixes https://github.com/GaloisInc/grease/issues/176.

### DIFF
--- a/grease-cli/src/Grease/Main.hs
+++ b/grease-cli/src/Grease/Main.hs
@@ -940,6 +940,13 @@ simulateMacawSyntax la halloc archCtx simOpts parserHooks = do
         forM_ (Map.elems cfgs) $ \entrypointCfgs ->
           forM_ (startupOvForwardDecs <$> entrypointStartupOv entrypointCfgs) $ \startupOvFwdDecs ->
             Macaw.registerMacawOvForwardDeclarations bak funOvs startupOvFwdDecs
+
+        -- Register defined functions.
+        forM_ (CSyn.parsedProgCFGs prog) $ \(C.Reg.AnyCFG defCfg) -> do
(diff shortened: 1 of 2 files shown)


grease-cli/src/Grease/Main.hs
```diff
@@ -941,6 +940,13 @@ simulateMacawSyntax la halloc archCtx simOpts parserHooks = do
         forM_ (Map.elems cfgs) $ \entrypointCfgs ->
           forM_ (startupOvForwardDecs <$> entrypointStartupOv entrypointCfgs) $ \startupOvFwdDecs ->
             Macaw.registerMacawOvForwardDeclarations bak funOvs startupOvFwdDecs
+
+        -- Register defined functions.
+        forM_ (CSyn.parsedProgCFGs prog) $ \(C.Reg.AnyCFG defCfg) -> do
+          C.SomeCFG defSsa <- pure $ C.toSSA defCfg
+          -- This could probably be a helper defined in Crucible...
+          let bindCfg c = C.bindFnHandle (C.cfgHandle c) (C.UseCFG c (C.postdomInfo c))
+          bindCfg defSsa
```
RyanGlScott, line 949: This code isn't quite right, since really we should be checking if each function has a corresponding function override and doing something different if an override exists. I briefly looked into this, but it ended up being more involved than I expected, partly because we would have to change `stubs` in order to make this possible. See #178.

I don't particularly need the ability to override functions defined in an S-expression program just yet, so I am going to accept this deficiency for now, leaving #178 as future work.

PR #186, opened 2025-05-21
`ResolveCall`: Add `Lookup{FunctionHandle,Syscall}Dispatch`

### LINKED ISSUE #179: `Grease.Macaw.ResolveCall`: Allow overriding behavior for skipped PLT stub calls
Currently, `Grease.Macaw.ResolveCall.lookupFunctionHandle` makes a somewhat opinionated choice: any invocation of a PLT stub for which there is not a corresponding override should be skipped. This is a reasonable assumption for `grease-cli`, which has no notion of shared libraries (and thus can't really do much better besides skipping), but downstream `grease` applications may want to do something different. We should make this part of the `grease` API more extensible so that users can choose how to handle this sort of scenario.

RyanGlScott:
This makes the API in `Grease.Macaw.ResolveCall` more extensible by parameterizing the behavior of the `lookupFunctionHandle` and `lookupSyscallHandle` functions by continuations that allow the user to control how functions or syscalls are simulated. The key data types are:

* `Lookup{FunctionHandle,Syscall}Dispatch`: newtypes around the continuations themselves
* `Lookup{FunctionHandle,Syscall}Result`: a data type that captures the result of resolving the function/syscall

Fixes #179.

### DIFF
--- a/grease/src/Grease/Macaw/ResolveCall.hs
+++ b/grease/src/Grease/Macaw/ResolveCall.hs
@@ -12,6 +12,9 @@ Maintainer       : GREASE Maintainers <grease@galois.com>
 module Grease.Macaw.ResolveCall
   ( -- * Looking up function handles
     lookupFunctionHandle
+  , LookupFunctionHandleDispatch(..)
+  , defaultLookupFunctionHandleDispatch
+  , LookupFunctionHandleResult(..)
 
     -- * Looking up syscall handles
   , lookupSyscallHandle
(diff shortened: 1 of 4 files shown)


grease/src/Grease/Macaw/ResolveCall.hs
```diff
@@ -124,8 +245,11 @@ lookupFunctionHandle ::
   -- | Map of names of overridden functions to their implementations
   Map.Map W4.FunctionName (MacawFunctionOverride p sym arch) ->
   ErrorSymbolicFunCalls ->
+  LookupFunctionHandleDispatch p sym arch ->
   Symbolic.LookupFunctionHandle p sym arch
-lookupFunctionHandle bak la halloc arch memory symMap pltStubs dynFunMap funOvs errorSymbolicFunCalls = Symbolic.LookupFunctionHandle $ \st mem regs -> do
+lookupFunctionHandle bak la halloc arch memory symMap pltStubs dynFunMap funOvs errorSymbolicFunCalls (LookupFunctionHandleDispatch dispatch) = Symbolic.LookupFunctionHandle $ \st mem regs -> do
+  let dispatch' st' = dispatch st' mem regs
```
langston-barrett: This line is quite long already, perhaps:
```suggestion
lookupFunctionHandle bak la halloc arch memory symMap pltStubs dynFunMap funOvs errorSymbolicFunCalls lfhd = Symbolic.LookupFunctionHandle $ \st mem regs -> do
  let LookupFunctionHandleDispatch dispatch = lfhd
  let dispatch' st' = dispatch st' mem regs
```

grease/src/Grease/Macaw/ResolveCall.hs
```diff
@@ -124,8 +245,11 @@ lookupFunctionHandle ::
   -- | Map of names of overridden functions to their implementations
   Map.Map W4.FunctionName (MacawFunctionOverride p sym arch) ->
   ErrorSymbolicFunCalls ->
+  LookupFunctionHandleDispatch p sym arch ->
```
langston-barrett: It looks like this function mostly computes a `LookupFunctionHandleResult`, and then applies `dispatch` to it. Perhaps it would be more clear to factor out a function that explicitly returns a LFHR, and replace the body of this function with something that just calls that one and applies `dispatch` to the result?

grease/src/Grease/Macaw/ResolveCall.hs
```diff
@@ -228,19 +338,92 @@ lookupFunctionHandle bak la halloc arch memory symMap pltStubs dynFunMap funOvs
...
+    Int ->
+    -- | The syscall's override.
+    Stubs.SomeSyscall p sym (Symbolic.MacawExt arch) ->
+    LookupSyscallResult p sym arch atps rtps
+
 -- | An implementation of 'Symbolic.LookupSyscallHandle' that attempts to look
--- up an override for the syscall, and if one is found, invokes it. If one
--- cannot be found, the syscall is simulated as a no-op.
+-- up an syscall override and dispatches on the result.
 lookupSyscallHandle ::
```
langston-barrett, line 496: Similar refactoring suggestion (compute result, call `dispatch` on it) applies here.

grease/src/Grease/Macaw/SkippedCall.hs
```diff
@@ -17,32 +20,35 @@ import What4.FunctionName qualified as W4
...
     W4.FunctionName ->
-    SkippedCall arch
+    SkippedFunctionCall arch
   NotExecutable ::
     (1 W4.<= MC.ArchAddrWidth arch, MemWidth (MC.ArchAddrWidth arch)) =>
     MemSegmentOff (MC.ArchAddrWidth arch) ->
-    SkippedCall arch
+    SkippedFunctionCall arch
+
+-- | The reasons that GREASE might skip a function call
```
langston-barrett:
```suggestion
-- | The reasons that GREASE might skip a syscall
```

langston-barrett (approved):

grease/src/Grease/Macaw/ResolveCall.hs
```diff
@@ -170,31 +294,16 @@ lookupFunctionHandle bak la halloc arch memory symMap pltStubs dynFunMap funOvs
...
-          -- Log that we have performed a function call.
-          logFunctionCall fnName = do
-            mbReturnAddr <-
-              (arch ^. archFunctionReturnAddr) bak (arch ^. archVals) regs mem
-            doLog la $ Diag.FunctionCall fnName bvMemWord mbReturnAddr
-
           hdls = st ^. stateDiscoveredFnHandles
           bvMemWord = EL.memWord bvWord64
 
           dispatchFuncAddrOff funcAddrOff
```
langston-barrett: While we're here: I find this definition of a long function inside a `let` a little confusing. First of all, it's just hard to parse because it's long and has a bunch of cases. Secondly, it puts the cases that are considered later earlier in the source code than the case that is considered first (namely, `InvalidAddress`). I would like to factor it out into its own function. Feel free to do this, or just let me know if you like the idea and I can do it in a follow-up.

RyanGlScott: I am not entirely convinced that the code will be much cleaner if we did factor out this function, since it closes over so many local values. If you did factor it out, it would end up having this type signature:

```hs
dispatchFuncAddrOff ::
  ( Symbolic.SymArchConstraints arch
  , HasGreaseSimulatorState p sym arch
  ) =>
  GreaseLogAction ->
  C.HandleAllocator ->
  ArchContext arch ->
  EL.Memory (MC.ArchAddrWidth arch) ->
  Discovery.AddrSymMap (MC.ArchAddrWidth arch) ->
  Map.Map (MC.ArchSegmentOff arch) W4.FunctionName ->
  Map.Map W4.FunctionName (MC.ArchSegmentOff arch) ->
  Map.Map W4.FunctionName (MacawFunctionOverride p sym arch) ->
  ( C.CrucibleState p sym (Symbolic.MacawExt arch) rtp blocks r ctx ->
    LookupFunctionHandleResult p sym arch ->
    IO ( C.FnHandle (Ctx.EmptyCtx Ctx.::> Symbolic.ArchRegStruct arch) (Symbolic.ArchRegStruct arch)
       , C.CrucibleState p sym (Symbolic.MacawExt arch) rtp blocks r ctx
       )) ->
  MC.ArchSegmentOff arch ->
  C.CrucibleState p sym (Symbolic.MacawExt arch) rtp blocks r ctx ->
  IO ( C.FnHandle (Ctx.EmptyCtx Ctx.::> Symbolic.ArchRegStruct arch) (Symbolic.ArchRegStruct arch)
     , C.CrucibleState p sym (Symbolic.MacawExt arch) rtp blocks r ctx
     )
```

Which doesn't add a lot of clarity, in my opinion. Perhaps you see a way to make this cleaner? If so, please file an issue :)

RyanGlScott: I ended up renaming this function to `go` (to reflect the fact that it is a recursive helper function) and converted it from a `let`-bound function to a `where`-bound function, which I think improves the readability of the overall function at least a little bit. Let me know what you think.

langston-barrett: I think that's much nicer, thank you! Sorry, didn't mean to hold up merge on this comment - I really meant I would be happy to do it myself :)

PR #188, opened 2025-05-23
Add tests for symbolic I/O overrides

### LINKED ISSUE #152: Add tests for symio-related overrides (`read`, `write`, `open`, and `close`)
#151 adds built-in overrides for symio-related functions (`read`, `write`, `open`, and `close`), but stops short of adding tests. We should do so.

Ideally, we would implement these as S-expression tests, but doing so proves non-trivial. For example, consider [this test case](https://github.com/GaloisInc/ambient-verifier/blob/eab04abb9750825a25ec0cbe0379add63f05f6c6/tests/binaries/x86_64/open-close.c) (adapted from `ambient-verifier`), which is written in C:

```c
int main(void) {
  char c;
  const char path[] = "/open-me.txt";
  long fd = open(path, O_RDONLY);
  if (fd >= 0) {
    long read_bytes = read(fd, &c, 1);
    assert(c == 'A');
    assert(read_bytes == 1);
    close(fd);
  }
  return (int)c;
}
```

How should we port the call to `open(...)` to an S-expression? In order to do so, we would need to write the string `"/open-me.txt"` to a pointer. But this is a _C-style_ string, not a `crucible-syntax` `String` value, and at the moment, it is not possible to convert from the latter to the former. (See https://github.com/GaloisInc/crucible/issues/1393.) The alternative would be to write to the pointer one character at a time, but this would be incredibly verbose, even by `crucible-syntax` standards.

https://github.com/GaloisInc/crucible/issues/1393 proposes adding new `crucible-llvm-syntax` primitives that would make it possible to port this test case to an S-expression program in a relatively succinct way. Once https://github.com/GaloisInc/crucible/issues/1393 is implemented, we should revisit this issue.

langston-barrett:
Fixes #152.


langston-barrett:
Actually, I think I've thought of a better workaround. These overrides are only called from S-expression files, so we can avoid going through the entirety of the GREASE override machinery and just bind them to forward declarations (if the user hasn't *further* overridden them) here:

https://github.com/GaloisInc/grease/blob/2b2e3ad3de85c291038be34620976eb6054c872a/grease/src/Grease/LLVM/Overrides.hs#L488

grease/src/Grease/LLVM/Overrides/Declare.hs
```diff
@@ -100,28 +96,3 @@ mkDeclare name args ret = do
...
-      CLLVM.SomeLLVMOverride $
-        CLLVM.LLVMOverride
-          { CLLVM.llvmOverride_declare = decl
-          , CLLVM.llvmOverride_args = argTys
-          , CLLVM.llvmOverride_ret = retTy
-          , CLLVM.llvmOverride_def =
-              \_mvar args -> C.typedOverrideHandler ov (TFC.fmapFC (C.RV . C.regValue) args)
-          }
-    Left err ->
-      panic ("Bad override for `" <> nm <> "`: " <> Text.unpack err) []
```
RyanGlScott: There's quite a bit of churn in the commit history involving `typedOverrideToSomeLLVMOverride`, only for it to be removed in the end. Worth rewriting the git commit history to avoid this churn?

grease/src/Grease/LLVM/Overrides.hs
```diff
@@ -394,9 +404,39 @@ registerLLVMForwardDeclarations ::
...
+            unless ok (cannotResolve fwdDecName hdl)
+          "write-bytes" -> do
+            ok <- tryBindTypedOverride hdl (StrOv.writeBytesOverride mvar)
+            unless ok (cannotResolve fwdDecName hdl)
+          "write-c-string" -> do
+            ok <- tryBindTypedOverride hdl (StrOv.writeCStringOverride  mvar)
+            unless ok (cannotResolve fwdDecName hdl)
+          _ -> cannotResolve fwdDecName hdl
+
+tryBindTypedOverride ::
```
RyanGlScott, line 428: Include a comment saying what the `Bool` return type indicates.

doc/builtins.md
```diff
@@ -201,6 +201,18 @@ available:
...
 
+### S-expression-specific overrides
+
+For LLVM S-expression files (programs or overrides), the following overrides are also available:
+```
+(declare @read-bytes ((x (Ptr 64))) (Vector (Bitvector 8)))
+(declare @read-c-string ((x (Ptr 64))) (String Unicode))
+(declare @write-bytes ((dest (Ptr 64)) (src (Vector (Bitvector 8)))) Unit)
+(declare @write-c-string ((dst (Ptr 64)) (src (String Unicode))) Unit)
+```
```
RyanGlScott, line 212: What about Macaw S-expression files? Are those intended to be supported, or are they unsupported for some reason? (If it's the former, it would be worth adding a test case.)

grease/src/Grease/LLVM/Overrides/Declare.hs
langston-barrett: I don't personally think so? But I can do so if you feel more strongly about it.

doc/builtins.md
langston-barrett: Currently unsupported, could probably be supported. How about I file a follow-up?

langston-barrett: I do think it would be considerably more work to support them in Macaw S-expression files, see https://github.com/GaloisInc/grease/issues/190

RyanGlScott (approved):

PR #195, opened 2025-05-28
Concretization of the symbolic filesystem

### LINKED ISSUE #184: Concretization for I/O overrides
GREASE currently provides concretized values for registers and memory when it finds a likely bug. We should do the same for symbolic bytes used by I/O overrides. It would be nice to sort out our testing story before doing this (#152).

langston-barrett:
Fixes #184.

Needs:

- [x] Printing the concretized filesystem
- [x] Test

### DIFF
--- a/grease-cli/tests/llvm/conc-fs.llvm.cbl
+++ b/grease-cli/tests/llvm/conc-fs.llvm.cbl
@@ -1,6 +1,6 @@
 ; Copyright (c) Galois, Inc. 2025
 
-; Ensure that `--fs-root` and the I/O overrides work.
+; Ensure that the IO overrides work.
 
 ;; flags {"--symbol", "test"}
 ;; flags {"--fs-root", "tests/llvm/extra/fs"}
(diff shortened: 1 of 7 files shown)


grease/src/Grease/Concretize.hs
```diff
@@ -87,9 +98,22 @@ concArgsToSym sym fm argTys (ConcArgs cArgs) =
 -- the future.
 newtype ConcMem sym = ConcMem { getConcMem :: Mem.MemImpl sym }
 
+-- | File system contents before execution ('SymIO.InitialFileSystemContents')
+-- that has been concretized
+data ConcFs = ConcFs { getConcFs :: Map (SymIO.FDTarget SymIO.In) [Word8] }
```
RyanGlScott: Make this a newtype?

grease-cli/tests/llvm/conc-fs.llvm.cbl
```diff
@@ -0,0 +1,49 @@
...
+    (let h (load-handle Unit () g))
+    (funcall h)
+    (return ()))
+  (defblock else:
+    (return ())))
+
+;; check [[
+;; Concretized filesystem:
+;; /symb
+;; 00
```
RyanGlScott: What happens if there are multiple files in the filesystem? Would it print something like this?

```
Concretized filesystem:
/symb
00
/symb/foo
11
```

I wonder if we should indent this a bit more to make it obvious which things are the contents of specific files. Something like this, perhaps:

```
Concretized filesystem:
  /symb
    00
  /symb/foo
    11
```

Or, alternatively:

```
Concretized filesystem:
* /symb
  00
* /symb/foo
  11
```

grease-cli/tests/llvm/io-overrides-fs.llvm.cbl
```diff
@@ -0,0 +1,46 @@
...
+
+(defun @test () Unit
+  (start start:
+
+    (let p (alloca none (bv 64 6)))
+    (funcall @write-c-string p "/symb")
+
+    ; int open(const char *pathname, int flags, ... /* mode_t mode */ );
+    (let open-global (resolve-global "open"))
+    (let open (load-handle (Ptr 32) ((Ptr 64) (Ptr 32)) open-global))
```
RyanGlScott: Silly question: instead of manually creating function handles using `load-handle`, what if we `declare`d functions like `open` in this test case? Then the type signatures would be evident from their forward declarations, so we wouldn't need to include them as comments.

grease-cli/tests/llvm/conc-fs.llvm.cbl
langston-barrett: Good call, done

RyanGlScott (approved):

grease-cli/tests/llvm/io-overrides-fs.llvm.cbl
langston-barrett: Not silly! I've changed the tests to use `declare`. I kept the C-style type signatures because they have source-level types, which makes it easier to remember what the semantics of the various arguments and return values are.

PR #202, opened 2025-06-04
Port `fresh-vec` built-in from ambient-verifier

### LINKED ISSUE #201: Allow S-expression overrides to create concretizable bytes
It would be nice if S-expression overrides could create fresh bytes that would later be fed through the concretization machinery. This would be sufficient for MVP versions of `read` and `recv` while we work on a more fleshed-out solution (#200), and would generally be a nice feature to have. The API might look like:
```
(declare @fresh-bytes ((name (String Unicode)) (num Integer)) (Vector (Bitvector 8)))
```
Perhaps someday, we could make [the `fresh` built-in](https://github.com/GaloisInc/crucible/blob/6a601740355b2384dbf4ecd37073ff0a144a8f82/crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs#L1587) configurable, so that we could track the symbolic values it creates.

langston-barrett:
Port `fresh-vec` from `ambient-verifier`.

### DIFF
--- a/doc/sexp.md
+++ b/doc/sexp.md
@@ -18,6 +18,17 @@ additional types and operations, see:
 [macaw-x86-syntax]: https://github.com/GaloisInc/macaw/tree/master/macaw-x86-syntax
 [crucible-llvm-syntax]: https://github.com/GaloisInc/crucible/tree/master/crucible-llvm-syntax
 
+## Additional operations
+
+GREASE provides the following operations in S-expression files, in addition to
+the operations provided by the base S-expression language:
+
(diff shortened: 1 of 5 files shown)


langston-barrett:
Hmm, upon further consideration I've realized that `fresh-vec` is likely not the ideal starting point for #201, as it generates the *syntax* of a vector constant full of fresh entries. I don't think it's possible to generalize `fresh-vec` to take a dynamic (but concrete) number of entries. I'll leave this open because `fresh-vec` is independently useful, but I'll shift my focus to creating an override rather than a built-in.

langston-barrett:
Closing as somewhat low-priority.

PR #203, opened 2025-06-04
Add `fresh-bytes` override for S-expression CFGs

### LINKED ISSUE #201: Allow S-expression overrides to create concretizable bytes

langston-barrett:
Fixes #201.

- [x] Initial `fresh-bytes` override
- [x] Thread fresh values through to concretization

### DIFF
--- a/grease-cli/src/Grease/Main.hs
+++ b/grease-cli/src/Grease/Main.hs
@@ -102,6 +102,7 @@ import Grease.Bug qualified as Bug
 import Grease.Cli (optsFromArgs)
 import Grease.Concretize (ConcretizedData, ConcArgs(..))
 import Grease.Concretize qualified as Conc
+import Grease.Concretize.ToConcretize qualified as ToConc
 import Grease.Concretize.JSON (concArgsToJson)
 import Grease.Cursor.Pointer ()
 import Grease.Diagnostic
@@ -1129,7 +1130,7 @@ simulateLlvmCfg la simOpts bak fm halloc llvmCtx initMem setupHook mbStartupOvCf
 
   profFeatLog <-
     traverse
-      (greaseProfilerFeature @(C.GlobalVar Conc.ToConcretizeType) @sym @CLLVM.LLVM @(C.RegEntry sym ret))
+      (greaseProfilerFeature @(C.GlobalVar ToConc.ToConcretizeType) @sym @CLLVM.LLVM @(C.RegEntry sym ret))
       (simProfileTo simOpts)
 
   C.Refl <-
--- a/grease-cli/src/Grease/Main.hs
+++ b/grease-cli/src/Grease/Main.hs
@@ -1123,7 +1123,7 @@ simulateLlvmCfg la simOpts bak fm halloc llvmCtx initMem setupHook mbStartupOvCf
 
   profFeatLog <-
     traverse
-      (greaseProfilerFeature @() @sym @CLLVM.LLVM @(C.RegEntry sym ret))
+      (greaseProfilerFeature @(Conc.ToConcretize sym) @sym @CLLVM.LLVM @(C.RegEntry sym ret))
       (simProfileTo simOpts)
 
   C.Refl <-
@@ -1150,7 +1150,8 @@ simulateLlvmCfg la simOpts bak fm halloc llvmCtx initMem setupHook mbStartupOvCf
           modifyIORef bbMapRef $ Map.insert ann (callStack, bb)
     let llvmExtImpl = CLLVM.llvmExtensionImpl ?memOpts
     (fs0, fs, globals, initFsOv) <- liftIO $ initialLlvmFileSystem halloc sym simOpts
-    st <- LLVM.initState bak la llvmExtImpl halloc (simErrorSymbolicFunCalls simOpts) setupMem fs globals initFsOv llvmCtx setupHook (argVals args) mbStartupOvCfg scfg
+    let p = Conc.ToConcretize []  -- personality
+    st <- LLVM.initState bak la llvmExtImpl p halloc (simErrorSymbolicFunCalls simOpts) setupMem fs globals initFsOv llvmCtx setupHook (argVals args) mbStartupOvCfg scfg
     let cmdExt = Debug.llvmCommandExt
     debuggerFeat <-
       liftIO $
--- a/grease-cli/src/Grease/Main.hs
+++ b/grease-cli/src/Grease/Main.hs
@@ -171,9 +171,7 @@ import Lang.Crucible.SymIO qualified as SymIO
 import Lang.Crucible.SymIO.Loader qualified as SymIO.Loader
 import Lang.Crucible.LLVM.TypeContext qualified as TCtx
 import Lang.Crucible.Simulator qualified as C
-import Lang.Crucible.Simulator.GlobalState qualified as C
 import Lang.Crucible.Simulator.SimError qualified as C
-import Lang.Crucible.Simulator.SymSequence qualified as C
 import Lang.Crucible.Syntax.Concrete qualified as CSyn
 import Lang.Crucible.Syntax.Prog qualified as CSyn
 import Lumberjack qualified as LJ
@@ -634,8 +632,7 @@ simulateMacawCfg la bak fm halloc macawCfgConfig archCtx simOpts setupHook mbCfg
         -- Grease.Macaw.SimulatorState.) If we are simulating an S-expression program,
         -- use an empty map instead. (See gitlab#118 for more discussion on this point.)
         let discoveredHdls = Maybe.maybe Map.empty (`Map.singleton` ssaCfgHdl) mbCfgAddr
-        toConcVar <- liftIO (C.freshGlobalVar halloc "to-concretize" W4.knownRepr)
-        let globals1 = C.insertGlobal toConcVar C.SymSequenceNil globals0
+        (toConcVar, globals1) <- liftIO $ ToConc.newToConcretize halloc globals0
         let personality =
               emptyGreaseSimulatorState toConcVar &
                 discoveredFnHandles .~ discoveredHdls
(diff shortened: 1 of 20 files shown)


langston-barrett:
CI failures likely due to issues with GHA:
```
Internal server error. Correlation ID: be0f99d6-a2ba-41c9-b3b0-2b4821a25623
```
https://www.githubstatus.com/

langston-barrett:
I think the best path forwards would be to create a Crucible global variable of type
```
SequenceType (StructType (EmptyCtx ::> StringType Unicode ::> AnyType))
```
Where the first component of the struct is a name (generally expected to be concrete), and the second component is the value to be concretized. This sequence will then participate in path merging, and generally be specific to the particular path it is on.

grease-cli/src/Grease/Main.hs
```diff
@@ -631,7 +633,11 @@ simulateMacawCfg la bak fm halloc macawCfgConfig archCtx simOpts setupHook mbCfg
         -- Grease.Macaw.SimulatorState.) If we are simulating an S-expression program,
         -- use an empty map instead. (See gitlab#118 for more discussion on this point.)
         let discoveredHdls = Maybe.maybe Map.empty (`Map.singleton` ssaCfgHdl) mbCfgAddr
-        let personality = emptyGreaseSimulatorState & discoveredFnHandles .~ discoveredHdls
+        toConcVar <- liftIO (C.freshGlobalVar halloc "to-concretize" W4.knownRepr)
+        let globals = C.insertGlobal toConcVar C.SymSequenceNil globals0
```
RyanGlScott: I'd prefer renaming `globals` to `globals1`. Having both `globals` and `globals0` seems like a recipe for confusion.

grease-cli/src/Grease/Main.hs
```diff
@@ -1149,8 +1155,10 @@ simulateLlvmCfg la simOpts bak fm halloc llvmCtx initMem setupHook mbStartupOvCf
     let ?recordLLVMAnnotation = \callStack (Mem.BoolAnn ann) bb ->
           modifyIORef bbMapRef $ Map.insert ann (callStack, bb)
     let llvmExtImpl = CLLVM.llvmExtensionImpl ?memOpts
-    (fs0, fs, globals, initFsOv) <- liftIO $ initialLlvmFileSystem halloc sym simOpts
-    st <- LLVM.initState bak la llvmExtImpl halloc (simErrorSymbolicFunCalls simOpts) setupMem fs globals initFsOv llvmCtx setupHook (argVals args) mbStartupOvCfg scfg
+    (fs0, fs, globals0, initFsOv) <- liftIO $ initialLlvmFileSystem halloc sym simOpts
+    p <- C.freshGlobalVar halloc "to-concretize" W4.knownRepr
+    let globals = C.insertGlobal p C.SymSequenceNil globals0
```
RyanGlScott: Similarly, I'd favor renaming `globals` to `globals1`.

grease/src/Grease/Macaw/SimulatorState.hs
```diff
@@ -42,6 +43,9 @@ data GreaseSimulatorState sym arch = GreaseSimulatorState
     -- function is discovered (see @Note [Incremental code discovery]@), it will
     -- be added to this map so that it can be looked up in future invocations of
     -- that function.
+  , _toConcretize :: C.GlobalVar Conc.ToConcretizeType
+    -- ^ Values created during runtime to be passed to the concretization
+    -- functionality and generally displayed to the user.
```
RyanGlScott: It feels quite strange to store a `GlobalVar` in a Crucible personality type, as the value of this `GlobalVar` never changes throughout simulation. (Only the intrinsic value associated with the `GlobalVar` changes, but that state is managed separately in the simulator.)

It would take some doing, but I wonder if we should store the `GlobalVar` in a read-only thing (either as a function argument or in a reader monad of some sort) that we pass down to `freshBytesOverride`. The upshot is that we wouldn't need the `HasToConcretize` class at all.

grease/src/Grease/Concretize.hs
```diff
@@ -43,33 +58,77 @@ import Lang.Crucible.LLVM.MemModel qualified as Mem
...
+    { initStateArgs :: Args sym ext argTys
+    , initStateFs :: SymIO.InitialFileSystemContents sym
+    , initStateMem :: InitialMem sym
+    }
+
+-- | The type of a global variable containing data to be concretized.
+--
+-- The first component of the struct is a name (generally expected, but not
+-- required, to be concrete), and the second component is the value to be
+-- concretized.
```
RyanGlScott: It might be worth saying that the first component is the `StringType` and the second component is the `AnyType`. If you're not familiar with the convention that `Ctx`s store their components in the opposite order from conventional lists, then calling them "first" and "second" could be counterintuitive.

grease/src/Grease/Concretize.hs
```diff
@@ -123,12 +188,14 @@ makeConcretizedData ::
   OnlineSolverAndBackend solver sym bak t st fm =>
   Mem.HasPtrWidth wptr =>
   (ExtShape ext ~ PtrShape ext wptr) =>
+  (WI.SymExpr sym ~ W4.Expr t) =>
```
RyanGlScott: I'm surprised that this constraint is necessary. Because of the `OnlineSolverAndBackend ... sym ...` constraint, we already know that `sym ~ ExprBuilder ...` and the `SymExpr (ExprBuilder ...)` instance should reduce to `Expr`.

grease/src/Grease/Macaw/SimulatorState.hs
langston-barrett: I guess it feels strange because the personality can be mutated, but this value never is? I could change it to be a function argument as is often done for the LLVM memory global variable. That seems a bit annoying as it is mostly passed along unchanged through override-registration functions. It might be best as an implicit parameter. In that case we could keep `HasToConcretize` just like we have `HasPtrWidth`. I find it tedious to type out the name and type of the implicit parameter each time, and using implicit parameters requires parentheses (as do equality constraints) when they are the only constraint, unlike other kinds of constraints. However, it's (even) more implicit, it makes it harder to tell where the implicit parameter comes from. What do you think?

thinkmoore: This might be worth deferring---it is a design pattern that repeats in several execution features. I agree it's a wart, but I think it might be nicer to use the opportunity to do a separate design and fix for the pattern everywhere it appears.

RyanGlScott: > I could change it to be a function argument as is often done for the LLVM memory global variable.

Indeed, I would prefer if we adopted similar conventions for both the `ToConcretize` global variable and the LLVM memory global variable.

> It might be best as an implicit parameter.

I would be fine with that too. I view passing explicit function arguments and passing implicit parameters as equally annoying, so I'll leave it to your taste to decide which option is more palatable :)

thinkmoore: FWIW, I don't think bundling things like this as part of the crucible personality is bad. It is state of the simulator that is not execution state specific and also should be abstracted away from the caller. I think it is an anti pattern to require the caller to keep track of and thread it through multiple invocations. Moreover I think this is exactly what Crucible personalities are for. If anything, I think more stuff should be pushed into personalities...

thinkmoore: That is, I take back saying it's a wart. I do think it is a common pattern that could be cleaned up (the need to create a fresh global variable to stash some stuff in and remember the name of the variable).

RyanGlScott: OK. While I'm not quite as excited to do so, I would also be fine with stashing all of the `GlobalVar`s in the personality type, provided that we advertise that these parts of the personality are read-only.

thinkmoore: For the latter, we could just not export the relevant accessors?

RyanGlScott: Yes, that would be nice as well. Or, to refine your suggestion a bit: we could export accessors, but as `Getter`s instead of `Lens`es so that they support getting but not setting.

langston-barrett: That seems like a good solution to me. In a sense, Ryan's suggestion was about constraining the programmer (or communicating intended constraints): The variable *shouldn't* get mutated, so let's make it immutable. Stashing it in the personality and exposing a `Getter` is actually the most restrictive solution of all (even more restrictive than passing it around as an explicit or implicit parameter). It makes it so that it can only be set by code that knows the full personality type and has access to its constructor, and makes it so that its value can only be used by code that has access to the personality. It also yields a nice phase distinction: we don't need the global variable to register the overrides, we just need to know that it will be in the personality by the time the overrides are invoked. (This phase issue has led to some complications for me in the Crucible-LLVM-CLI, for example.)

langston-barrett: @RyanGlScott Is there a reason to expose a `Getter`, rather than just a function `p -> GlobalVar ...`? My intuition would be that the latter is (very slightly, you can just use `to`) easier to compose with lenses, whereas the former offers a substantial advantage in terms of error messages, type inference, and familiarity.

RyanGlScott: No particular reason, other than that we are already retrieving the personality state via lenses anyway, and we already offer lenses for other personality fields. We could just as well make them ordinary functions.

langston-barrett: Okay, done. Thanks for the discussion, y'all, I found it quite enlightening and I'm happy with the conclusion we came to!

langston-barrett: One last point of abstraction to consider: we could wrap the `GlobalVar ToConcretizeType` in a `newtype` and not export the constructor, making the implementation even more opaque to consumers. Any preference here @RyanGlScott?

RyanGlScott: I'm not sure if that level of abstraction would pay its weight just yet, so I'm content with the way things look currently.

langston-barrett:
Fixed by muxing the old `SymSequence` with the `cons`'d one based on the current path condition. Still needs Oughta `CHECK-NOT` assertion.

RyanGlScott (approved):

grease-cli/tests/test.lua
```diff
@@ -14,3 +14,17 @@ function no_heuristic() check 'Unable to find a heuristic for any goal' end
 function req_failed() check 'One or more assertions failed' end
 
 function uninit_stack() check 'Likely bug: uninitialized stack read' end
+
+-- TODO: Upstream to Oughta
```
langston-barrett: See https://github.com/GaloisInc/oughta/pull/25

I would be happy to cut a release and bump our dependency if we'd like, or to create a numbered TODO to remove this when we do so.

RyanGlScott: I'm fine with either choice.

langston-barrett: https://github.com/GaloisInc/oughta/pull/26

langston-barrett: https://hackage.haskell.org/package/oughta-0.3.0.0

Currently waiting for the package list to update.

PR #205, opened 2025-06-06
Remove timestamps from logs

### LINKED ISSUE #113: Remove timestamps from logs
I personally find them distracting. GREASE is not a long-running, networked application. It is unlikely that someone would cross-check GREASE's logs against those from another service.

2over12:
Closes #113

### DIFF
--- a/grease/grease.cabal
+++ b/grease/grease.cabal
@@ -146,7 +146,6 @@ library
     , prettyprinter ^>= 1.7.1
     , safe-exceptions ^>= 0.1.7.4
     , text ^>= { 2.0, 2.1 }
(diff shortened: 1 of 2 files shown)


langston-barrett (approved):
Looks great! Would you mind checking if we could also remove our dependency on `time` at this point?

https://github.com/GaloisInc/grease/blob/3aa9bc2822ace023f425d5e35348d91863e17be2/grease/grease.cabal#L149

langston-barrett:
LGTM, perhaps we could add a bit more content to the commit message when merging? In particular, mentioning the logs and saying `Fixes #113` would be nice.

PR #214, opened 2025-06-09
Disable `-Werror=prepositive-qualified-module` in Cabal files

### LINKED ISSUE #213: `-Werror=prepositive-qualified-module` interferes with `ghcid`
https://github.com/ndmitchell/ghcid
```
<interactive>:2:8-16: error: [GHC-07924] [-Wprepositive-qualified-module, Werror=prepositive-qualified-module]
    Found ‘qualified’ in prepositive position
    Suggested fixes:
      • Place ‘qualified’ after the module name.
      • Perhaps you intended to use ImportQualifiedPost
```
We should probably only `-Werror` this via `cabal configure` in CI, rather than in the Cabal files.

langston-barrett:
Otherwise it messes with `ghcid`. Fixes #213.

### DIFF
--- a/grease-aarch32/grease-aarch32.cabal
+++ b/grease-aarch32/grease-aarch32.cabal
@@ -27,7 +27,7 @@ common shared
   -- See doc/style.md
   ghc-options:
     -Werror=missing-export-lists
-    -Werror=prepositive-qualified-module
+    -Wprepositive-qualified-module
 
   -- Specifying -Wall and -Werror can cause the project to fail to build on
   -- newer versions of GHC simply due to new warnings being added to -Wall. To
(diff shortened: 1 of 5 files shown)


RyanGlScott (approved):
Looks reasonable, although it might be worth citing https://github.com/ndmitchell/ghcid/issues/358 as a specific reason why we downgrade this to a warning instead of a full-blown error in the `.cabal` file.

langston-barrett:
Ah shoot, this won't work because the `-Werror` passed to `cabal configure` will apply to *all* local packages, including submodules like Macaw. Perhaps we can just leave it as a warning locally *and* in CI for now.

PR #215, opened 2025-06-09
Also register `fresh-bytes` when executing binaries 

langston-barrett:
...by sharing a bit more code between the override-registration code for S-expression programs (which is eager) and that for binaries (which is lazy and recursive).

### DIFF
--- a/grease/src/Grease/Macaw/FunctionOverride.hs
+++ b/grease/src/Grease/Macaw/FunctionOverride.hs
@@ -4,6 +4,7 @@ Maintainer       : GREASE Maintainers <grease@galois.com>
 -}
 
 {-# LANGUAGE ExplicitNamespaces #-}
+{-# LANGUAGE GADTs #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ImplicitParams #-}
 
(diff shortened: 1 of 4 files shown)


langston-barrett:
The right thing to do would be to add a test-case with a binary that uses an override that calls `fresh-bytes`... But I'm always a bit reticent to check in another opaque binary blob. I will do this if requested!

grease/src/Grease/Macaw/FunctionOverride.hs
```diff
@@ -298,21 +299,72 @@ registerMacawForwardDeclarations ::
...
+    {- ^ The map of public function names to their overrides. -} ->
+  (forall args ret.
+    W4.FunctionName ->
+    C.FnHandle args ret ->
+    C.OverrideSim p sym (Symbolic.MacawExt arch) rtp a r ())
+    {- ^ What to do when a forward declaration cannot be resolved. -} ->
+  W4.FunctionName
+    {-^ Name of the forward declaration -} ->
+  C.FnHandle args' ret'
+    {-^ Handle to bind -} ->
```
RyanGlScott: Style nit pick: use consistent spacing:

```suggestion
    {- ^ What to do when a forward declaration cannot be resolved. -} ->
  W4.FunctionName
    {- ^ Name of the forward declaration -} ->
  C.FnHandle args' ret'
    {- ^ Handle to bind -} ->
```

RyanGlScott (approved):

PR #220, opened 2025-06-10
Move Macaw "setup hooks" from `grease-cli` into `grease`

langston-barrett:
So that downstream applications can use them.


grease-cli/src/Grease/Main.hs
```diff
@@ -940,36 +939,8 @@ simulateMacawSyntax la halloc archCtx simOpts parserHooks = do
...
-            Macaw.registerMacawOvForwardDeclarations bak funOvs startupOvFwdDecs
-
-        -- Register defined functions.
-        forM_ (CSyn.parsedProgCFGs prog) $ \(C.Reg.AnyCFG defCfg) -> do
-          C.SomeCFG defSsa <- pure $ C.toSSA defCfg
-          -- This could probably be a helper defined in Crucible...
-          let bindCfg c = C.bindFnHandle (C.cfgHandle c) (C.UseCFG c (C.postdomInfo c))
-          bindCfg defSsa
+  let setupHook :: forall sym. Macaw.SetupHook sym arch
+      setupHook = Macaw.syntaxSetupHook la dl cfgs prog
```
RyanGlScott, line 943: This definition is now short enough that I wonder if we shouldn't just inline it.

grease-cli/src/Grease/Main.hs
```diff
@@ -1067,25 +1038,8 @@ simulateMacaw la halloc elf loadedProg mbPltStubInfo archCtx txtBounds simOpts p
...
-  -- `lookupFunctionHandle` and register them incrementally, but that is more
-  -- work. Given that startup overrides can't invoke anything defined in the
-  -- main program itself, it's much less work to register them ahead of time
-  -- here.)
   let setupHook :: forall sym. SetupHook sym arch
-      setupHook = Macaw.SetupHook $ \bak _mvar funOvs ->
-        forM_ (Map.elems cfgs) $ \(MacawEntrypointCfgs entrypointCfgs _) ->
-          forM_ (startupOvForwardDecs <$> entrypointStartupOv entrypointCfgs) $ \startupOvFwdDecs ->
-            Macaw.registerMacawOvForwardDeclarations bak funOvs startupOvFwdDecs
+      setupHook = Macaw.binSetupHook cfgs
```
RyanGlScott, line 1042: Similar comments here.

grease/src/Grease/Macaw/SetupHook.hs
```diff
@@ -0,0 +1,119 @@
...
+import Lang.Crucible.Syntax.Concrete qualified as CSyn
+import Stubs.FunctionOverride qualified as Stubs
+import What4.Expr.Builder qualified as WEB
+import What4.FunctionName qualified as WF
+import What4.Protocol.Online qualified as WPO
+
+-- | Hook to run before executing a CFG.
+--
+-- Note that @sym@ is a type parameter so that users can define 'SetupHook's
+-- that reference a fixed @sym@ type.
```
RyanGlScott: This isn't your fault (the comments were like this before), but this should mention `arch` as a fixed type parameter as well.

grease/src/Grease/Macaw/SetupHook.hs
RyanGlScott, line 1: Worth splitting out `Grease.LLVM.SetupHook` as well?

RyanGlScott (approved):

grease/src/Grease/Macaw/SetupHook.hs
langston-barrett: Yep, just less urgent!

grease-cli/src/Grease/Main.hs
langston-barrett: I see what you're saying, but the use-site is already a pretty long line so I think I'll keep it. The type signature also highlights that this needs to be polymorphic over `sym`, which is nice.

grease-cli/src/Grease/Main.hs
langston-barrett: Same response: https://github.com/GaloisInc/grease/pull/220#discussion_r2138571808

langston-barrett:
I've broken out a bunch of additional functions to make this more useful to downstream apps that don't use `Grease.Entrypoint`.

RyanGlScott (approved):

PR #226, opened 2025-06-11
doc: Describe some "fancy Haskell" idioms

### LINKED ISSUE #219: doc: Add a few dependently-typed Haskell tricks to the onboarding docs
- Pattern-matching on constraint-providing constructors (e.g., `Refl`) in a `do` block
- Pattern-matching on existentials in a `do` block
- Creating existentially-quantified `Some*` types
- "`let`-binding" variables in constraints with `~`

langston-barrett:
Fixes #219

### DIFF
--- a/doc/dev/haskell-idioms.md
+++ b/doc/dev/haskell-idioms.md
@@ -109,18 +109,18 @@ However, this *is* possible using `do`-notation. Since our `Some` is just a
 pure value (not a monadic action), we have to "inject" it into the monad using
 `pure`:
 ```haskell
-printWasItJust :: Some Maybe -> IO ()
-printWasItJust someMaybe = do
-  Some x <- pure someMaybe
+printWasItJust :: IO ()
+printWasItJust = do
+  Some x <- pure (makeSome 5)
   -- Again, can use x here, but only polymorphically
   when (isJust x) $
     putStrLn "it was Just"
 ```
 To understand why this works, desugar the `do` notation:
 ```haskell
-printWasItJust :: Some Maybe -> IO ()
-printWasItJust someMaybe =
-  pure someMaybe >>= \(Some x) -> when (isJust x) (putStrLn "it was Just")
+printWasItJust :: IO ()
+printWasItJust =
+  pure (makeSome 5) >>= \(Some x) -> when (isJust x) (putStrLn "it was Just")
 ```
 That's why you might see `do pat <- pure val`.
 
--- a/doc/dev/haskell-idioms.md
+++ b/doc/dev/haskell-idioms.md
@@ -12,7 +12,7 @@ emulated with [GADT]s like so:
 {-# LANGUAGE GADTs #-}
 
 data Some f where
-  Some :: f x -> Some f  -- the type does not mention `x`
+  Some :: f x -> Some f  -- the return type does not mention `x`
 ```
 (`Some` is canonically available as [`Data.Parameterized.Some.Some`].) As
 defined above, `Some` has kind `forall k. (k -> Type) -> Type`. What if you want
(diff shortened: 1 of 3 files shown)


doc/dev/haskell-idioms.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Types with names that start with `Some`
+
+Haskell currently lacks support for [first-class existential types]. They can be
+emulated with [GADT]s like so:
+```haskell
+{-# LANGUAGE GADTs #-}
+
+data Some f where
+  Some :: f x -> Some f  -- the type does not mention `x`
```
RyanGlScott: A bit pedantic, but I think this should say that the _return_ type does not mention `x`.

doc/dev/haskell-idioms.md
```diff
@@ -0,0 +1,182 @@
...
+```
+
+## Pattern matching on `pure` values in `do` blocks
+
+[GADT] constructors can existentially quantify types.
+```haskell
+{-# LANGUAGE GADTs #-}
+
+data Some f where
+  Some :: f x -> Some f  -- the type does not mention `x`
```
RyanGlScott: Similarly, say the _return_ type.

doc/dev/haskell-idioms.md
```diff
@@ -0,0 +1,182 @@
...
+```
+(`Some` is canonically available as [`Data.Parameterized.Some.Some`].) To use
+a `Some`, you generally pattern match on it. This brings the captured variable
```
RyanGlScott: Perhaps:

```suggestion
a `Some` value, you generally pattern match on it. This brings the captured variable
```

doc/dev/haskell-idioms.md
```diff
@@ -0,0 +1,182 @@
...
+into scope (though just as an unconstrained variable, so any code you write
+after the pattern match has to be polymorphic over it).
+```haskell
+wasItJust :: Some Maybe -> String
+wasItJust someMaybe =
+  case someMaybe of
+    -- Can use _x here, but have to be completely polymorphic over its type. Not
+    -- really much you *can* do with a value of an unknown type.
+    Some (Just _x) -> "it was Just"
+    Some Nothing -> "it was Nothing"
```
RyanGlScott: We might consider using an example where we _do_ use a value with an unknown type, e.g.,

```hs
someLength :: Some [] -> Int
someLength someList =
  case someList of
    -- Can use _x here, but you have to be completely polymorphic over its type.
    -- We use the `length` function, which works for list values of type `[x]` for all
    -- types `x`.
    Some l -> length l
```

doc/dev/haskell-idioms.md
```diff
@@ -0,0 +1,182 @@
...
+
+main :: IO ()
+main = putStrLn (wasItJust (Some (Just "ok")))
+```
+Sometimes, we want to bring the variable into scope by pattern-matching on
+the constructor, but we'd rather not use a `case` statement, as it increases
+indentation (e.g., we might be pattern-matching on several `Some`s in a row). We
+can't do this with `let`, because the type variable is only in scope for the RHS
```
RyanGlScott: "RHS" is jargon, and it might be worth spelling it out as "right-hand side".

doc/dev/haskell-idioms.md
```diff
@@ -0,0 +1,182 @@
...
+of the `=`.
+```haskell
+wontWork :: Some Maybe -> IO ()
+wontWork someMaybe = do
+  let Some x = someMaybe
+  -- try something with x...
+  pure ()
```
RyanGlScott: This example is perhaps not the most compelling, as you could simply match on `Some` in the function clause (i.e., `wontWork (Some x) = do ...`). To make it a bit more compelling, it would be worth introducing another function that returns a `Some Maybe` value, then call that function from `wontWork`.

doc/dev/haskell-idioms.md
```diff
@@ -0,0 +1,182 @@
...
+data Showable a where
+  Showable :: Show a => a -> Showable a
+
+showIt :: Showable a -> String
+showIt (Showable x) = show x  -- pattern matching brings Show a into scope
+
+main :: IO ()
+main = putStrLn (showIt (Showable "hello"))
+```
+(Operationally, the GADT constructor holds the vtable for the class.) GHC's
```
RyanGlScott: "vtable" is jargon, and one that I'm not sure if most Haskell programmers will know. We could consider "dictionary" instead, although that is also jargon. Perhaps it would be worth leaving a link that defines one of these two terms?

doc/dev/haskell-idioms.md
```diff
@@ -0,0 +1,182 @@
...
+built-in type equality constraint `~` is often "captured" this way, e.g.,
+```haskell
+{-# LANGUAGE GADTs #-}
+
+data IsString a where
+  IsString :: a ~ String => IsString a
```
RyanGlScott, line 155: We might also say that you can write this in an alternative way:

```hs
  IsString :: IsString String
```

This is relevant if you look up the definition of `Refl`, which uses an "implicit" equality constraint in this way.

doc/dev/haskell-idioms.md
langston-barrett: Good idea, thanks!

doc/dev/haskell-idioms.md
langston-barrett: Done, thanks!

doc/dev/haskell-idioms.md
```diff
@@ -0,0 +1,196 @@
...
+        a pattern with constructor:
+        ...
+```
+However, this *is* possible using `do`-notation. Since our `Some` is just a
+pure value (not a monadic action), we have to "inject" it into the monad using
+`pure`:
+```haskell
+printWasItJust :: Some Maybe -> IO ()
+printWasItJust someMaybe = do
+  Some x <- pure someMaybe
```
RyanGlScott: Should this use `makeSome`, like in the previous example?

doc/dev/haskell-idioms.md
```diff
@@ -0,0 +1,196 @@
...
+  -- Again, can use x here, but only polymorphically
+  when (isJust x) $
+    putStrLn "it was Just"
+```
+To understand why this works, desugar the `do` notation:
+```haskell
+printWasItJust :: Some Maybe -> IO ()
+printWasItJust someMaybe =
+  pure someMaybe >>= \(Some x) -> when (isJust x) (putStrLn "it was Just")
```
RyanGlScott: Similar question here

RyanGlScott (approved):

PR #230, opened 2025-06-15
Error on call to invalid address by default

### LINKED ISSUE #137: Make it an error to call an invalid address
We currently skip calls to invalid addresses. I think it would be less surpising to throw an error (i.e., add a failing assertion) as we do for symbolic function calls when `--error-symbolic-fun-calls` is set. We could make this configurable, but I think the use-case is less clear than it is in the case of symbolic calls.

https://github.com/GaloisInc/grease/blob/3e85804b12a144aaddadee0981b592f9a272be78/grease/src/Grease/Macaw/ResolveCall.hs#L257

2over12:
Fixes #137


grease/src/Grease/Options.hs
```diff
@@ -92,6 +99,7 @@ data SimOpts
     , simErrorSymbolicFunCalls :: ErrorSymbolicFunCalls
       -- | Default: 'False'.
     , simErrorSymbolicSyscalls :: ErrorSymbolicSyscalls
+    , simSkipInvalidCallAddrs :: SkipInvalidCallAddrs
```
RyanGlScott, line 103: For consistency with similar options above it:

```suggestion
      -- | Default: 'False'.
    , simSkipInvalidCallAddrs :: SkipInvalidCallAddrs
```

grease-cli/tests/prop/in-text/pos/invalid_call/test.c
```diff
@@ -0,0 +1,13 @@
+/* Copyright (c) Galois, Inc. 2024 */
+
+// all: flags {"--symbol", "test"}
+// all: flags {"--req", "in-text"}
```
RyanGlScott: Is this check relevant to the test case? If not, I'd propose to remove it to make the test case more minimal.

grease-cli/tests/prop/in-text/neg/invalid_call_skip/test.c
```diff
@@ -0,0 +1,13 @@
+/* Copyright (c) Galois, Inc. 2025 */
+
+// all: flags {"--symbol", "test", "--skip-invalid-call-addrs"}
+// all: flags {"--req", "in-text"}
```
RyanGlScott: Similarly, I wonder if this check can be removed.

grease-cli/tests/prop/in-text/pos/invalid_call/test.c
```diff
@@ -0,0 +1,13 @@
+/* Copyright (c) Galois, Inc. 2024 */
```
RyanGlScott:
```suggestion
/* Copyright (c) Galois, Inc. 2025 */
```

grease-cli/tests/prop/in-text/neg/simple/test.c
```diff
@@ -1,4 +1,4 @@
-/* Copyright (c) Galois, Inc. 2024 */
+/* Copyright (c) Galois, Inc. 2025 */
```
RyanGlScott: Any particular reason to update this test case's copyright year in particular? (There are no other changes to this file, so I'm not quite sure what the motivation is here.)

2over12: Should be fixed probably changed the wrong file

grease-cli/tests/sanity/pass/invalid_call/test.c
```diff
@@ -0,0 +1,16 @@
...
+int test() {
+    long x = 0xDEADBEEF;
+    int(*fptr)()=(int(*)()) x;
+    return fptr() + 1;
+}
+
+
+// arm: check "Invalid address: 0xdeadbeef"
+//TODO: This looks like a ppc32 bug?
+// ppc32: check "Invalid address: 0xdeadbeec"
```
RyanGlScott: I don't think so. Based on some searching, PowerPC is just a bit odd when it comes to setting the value of the link register before performing a function call, as it is known to add offsets to the current address. Perhaps worth a comment saying as such here, but I don't think this is a bug. (It would be surprising if anything worked at all in `macaw-ppc` otherwise!)

grease-cli/tests/sanity/pass/startup-override/test.c
```diff
@@ -14,6 +14,5 @@ void test(uint8_t *buf, size_t sz) {
   memset(buf, 'a', sz);
 }
 // arm: ok()
-// TODO: Why does PPC fail here?
-// ppc32: check "PPCInvalidInstruction"
+// ppc32: check "Could not identify block at"
```
RyanGlScott: This is happening because of [limitations in `macaw`'s heuristics for finding PLT stub addresses](https://github.com/GaloisInc/macaw/issues/386). To work around these limitations, we need to manually tell `grease` where the `memset` PLT stub is on PowerPC. To do so: add the following to `// ppc32: flags ...` above:

```c
// ppc32: flags {..., "--plt-stub", "0x10000220:memset"}
```

2over12: got it, thanks for the help

grease-cli/tests/sanity/pass/invalid_call_skip/test.c
```diff
@@ -0,0 +1,14 @@
...
+
+int test() {
+    long x = 0xDEADBEEF;
+    int(*fptr)()=(int(*)()) x;
+    return fptr() + 1;
+}
+
+// arm: check "TopV where PSTATE_T expected"
+// x64: ok()
+// ppc32: ok()
```
RyanGlScott: As far as I can tell, this is just a copy of `invalid_call` with different `grease` flags. Rather than checking in separate binaries for this, I propose we reuse the existing binaries for `invalid_call`. The `grease` test suite supports passing different sets of flags in multiple `grease` invocations. Concretely, I propose that `invalid_call/test.c` be changed to look like this:

```c
int test() {
    long x = 0xDEADBEEF;
    int(*fptr)()=(int(*)()) x;
    return fptr() + 1;
}


// all: flags {"--symbol", "test"}
// all: go(prog)
// arm: check "Invalid address: 0xdeadbeef"
// on ppc the last two bits are cleared
// ppc32: check "Invalid address: 0xdeadbeec"
// x64: check "Invalid address: 0xdeadbeef"

// all: flags {"--symbol", "test", "--skip-invalid-call-addrs"}
// all: go(prog)
// arm: check "TopV where PSTATE_T expected"
// x64: ok()
// ppc32: ok()
```

RyanGlScott (approved):
Looks great! One minor suggestion, but otherwise LGTM.

PR #233, opened 2025-06-17
`Grease.Shape.Parse`: Allow parsing exact bytes with RLE

### LINKED ISSUE #231: shapes dsl: Allow run-length encoding for individual bytes
The [shapes DSL](https://github.com/GaloisInc/grease/blob/455c740bd1a9ff0859352790e9d6e6e97edc00fd/doc/shape-dsl.md#concrete-syntax) allows writing long sequences of (un)initialized bytes using *run-length encoding*. For example, `##` is the syntax for an uninitialized byte, and `##*00000040` denotes 64 uninitialized bytes in a row. We should allow using this syntax for literal bytes, so that `80*00000040` would be 64 `80` bytes in a row.

RyanGlScott:
Fixes #231.

### DIFF
--- a/doc/shape-dsl.md
+++ b/doc/shape-dsl.md
@@ -78,10 +78,10 @@ numbers of that bit-width, namely 8 for 32-bit and 16 for 64-bit. This scheme
 ensures that pointers have the same textual width as a pointer-width sequence
 of bytes.
 
-Long, repeated sequences of symbolic or uninitialized bytes can be written with
-*run-length encoding*. For example, a sequence of 64 uninitialized bytes would
-be `##*00000040`. Just as for pointer offsets, the run length is represented as
-a hexadecimal number of the size of the machine bit-width.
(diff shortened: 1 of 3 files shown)


langston-barrett (approved):
LGTM! I expect the Hedgehog generator will only rarely exercise this code path, but that seems OK. Could perhaps consider adding a branch to the `MemShape` generator that generates a long sequence of identical bytes, but I'm fine if we don't do this.

RyanGlScott:
> Could perhaps consider adding a branch to the `MemShape` generator that generates a long sequence of identical bytes

Correct me if I'm wrong, but run-length encodings don't actually get represented directly in the `MemShape` AST, right? As such, I'm not sure if we'd gain any additional test coverage by doing this.

RyanGlScott:
In that case, I'm not sure I understand what exact the point of doing this would be. I think I'm missing something.

PR #240, opened 2025-06-18
Fix parsing of exact-byte RLE

### LINKED ISSUE #238: `--initial-precondition` parse error involving RLE for bytes
`grease` successfully parses this initial precondition file, which specifies that the first argument should point to a 32-bit `0` value:

```
rdi: 00+0000000000000000

00: 00*4
```

Now let us suppose we want the first argument to point to the value `1` instead. I tried this:

```
rdi: 00+0000000000000000

00: 01 00*3
```

But surprisingly, `grease` fails to parse this:

```
$ cabal -v0 run exe:grease -- grease-cli/tests/x86/id.x64.cbl --symbol test --initial-precondition initial-precondition2.txt
grease: initial-precondition2.txt:3:10:
  |
3 | 00: 01 00*3
  |          ^
unexpected '*'
expecting end of input, newline, or space
```

langston-barrett:
Fixes #238.

### DIFF
--- a/grease/src/Grease/Shape/Parse.hs
+++ b/grease/src/Grease/Shape/Parse.hs
@@ -283,20 +283,19 @@ parseUninitExplicit :: Parser Bytes.Bytes
 parseUninitExplicit =
   Bytes.toBytes . List.length Functor.<$> trySepBy1 (MP.chunk "##") memShapeSep
 
-parseRle :: Parser Char -> Parser (Char, Char, Bytes)
+parseRle :: Parser Char -> Parser Bytes
 parseRle parseChar = do
   -- The `try` is necessary to disambiguate from explicit bytes
-  (c1, c2) <-
-    MP.try $ do
-      c1 <- parseChar
-      c2 <- parseChar
-      _ <- MP.single '*'
-      pure (c1, c2)
+  MP.try $ do
+    _ <- parseChar
+    _ <- parseChar
+    _ <- MP.single '*'
+    pure ()
   rl <- MPCL.hexadecimal
-  pure (c1, c2, Bytes.toBytes @Int rl)
+  pure (Bytes.toBytes @Int rl)
 
 parseUninitRle :: Parser Bytes.Bytes
-parseUninitRle = thd Functor.<$> parseRle (MP.single '#')
+parseUninitRle = parseRle (MP.single '#')
 
 parseInit :: Parser Bytes.Bytes
 parseInit = parseInitRle MP.<|> parseInitExplicit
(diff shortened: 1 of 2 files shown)


grease/src/Grease/Shape/Parse.hs
```diff
@@ -315,20 +316,17 @@ hexCharsToWord8 c1 c2 = Tuple.fst (List.head (Numeric.readHex (c1:c2:[])))
...
+parseExactly :: Parser (Seq Word8)
+parseExactly = mconcat <$> trySepBy1 parseExactlyByte memShapeSep
+
+-- | Parse either a byte (e.g., @e0@) or a RLE'd byte (e.g., @e0*0a@).
+parseExactlyByte :: Parser (Seq Word8)
+parseExactlyByte = do
+  (c0, c1) <- (,) Functor.<$> MPC.hexDigitChar Applicative.<*> MPC.hexDigitChar
+  num <-
+    (MPC.char '*' Applicative.*> MPCL.hexadecimal) MP.<|>
+      (MP.notFollowedBy MPC.hexDigitChar Applicative.*> Applicative.pure 1)
```
RyanGlScott: For readability's sake, could you put each of these parsers into their own definition? Something like:

```hs
  let rle = MPC.char '*' Applicative.*> MPCL.hexadecimal
  let nonRle = MP.notFollowedBy MPC.hexDigitChar Applicative.*> Applicative.pure 1
  num <- rle MP.<|> nonRle
```

grease/src/Grease/Shape/Parse.hs
```diff
@@ -315,20 +316,17 @@ hexCharsToWord8 c1 c2 = Tuple.fst (List.head (Numeric.readHex (c1:c2:[])))
...
-    trySepBy1 (MP.lookAhead justTwoHex Applicative.*> twoHex) memShapeSep
-  where twoHex = (,) Functor.<$> MPC.hexDigitChar Applicative.<*> MPC.hexDigitChar
-        justTwoHex = twoHex Applicative.<* MP.notFollowedBy MPC.hexDigitChar
+parseExactly :: Parser (Seq Word8)
+parseExactly = mconcat <$> trySepBy1 parseExactlyByte memShapeSep
+
+-- | Parse either a byte (e.g., @e0@) or a RLE'd byte (e.g., @e0*0a@).
+parseExactlyByte :: Parser (Seq Word8)
+parseExactlyByte = do
+  (c0, c1) <- (,) Functor.<$> MPC.hexDigitChar Applicative.<*> MPC.hexDigitChar
```
RyanGlScott: How about just:

```suggestion
  c0 <- MPC.hexDigitChar
  c1 <- MPC.hexDigitChar
```

This avoids an intermediate tuple.

grease/src/Grease/Shape/Parse.hs
```diff
@@ -315,20 +316,17 @@ hexCharsToWord8 c1 c2 = Tuple.fst (List.head (Numeric.readHex (c1:c2:[])))
...
-parseExactlyExplicit =
-  Functor.fmap (Tuple.uncurry hexCharsToWord8) Functor.<$>
-    trySepBy1 (MP.lookAhead justTwoHex Applicative.*> twoHex) memShapeSep
-  where twoHex = (,) Functor.<$> MPC.hexDigitChar Applicative.<*> MPC.hexDigitChar
-        justTwoHex = twoHex Applicative.<* MP.notFollowedBy MPC.hexDigitChar
+parseExactly :: Parser (Seq Word8)
+parseExactly = mconcat <$> trySepBy1 parseExactlyByte memShapeSep
+
+-- | Parse either a byte (e.g., @e0@) or a RLE'd byte (e.g., @e0*0a@).
+parseExactlyByte :: Parser (Seq Word8)
```
RyanGlScott, line 318: It might be worth a comment here explaining why we don't use `parseRle` in the implementation of this function.

RyanGlScott, line 318: Relatedly, after this PR, the only two call sites to `parseRle` are in `parseUninitRle` and `parseInitRle`, both of which simply take the third element of the tuple and discard the rest. I think we could simplify the type of `parseRle` to just:

```hs
parseRle :: Parser Char -> Parser Bytes
```

grease/src/Grease/Shape/Parse.hs
langston-barrett: Good call

PR #242, opened 2025-06-20
Factor LLVM setup hook into its own module

### LINKED ISSUE #221: Factor out `Grease.LLVM.SetupHook`
As is done in #220 for Macaw

_Originally posted by @RyanGlScott in https://github.com/GaloisInc/grease/pull/220#discussion_r2138560545_

langston-barrett:
Fixes #221.

### DIFF
--- a/grease/src/Grease/LLVM/SetupHook.hs
+++ b/grease/src/Grease/LLVM/SetupHook.hs
@@ -1,0 +1,142 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE ImplicitParams #-}
+
+-- | c.f. "Grease.Macaw.SetupHook"
+module Grease.LLVM.SetupHook
+  ( SetupHook(..)
+  , syntaxSetupHook
+  , moduleSetupHook
+  ) where
+
+import Control.Lens ((^.))
+import Control.Monad qualified as Monad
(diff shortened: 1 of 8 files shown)


grease/src/Grease/LLVM/SetupHook.hs
```diff
@@ -0,0 +1,142 @@
...
+doLog :: MonadIO m => GreaseLogAction -> Diag.Diagnostic -> m ()
+doLog la diag = LJ.writeLog la (LLVMSetupHookDiagnostic diag)
+
+-- | Hook to run before executing a CFG.
+--
+-- Note that @sym@ is a type parameter so that users can define 'SetupHook's
+-- that reference a fixed @sym@ type.
+--
+-- c.f. 'Grease.Macaw.SetupHook.SetupHook'
+newtype SetupHook sym arch
```
RyanGlScott, line 50: This move changed the definition of `SetupHook` slightly: while it was previously only quantified over `sym`, it is now quantified over `sym` and `arch`. Any particular reason for doing this? (I don't feel strongly either way, but if we do this, we should update the Haddocks accordingly to mention both `sym` and `arch`.)

langston-barrett: It's necessary because `ModuleSetupHook` takes a `ModuleTranslation arch`. Not really sure how / why this worked previously. Will document.

PR #243, opened 2025-06-20
Add `--overrides-yaml`

### LINKED ISSUE #20: Provide way for users to map function names/addresses to override files
Currently, `grease` assumes that if you supply `--overrides <fun>.cbl`, then `<fun>` will correspond exactly to the name of the function symbol defined in the program being simulated. This works for a good number of use cases, but leaves some things to be desired. In particular:

1. Some binaries have functions at addresses which lack function symbols (e.g., stripped binaries), which means that there is effectively no way to override these functions.
2. Some function names contain periods (e.g., `llvm.assume`) or characters that are not legal in file names on certain operating systems, which leaves users no way to name the `.cbl` file after them.
3. In the context of Rust binaries, a single override might take on multiple function symbols depending on the name mangling scheme in use. It is annoying to have to copy override files for the sake of appeasing `grease`, so it would be nice if there were a more lightweight way to map between these names.

My proposed solution is to take inspiration from `ambient-verifier` and offer an [`overrides.yaml` file](https://github.com/GaloisInc/ambient-verifier/blob/eab04abb9750825a25ec0cbe0379add63f05f6c6/README.rst?plain=1#L164-L210), where a user can specify how function names and addresses map to specific override files. We might not need _everything_ that `ambient-verifier`'s `overrides.yaml` offers, however (e.g., startup overrides).

RyanGlScott:
This adds a way to specify function overrides for binaries by address (instead of by name, as is the case with `--overrides`). The new `--overrides-yaml` flag takes a YAML file that looks like this:

```yaml
function address overrides:
  ADDR_1: "NAME_1"
  ...
  ADDR_n: "NAME_n"
```

For more details, see the new documentation in `doc/overrides.md`.

Fixes #20.

### DIFF
--- a/grease/src/Grease/Syntax.hs
+++ b/grease/src/Grease/Syntax.hs
@@ -150,28 +150,43 @@ resolveOverridesYaml loadOpts mem fnOvNames (ParsedOverridesYaml parsedFunAddrs)
 -- | Helper, not exported.
 --
 -- Parse the contents of an overrides-related @.yaml@ file from a 'Aeson.Value'.
+-- This also ensures that the top-level YAML object does not have any unexpected
+-- keys.
 parseFunctionAddressOverrides ::
   Aeson.Value ->
   IO ParsedOverridesYaml
 parseFunctionAddressOverrides val = do
-  obj <- asObject val
-  case KeyMap.lookup "function address overrides" obj of
+  mbObj <- asNullableObject val
+  case mbObj of
     Nothing -> pure mempty
-    Just funAddrOvs -> do
-      funAddrOvsObj <- asObject funAddrOvs
-      funAddrOvPairs <-
-        traverse
-          (\(addrKey, funName) -> do
-            addr <-
-              case Read.readMaybe (Key.toString addrKey) of
-                Just addr -> pure addr
-                Nothing -> throw $ GreaseException $
-                  "Expected address in overrides YAML file, but encountered " <>
-                  Key.toText addrKey
-            funNameText <- asString funName
-            pure (addr, W4.functionNameFromText funNameText))
-          (KeyMap.toList funAddrOvsObj)
-      pure $ ParsedOverridesYaml $ Map.fromList funAddrOvPairs
+    Just obj -> do
+      let objSansFunAddrOvs = KeyMap.delete funAddrOvsKey obj
+      unless (KeyMap.null objSansFunAddrOvs) $
+        throw $ GreaseException $ Text.unlines $
+          "Unexpected keys in overrides.yaml file:" :
+          map
+            (\key -> "- " <> Key.toText key)
+            (KeyMap.keys objSansFunAddrOvs)
+      case KeyMap.lookup funAddrOvsKey obj of
+        Nothing -> pure mempty
+        Just funAddrOvs -> do
+          funAddrOvsObj <- asObject funAddrOvs
+          funAddrOvPairs <-
+            traverse
+              (\(addrKey, funName) -> do
+                addr <-
+                  case Read.readMaybe (Key.toString addrKey) of
+                    Just addr -> pure addr
+                    Nothing -> throw $ GreaseException $
+                      "Expected address in overrides YAML file, but encountered " <>
+                      Key.toText addrKey
+                funNameText <- asString funName
+                pure (addr, W4.functionNameFromText funNameText))
+              (KeyMap.toList funAddrOvsObj)
+          pure $ ParsedOverridesYaml $ Map.fromList funAddrOvPairs
+  where
+    funAddrOvsKey :: Key.Key
+    funAddrOvsKey = "function address overrides"
 
 -- | Helper, not exported.
 --
@@ -191,4 +206,17 @@ asObject :: Aeson.Value -> IO Aeson.Object
 asObject (Aeson.Object o) = pure o
 asObject v = throw $ GreaseException $
   "Expected object in overrides YAML file, but encountered " <>
+  Text.decodeUtf8 (BS.toStrict (Aeson.encode v))
+
+-- | Helper, not exported.
+--
+-- Assert that a JSON 'Aeson.Value' is an 'Aeson.Object' or 'Aeson.Null'. If it
+-- is an 'Aeson.Object', return 'Just' the underlying object. If it is
+-- 'Aeson.Null', return 'Nothing'. If neither of these are the case, throw an
+-- exception.
+asNullableObject :: Aeson.Value -> IO (Maybe Aeson.Object)
+asNullableObject (Aeson.Object o) = pure $ Just o
+asNullableObject Aeson.Null = pure Nothing
+asNullableObject v = throw $ GreaseException $
+  "Expected null or object in overrides YAML file, but encountered " <>
   Text.decodeUtf8 (BS.toStrict (Aeson.encode v))
--- a/grease-cli/tests/ux/overrides-yaml-bad-key.aux.yaml
+++ b/grease-cli/tests/ux/overrides-yaml-bad-key.aux.yaml
@@ -1,0 +1,2 @@
+funktion addres overides:
+  0x1234: "wat"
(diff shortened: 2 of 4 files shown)


doc/overrides.md
```diff
@@ -151,6 +151,60 @@ grease prog.bc --overrides id_ptr.llvm.cbl --overrides id_ptr2.llvm.cbl
...
+
+  Passing `--overrides chatty_malloc.x86.cbl` won't work, however, since the
+  name of the function in the binary is `malloc`, not `chatty_malloc`. Nor
+  would it work to rename `chatty_malloc` to `malloc`, as that name would clash
+  with the built-in override, which we need to use in the implementation of the
+  custom override.
+
+  Overrides YAML files offer a solution to this problem. We can pass
+  `--overrides-yaml overrides.yaml` to `grease`, where the contents of
+  `overrids.yaml` look something like this:
```
langston-barrett: While I agree that this schema is necessary and appropriate for stripped binaries, this part seems like a suboptimal workaround. The user knows the name of the function they want to override and the name of the S-expression CFG they want to override it. Why should they need to know the address of the function they want to override? Perhaps we should consider having a separate table `function name overrides` that would allow a key like `malloc: chatty_malloc`?

Happy to defer this to an issue.

grease-cli/src/Grease/Cli.hs
```diff
@@ -136,6 +136,11 @@ simOpts = do
                               <> Opt.metavar "FILE"
                               <> Opt.help "function overrides, in Crucible S-expression syntax"
                               ))
+  simOverridesYaml <-
+    Opt.many (Opt.strOption ( Opt.long "overrides-yaml"
+                              <> Opt.metavar "FILE"
+                              <> Opt.help "function overrides, in YAML format"
```
langston-barrett: Overrides only apply to functions, so "function overrides" strikes me as a bit redundant. nbd, though.

grease-cli/tests/refine/pos/stripped_with_overrides/test.c
```diff
@@ -0,0 +1,29 @@
...
+}
+
+void my_free(void* ptr) {}
+
+void test(void) {
+  int* x = my_malloc(sizeof(int));
+  *x = 42;
+  my_free(x);
+}
+// all: ok()
```
langston-barrett, line 29: I'm concerned that this assertion could pass even if we never applied any overrides - both functions would be skipped, and the store might succeed if `rax` already had a pointer. Perhaps we should add some assertions saying that the overrides were called?

doc/overrides.md
```diff
@@ -151,6 +151,60 @@ grease prog.bc --overrides id_ptr.llvm.cbl --overrides id_ptr2.llvm.cbl
 the public function of the same name. If `grease` cannot find a public function
 for an override named `id_ptr`, then `grease` will raise an error.
 
+## Overrides YAML files
```
langston-barrett, line 154: What happens (and what should happen) if I specify `--overrides` for a function `f` but also `--overrides-yaml` and an override for `f`'s address?

langston-barrett: Answering my own question: the address override takes priority. Perhaps let's just mention this in the docs!

grease/src/Grease/Macaw/ResolveCall.hs
```diff
@@ -247,14 +248,15 @@ lookupFunctionHandleResult ::
   Map.Map W4.FunctionName (MC.ArchSegmentOff arch) ->
   -- | Map of names of overridden functions to their implementations
   Map.Map W4.FunctionName (MacawSExpOverride p sym arch) ->
+  ResolvedOverridesYaml (MC.ArchAddrWidth arch) ->
```
langston-barrett, line 251: We now take two maps from addresses to function names, one for PLT stubs and one for these resolved overrides. Do you think it might be worth combining them ahead of time to reduce the number of parameters passed around and optimize lookups? (This would be for a follow-up isssue)

grease/src/Grease/Syntax.hs
```diff
@@ -56,3 +83,112 @@ parsedProgramCfgMap prog =
...
+
+-- | Helper, not exported.
+--
+-- Parse the contents of an overrides-related @.yaml@ file from a 'Aeson.Value'.
+parseFunctionAddressOverrides ::
+  Aeson.Value ->
+  IO ParsedOverridesYaml
+parseFunctionAddressOverrides val = do
+  obj <- asObject val
+  case KeyMap.lookup "function address overrides" obj of
```
langston-barrett: Looks like this will silently do nothing if the user misspells the key - perhaps we should check for unexpected keys?

langston-barrett (approved):
Very nice, this is a key improvement!

doc/overrides.md
RyanGlScott: I agree that this example is less compelling, and you're right that we could imagine having another mechanism in the `overrides.yaml` file that associates override names to custom overrides. No such mechanism exists yet, so I thought it would be worth at least explaining how to use `function address overrides` to support this use case for now.

I'll open a separate issue about implementing a more direct means of accomplishing this.

langston-barrett: For what it's worth, I appreciate the explanation of how to use this feature to address this use-case! Just also made me think we could support it even better with a dedicated feature.

grease-cli/src/Grease/Cli.hs
RyanGlScott: I just copy-pasted this help text from `--overrides`, which also describes them as "function overrides". I'm happy to drop the "function" part in a separate commit.

grease-cli/tests/refine/pos/stripped_with_overrides/test.c
RyanGlScott: I don't agree with this assessment. If we somehow failed to apply the overrides, then `my_malloc` would return `NULL`, which would cause the `*x = 42;` line to fail.

grease/src/Grease/Macaw/ResolveCall.hs
RyanGlScott: I'm not sure we would want to completely combine the maps, since we do want to treat the two cases a little differently (e.g., for logging purposes).

grease/src/Grease/Syntax.hs
RyanGlScott: I'm of two minds on this. On the one hand, I like the idea of having the ability to pass an empty `.yaml` file and have it produce an empty `Map`. (This would be especially nice if we add more features later to `overrides.yaml` and want the ability to opt in to certain features.) On the other hand, you are right that if the user misspells `function address overrides`, then this will silently drop the overrides.

As an alternative proposal, perhaps we should have another check that sees if there are any non-`function address overrides` keys in the top-level YAML object and throw an error if so. This would still allow the user to omit `function address overrides` if they wish while disallowing obvious typos.

grease-cli/tests/refine/pos/stripped_with_overrides/test.c
langston-barrett: D'oh! Of course.

grease/src/Grease/Macaw/ResolveCall.hs
langston-barrett: The logging angle could potentially be handled by adding an enumeration (to distinguish PLT stubs vs YAML overrides) to the values, but I trust your judgment.

grease/src/Grease/Syntax.hs
langston-barrett: Sounds good to me!

langston-barrett: Potentially worth UX tests for the misspelled key and empty YAML cases.

grease-cli/src/Grease/Cli.hs
langston-barrett: No biggie, up to you!

doc/overrides.md
RyanGlScott: I've opened https://github.com/GaloisInc/grease/issues/244 for this.

PR #245, opened 2025-06-20
`Grease.Shape.Parse`: Export `ParseError`

RyanGlScott:
This is used in the return type of `parseShapes` (part of the public API), and therefore `ParseError` ought to be part of the public API too. I'd like to use `ParseError` elsewhere in a downstream application.


langston-barrett (approved):

langston-barrett:
If only there were a GHC warning for this...

PR #248, opened 2025-06-24
Fourmolu format

2over12:
Also enforces checks in CI


langston-barrett:
We should be sure to document how to install and use fourmolu at the CLI in `doc/dev/dev.md`. Might also be worth linking to and/or describing how to set it up as a git pre-commit hook and/or editor integrations.

doc/dev/dev.md
```diff
@@ -73,6 +73,20 @@ We have a few Python scripts in `scripts/lint/` that perform one-off
 checks. They generally take some number of paths as arguments, check
 `.github/workflows/lint.yml` to see how they are invoked in CI.
 
+### Fourmolu
+
+This repo enforces Fourmolu formatting based on the committed configuration in `fourmolu.yaml`. Installation and usage directions are available
+[here](https://github.com/fourmolu/fourmolu). In short:
+
+```
```
langston-barrett:
```suggestion
```sh
```

doc/dev/dev.md
```diff
@@ -73,6 +73,20 @@ We have a few Python scripts in `scripts/lint/` that perform one-off
...
+cabal install fourmolu
+fourmolu --mode inplace $(git ls-files '*.hs')
+```
+
+One can either configure formatting commands as a [pre-commit hook](https://git-scm.com/docs/githooks#_pre_commit) or [auto-format on save](https://code.visualstudio.com/docs/editing/codebasics#_formatting) to avoid formatting issues.
+The repo is already formatted with fourmolu formatted changes should be localized to behavioral changes. Further discussion of rational
+and commit hygiene are available in this [formatting discussion](./formatting.md).
```
langston-barrett: Some typos / missing periods here I think? I'm finding these sentences hard to parse.

doc/dev/formatting.md
```diff
@@ -0,0 +1,42 @@
+# Discussion of Grease Formatting Choices
```
langston-barrett: The rest of the docs just use first-caps not title-case. Also, GREASE is all caps.
```suggestion
# Discussion of GREASE formatting choices
```

doc/dev/formatting.md
```diff
@@ -0,0 +1,42 @@
...
+
+Grease enforces an opinionated formatter in continuous integration (CI) via a [fourmolu](https://github.com/fourmolu/fourmolu) formatting lint.
+Enforcing a formatting choice in CI enables Grease developers to use automated formatters in either pre-commit or format-on-save without formatting unrelated
+text by maintaining a consistent format across the repository. This consistency limits time spent in review discussing formatting choices (e.g. case branch indentation or import grouping)
+and sets clear expectations with respect to formatting requirements for accepting a pull request. These clear expectations are particularly useful for new developers and open source contributors
+who may not be familiar with the style preferences of Grease developers. Below we highlight some pain points introduced by this approach and suggested mitigations.
```
langston-barrett: nbd but what's up with these line lengths? I would suggest either breaking at 80 characters or breaking after each sentence.

doc/dev/formatting.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Additional Setup Overhead Introduced by a Formatting Tool
```
langston-barrett: Same comment here and elsewhere re: title-case.

doc/dev/formatting.md
```diff
@@ -0,0 +1,42 @@
...
+
+In the [dev guide](./dev.md) we provide installation and setup instructions. Additionally, in VSCode, selecting fourmolu as the autoformatter for a workspace asks to install
+the formatting tool automatically.
+
+## Code Readability
+
+While, in general, autoformatters lead to consistent expectations with respect to style, they can alter developer intended whitespace.
+Sometimes these formatting choices may negatively impact readability.
+
+In general the consistency provided by fourmolu is preferred over altering formatting decisions, however, if the formatting is substantially negatively
```
langston-barrett: I might weaken this recommendation a bit. I would say something like: "We trust contributors to make conscientious choices about readability, including disabling the formatter in cases where it is appropriate. Developers should balance the benefits of local readability improvements against the corresponding reduction in global consistency. These trade-offs may be discussed during code review."

doc/dev/formatting.md
```diff
@@ -0,0 +1,42 @@
...
+impacting readability the following magic comments can be used to locally disable formatting:
+```
```
langston-barrett:
```suggestion
```haskell
```

doc/dev/formatting.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Developer Confusion Encountering CI Formatting Failures
+
+Developers may encounter unexpected formatting failures if they do not use fourmolu locally. The [dev guide](./dev.md) offers advice on how to
+setup fourmolu locally for autoformatting. Additionally, the CI job is named `Check formatting with fourmolu` to provide hints as to where a
+developer should look if encountering this failure.
+
+## Commit Hygiene
+
+In general, enforcing a consistent style helps maintain commit hygiene by avoiding commits with inconsistent formatting being merged into the main branch.
```
langston-barrett: Perhaps we can spell out how it helps maintain commit hygene in  a bit more detail? Something like "avoids commits with inconsistent formatting being created, which might later require formatting-only commits to fix" or something along those lines?

doc/dev/formatting.md
```diff
@@ -0,0 +1,42 @@
...
+Enforcing a formatting style, however, does mean developers cannot delay formatting choices to a later date. Ideally, developers use a local formatting setup described in the [dev guide](./dev.md) to prevent
+the need for post-hoc formatting commits, but these types of formatting fix commits are somewhat inevitable. There are at least two ways to handle this situation:
+1. If the PR is amenable, squashing the commits can be a good option. Then the feature commits are merged with the formatting fix, resulting in a single commit that implements the intended feature with the correct formatting, resulting in a minimal,
+local change.
+2. If squashing is not possible, then sequencing the commit fixups into a single formatting commit that applies to the entire feature branch. This strategy makes the formatting commit explicit.
```
langston-barrett: nit: Squashing is always *possible*, it is just sometimes *undesirable*.
```suggestion
2. If squashing would result in too many unrelated changes being combined into a single commit, then sequencing the formatting fix-up commit into a single commit that applies to the entire feature branch. This strategy makes the formatting commit explicit, and separates it from functional changes.
```

doc/diagrams/app/Main.hs
```diff
@@ -1,12 +1,12 @@
-{-|
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE NoMonomorphismRestriction #-}
```
langston-barrett: Huh, I'm surprised that Formolu doesn't alphabetize these...?

elf-edit-core-dump/extract-core-dump/Main.hs
```diff
@@ -21,63 +22,68 @@ import Data.ElfEdit.CoreDump qualified as CoreDump
...
+                -- the core was dumped. Otherwise, print nothing.
+                case CoreDump.coreDumpHeaderProgramCounter ehi notes of
+                    Left _err ->
+                        pure ()
+                    Right prStatusPc ->
+                        putStrLn $
+                            unlines
+                                [ ""
+                                , "Address where core was dumped: 0x" ++ showHex prStatusPc ""
+                                ]
```
langston-barrett: If we uniformly switch to four-space indentation, we'll likely get a lot of poorly-formatted code like this because it wasn't written with four-space indentation in mind (and so now will take up excessive vertical space to meet line-length requirements). I would probably prefer four-space indentation on a fresh project, but we'd better stick to two.

elf-edit-core-dump/src/Data/ElfEdit/Prim/Nhdr.hs
```diff
@@ -3,51 +3,58 @@
...
-    -- interpretations of a single 'NhdrType' value may exist. Thus, a program
-    -- must recognize both the 'nhdrNameSize' and the 'nhdrType' to recognize a
-    -- descriptor.
-  } deriving Show
+    { nhdrNameSize :: !Word32
+    {- ^ The number of bytes used to store a note's name (including a byte for a
+    null terminator). This does /not/ include the size of padding bytes
+    needed ensure that the name uses 4-byte alignment (if using
+    'Elf.ELFCLASS32') or 8-byte alignment (if using 'Elf.ELFCLASS64').
+    -}
```
langston-barrett: Potential configuration note: I don't love the insistence on block comments, but it's not a big deal to me.

fourmolu.yaml
```diff
@@ -0,0 +1,72 @@
...
+function-arrows: trailing
+
+# How to place commas in multi-line lists, records, etc. (choices: leading or trailing)
+comma-style: leading
+
+# Styling of import/export lists (choices: leading, trailing, or diff-friendly)
+import-export-style: diff-friendly
+
+# Rules for grouping import declarations
+import-grouping: legacy
```
langston-barrett: I think this would be consistent with our style guide:
```suggestion
import-grouping: single
```

fourmolu.yaml
```diff
@@ -0,0 +1,72 @@
...
+indent-wheres: false
+
+# Whether to leave a space before an opening record brace
+record-brace-space: false
+
+# Number of spaces between top-level declarations
+newlines-between-decls: 1
+
+# How to print Haddock comments (choices: single-line, multi-line, or multi-line-compact)
+haddock-style: multi-line
```
langston-barrett: I slightly prefer single line here, but I'm fine with any.
```suggestion
haddock-style: single-line
```

fourmolu.yaml
```diff
@@ -0,0 +1,72 @@
...
+haddock-style-module: null
+
+# Styling of let blocks (choices: auto, inline, newline, or mixed)
+let-style: auto
+
+# How to align the 'in' keyword with respect to the 'let' keyword (choices: left-align, right-align, or no-space)
+in-style: right-align
+
+# Whether to put parentheses around a single constraint (choices: auto, always, or never)
+single-constraint-parens: always
```
langston-barrett: I would prefer `auto` or `never` here, in line with https://galoisinc.github.io/grease/dev/style.html#superfluous-syntax

fourmolu.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+# Whether to put parentheses around a single deriving class (choices: auto, always, or never)
+single-deriving-parens: always
```
langston-barrett: I would prefer `auto` or `never` here, in line with https://galoisinc.github.io/grease/dev/style.html#superfluous-syntax

doc/dev/formatting.md
langston-barrett: (We could also potentially opt for the simpler title "Formatting" here.)

langston-barrett:
Thank you for writing up the discussion of pain points and their mitigations! Yet again, I'd be especially interested to hear from @RyanGlScott if all our gripes about formatters are described, and/or if any of them have insufficient plans for remediation.

I would say at a high level it might be helpful to have more explicit and detailed instructions for setting up format-on-save for VSCode. I personally would also appreciate a copy-pasteable shell-script block that sets up a pre-commit hook. I think a bit of duplication with docs hosted elsewhere is worth it to make developers' lives easier.

RyanGlScott:
I think the documentation here is sufficient, so please don't feel blocked on me to proceed.

doc/dev/formatting.md
2over12: I honestly do think the formatter selected choice should be preferred in all circumstances except when it drastically compromises readability (the formatter should not be disabled for stylistic preferences etc etc). But ill weaken this since i recognize im the outlier here

langston-barrett:
One last thought: it would probably be worth adding a git blame "ignore" file like this one that Cargo uses, and adding a section to the dev docs encouraging new devs to run
```sh
git config blame.ignorerevsfile .git-blame-ignore-revs
```
https://github.com/rust-lang/cargo/blob/7388d86d28b8599968e6e6f58a6f4e5528b604a1/.git-blame-ignore-revs#L1

doc/dev/formatting.md
langston-barrett: Yeah we can talk more about it offline but generally I find readability very important and use formatters as a tool to *improve* readability, so when they *reduce* readability they really aren't adding anything IMO. As an example of when I might disable a formatter, I'm thinking of cases like:
```haskell
let x = coefficient1         * var1 + offset1
    y = otherCoefficientName * var1 + offset2
in x + y
```
In this case, it's nice to be able to see at a glance that `var1` is used in both equations, despite the difference in name-length between the coefficients.

> the formatter should not be disabled for stylistic preferences

I agree with this statement and also acknowledge that the barrier between "readability" and "stylistic preferences" is an ambiguous one.

.git-blame-ignore-revs
```diff
@@ -0,0 +1,8 @@
+# 102 limit reformat
+a4cdac4778d1c1800c18d3288a0bb3ea37959c7c
+
+# reformat 
+b0b614d847cab7d5a6992ba4fc9f2810ec587c0b
+
+# fourmolu format
+f52ea87bd3ba9740bdff344aa98396e378240fbd
```
langston-barrett: Should we maybe squash all these?

2over12: we cant without losing the blame on docs

2over12: we could squash everything but then we end up ignoring things we dont want to. I guess in theory we could reorder commits but that's a bit of a pain

langston-barrett: Okay, up to you!

.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# fourmolu format
```
langston-barrett, line 3: Maybe worth leaving a comment at the beginning of this file saying what it's for, perhaps pointing to the dev docs that cover it

doc/dev/dev.md
```diff
@@ -73,6 +73,28 @@ We have a few Python scripts in `scripts/lint/` that perform one-off
 checks. They generally take some number of paths as arguments, check
 `.github/workflows/lint.yml` to see how they are invoked in CI.
 
+### Fourmolu
+
+This repo enforces Fourmolu formatting based on the committed configuration in `fourmolu.yaml`. Installation and usage directions are available
+[here](https://github.com/fourmolu/fourmolu). In short:
+
+```sh
+cabal install fourmolu
```
langston-barrett: Is it important for the version installed to match the version we use in CI? If so, is there a way to specify a version to `cabal install`?

doc/diagrams/app/Main.hs
```diff
@@ -1,11 +1,9 @@
-{-|
-Copyright        : (c) Galois, Inc. 2024
-Maintainer       : GREASE Maintainers <grease@galois.com>
--}
-
-{-# LANGUAGE NoMonomorphismRestriction #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE TypeFamilies #-}
+-- \|
```
langston-barrett: Is this escaping right? I would expect to see a module Haddock

doc/diagrams/app/Main.hs
```diff
@@ -1,11 +1,9 @@
...
+-- Copyright        : (c) Galois, Inc. 2024
+-- Maintainer       : GREASE Maintainers <grease@galois.com>
+{-# LANGUAGE NoMonomorphismRestriction #-}
```
langston-barrett, line 3: Seems odd that this is not in a block with the other pragmas

grease/src/Grease/Concretize/ToConcretize.hs
```diff
@@ -133,8 +131,8 @@ stateToConcretize =
 -- 'C.SimState'.
 addToConcretize ::
   forall p sym ext rtp args ret ty.
-  LCB.IsSymInterface sym =>
-  HasToConcretize p =>
+  (LCB.IsSymInterface sym) =>
+  (HasToConcretize p) =>
```
langston-barrett: Were we going to disable the unnecessary parens for single constraints?

grease/src/Grease/Macaw/Load.hs
```diff
@@ -313,10 +325,10 @@ resolveCoreDumpEntrypointAddress ::
   -- | A 'Map.Map' of function entrypoint addresses to their names.
   Map.Map (MM.MemSegmentOff w) BS.ByteString ->
   -- | The path to the core dump file, which was obtained by running the binary
-  -- to simulate.
+  --     to simulate.
```
langston-barrett: Huh, is there an option to control this? Don't love the extra spaces here. nbd though.

grease/src/Grease/Macaw/Load/Relocation.hs
```diff
@@ -66,57 +64,62 @@ elfRelocationMap _ loadOpts ehi =
...
+  resolveRelocSym ::
+    (Show reloc) =>
+    -- \| The relocation type (only used for displaying information in an
+    -- exception).
+    reloc ->
+    -- \| The relocation's address.
+    MM.MemWord w ->
+    -- \| The relocation symbol's index in the dynamic symbol table.
+    Word32 ->
+    BS.ByteString
```
langston-barrett: Looks like these were unnecessarily escaped too. Maybe let's grep for these escapes to ensure they're all handled before merging?

langston-barrett (approved):
Thanks for putting this together! Obviously it's hard to completely review such a large diff but we can fix issues as we find them. The main things I would want to see resolved before merge are (1) unnecessary parens in constraints and (2) some Haddocks seem unnecessarily escaped. Otherwise, LGTM!

doc/dev/dev.md
RyanGlScott: Alternatively, it might be worth pointing readers to the [releases page for `fourmolu`](https://github.com/fourmolu/fourmolu/releases).

grease/src/Grease/Macaw/Load/Relocation.hs
2over12: These should be like this `-- ^ The relocation's address.` right?

langston-barrett: I thought that `-- |` was valid syntax, but I might just be wrong. It doesn't seem to be listed here: https://haskell-haddock.readthedocs.io/latest/markup.html#function-arguments

RyanGlScott: I think that documentation is just wrong, as you can absolutely use `-- |` for function arguments...

PR #254, opened 2025-06-25
llvm: Basic support for reading function signatures from LLVM debug info

### LINKED ISSUE #115: llvm: Use type information to bootstrap refinement loop
Consider the following C program:
```c
typedef struct list {
  int val;
  struct list *next;
} list;

int test(list *l) {
  int len = 0;
  while (l->next != 0) {
    len++;
    l = l->next;
  }
  return len;
}
```
The type `list *l` tells us a lot about what the function expects from its first argument. This information is not immediately apparent at the LLVM level, because the argument type gets replaced by the opaque `ptr` type:
```
clang -fno-discard-value-names -emit-llvm -grecord-gcc-switches -S -O1 -g struct-ptr.c

rg 'define' struct-ptr.ll

define i32 @test(ptr nocapture noundef readonly %l) local_unnamed_addr #0 !dbg !10 {
```
However, the debug info `!10` *does* contain the argument type:
```
!10 = distinct !DISubprogram(name: "test", scope: !8, file: !8, line: 6, type: !11, scopeLine: 6, flags: DIFlagPrototyped | DIFlagAllCallsDescribed, spFlags: DISPFlagDefinition | DISPFlagOptimized, unit: !7, retainedNodes: !21)
!11 = !DISubroutineType(types: !12)
!12 = !{!13, !14}
!13 = !DIBasicType(name: "int", size: 32, encoding: DW_ATE_signed)
!14 = !DIDerivedType(tag: DW_TAG_pointer_type, baseType: !15, size: 64)
!15 = !DIDerivedType(tag: DW_TAG_typedef, name: "list", file: !8, line: 4, baseType: !16)
!16 = distinct !DICompositeType(tag: DW_TAG_structure_type, name: "list", file: !8, line: 1, size: 128, elements: !17)
!17 = !{!18, !19}
!18 = !DIDerivedType(tag: DW_TAG_member, name: "val", scope: !16, file: !8, line: 2, baseType: !13, size: 32)
!19 = !DIDerivedType(tag: DW_TAG_member, name: "next", scope: !16, file: !8, line: 3, baseType: !20, size: 64, offset: 64)
!20 = !DIDerivedType(tag: DW_TAG_pointer_type, baseType: !16, size: 64)
```
We should use this information to help GREASE come up with a better initial precondition so that the refinement loop takes less time.

These `DI*` nodes are documented [in the LLVM Language Reference](https://llvm.org/docs/LangRef.html#specialized-metadata-nodes). We would likely access them through [the `llvm-pretty` API](https://hackage-content.haskell.org/package/llvm-pretty-0.13.0.0/docs/Text-LLVM-AST.html#t:DebugInfo-39-)

langston-barrett:
Fixes #115.

Watch in amazement as it treats a pointer-size integer as an integer, not a pointer!
```c
#include <stdint.h>
uintptr_t test(uintptr_t x) { return x; }
int main() { return test(0); }
```
```
cabal run exe:grease -- --symbol test -v test.bc

-- snip --

Using precondition:

%0: XX XX XX XX XX XX XX XX
```

Needs:

- [x] More haddocks
- [x] Tests for struct types / padding 
- [x] Tests for union types
- [x] Tests for float types
- [x] Reference to https://github.com/GaloisInc/llvm-pretty/issues/162
- [x] CLI-level toggle


grease/src/Grease/LLVM/DebugInfo.hs
```diff
@@ -0,0 +1,203 @@
...
+    _ -> Nothing
+
+-- It is acceptable for this function to under-estimate the size of a type:
+-- GREASE's heuristics will expand the allocation as necessary.
+minTypeSize :: LDU.Info -> Bytes
+minTypeSize =
+  \case
+    LDU.ArrInfo _elemTy -> 0 -- arrays can be empty
+    LDU.BaseType _nm dibt -> Bytes.bitsToBytes (L.dibtSize dibt)
+    LDU.Pointer _ty -> 8 -- assuming x86_64
```
RyanGlScott: Perhaps we should derive this value from a `HasPtrWidth` constraint just in case we ever support different pointer sizes? We are already passing `HasPtrWidth` through to `diArgShapes`, so it shouldn't take _too_ much additional plumbing to do this.

grease/src/Grease/LLVM/DebugInfo.hs
```diff
@@ -0,0 +1,203 @@
...
+    LDU.Structure _nm fields -> structPtrTarget fields
+    LDU.Typedef _nm ty -> asPtrTarget ty
+    t@(LDU.Union _nm _fields) -> uninit (minTypeSize t)
+    LDU.Unknown -> uninit 0
+ where
+  uninit sz = PtrShape.PtrTarget Nothing (Seq.singleton (PtrShape.Uninitialized sz))
+
+  -- For structs, we need to add uninitialized padding bytes between their
+  -- fields.
+  structPtrTarget fields =
```
RyanGlScott, line 111: This could benefit from a type signature, I think.

grease/src/Grease/LLVM/DebugInfo.hs
```diff
@@ -0,0 +1,203 @@
...
+         in if tgtSize >= minSize
+              then tgt
+              else
+                panic
+                  "Size mismatch in struct shape calculation"
+                  [ "Expected size at least " ++ show minSize
+                  , "Found size " ++ show tgtSize
+                  ]
+
+  fieldsPtrTarget =
```
RyanGlScott, line 126: This could benefit from a type signature, I think.

grease/src/Grease/LLVM/DebugInfo.hs
```diff
@@ -0,0 +1,203 @@
...
+     in snd . Foldable.foldl' go (0, Seq.empty)
+   where
+    go (bytesWritten, memShapes) sfi =
+      let
+        !bitsWritten = fromIntegral (Bytes.bytesToBits bytesWritten)
+        !missingBits = LDU.sfiOffset sfi - bitsWritten
+        !missingBytes = Bytes.bitsToBytes missingBits
+        !fieldPtrTarget = asPtrTarget (LDU.sfiInfo sfi)
+        !fieldSize = PtrShape.ptrTargetSize (knownNat @64) fieldPtrTarget
+        !bytesWritten' = bytesWritten + missingBytes + fieldSize
```
RyanGlScott, line 139: This code is pretty subtle. If I understand correctly, you're computing the size of each field plus its padding bytes and adding those to the target shape. Could you leave some comments to this effect here?

grease/src/Grease/LLVM/DebugInfo.hs
```diff
@@ -0,0 +1,203 @@
...
+            panic
+              "Size mismatch in struct shape calculation"
+              [ "Bytes written before this field: " ++ show bytesWritten
+              , "Padding bytes: " ++ show missingBytes
+              , "Field size: " ++ show fieldSize
+              , "Total bytes written including this field: " ++ show bytesWritten'
+              , "Size: " ++ show tgtSz
+              ]
+
+decodeType :: LDU.Info -> Maybe DITypeView
```
RyanGlScott: It might be worth a comment that this only supports simple pointer and integer-like types for now.

PR #257, opened 2025-06-26
tests: Embed `CFLAGS` in comments in C sources for LLVM bc tests

### LINKED ISSUE #256: tests: Per-test `CFLAGS`
I want to add an LLVM bitcode test that compiles with `-g` to test #115. This flag is not used by default in our Makefile, probably to save on repo size:

https://github.com/GaloisInc/grease/blob/c38fa7bd67bb82bd7d90656a9d0d7aac51d3020b/grease-cli/tests/Makefile#L203

I could add some more rules to the Makefile, but it's already pretty long and complicated. To keep the Makefile simple, and to make the tests more self-contained, I wonder if we could instead embed additional `CFLAGS` in the tests themselves. I'm thinking adding comments like `// CFLAGS: -g` to the C files, and then using `grep`/`sed`/whatever in the Makefile to extract them and add them to the command line. This could possibly be done inline, or in a separate script.

langston-barrett:
Fixes #256. Only does it for tests in `llvm-bc/` at the moment because that's my immediate need, but paves the way for future extensions.

### DIFF
--- a/grease-cli/tests/Makefile
+++ b/grease-cli/tests/Makefile
@@ -119,10 +119,11 @@ SOS = $(SHARED_LIBRARY_ARM_SOS) $(SHARED_LIBRARY_X64_SOS) $(SHARED_LIBRARY_PPC32
 
 BC_DIRS = $(shell find llvm-bc/ -mindepth 1 -maxdepth 1 -type d)
 BCS = $(addsuffix /test.bc,$(BC_DIRS))
+STANDALONE_BCS = $(patsubst %.c, %.bc, $(wildcard llvm-bc/*.c))
 
 .PHONY: all fmt clean
 
-all: $(EXES) $(SOS) $(BCS)
+all: $(EXES) $(SOS) $(BCS) $(STANDALONE_BCS)
 
 fmt:
 	clang-format -i $(DIRS:=/test.c)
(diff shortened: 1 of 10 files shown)


grease-cli/tests/Makefile
```diff
@@ -119,10 +119,11 @@ SOS = $(SHARED_LIBRARY_ARM_SOS) $(SHARED_LIBRARY_X64_SOS) $(SHARED_LIBRARY_PPC32
 
 BC_DIRS = $(shell find llvm-bc/ -mindepth 1 -maxdepth 1 -type d)
 BCS = $(addsuffix /test.bc,$(BC_DIRS))
+STANDALONE_BCS = $(patsubst %.c, %.bc, $(wildcard llvm-bc/*.c))
```
langston-barrett, line 122: AFAICT rules for these files were just missing.

grease-cli/tests/llvm-bc/declare-in-override.c
```diff
@@ -1,5 +1,7 @@
 /* Copyright (c) Galois, Inc. 2024 */
 
+// CFLAGS: -emit-llvm -frecord-command-line
```
RyanGlScott: To avoid duplicating these flags across all `llvm-bc` tests, I wonder if we should make some pre-defined flag groups available (e.g., something like `// CFLAGS: $LLVM_BC_FLAGS`).

grease-cli/tests/Makefile
```diff
@@ -200,7 +201,10 @@ $(STANDALONE_SHARED_LIBRARY_PPC32_SOS): %/test.ppc32.elf: %/test.c
 	$(PPC32CC) $(CFLAGS_COMMON) $(CFLAGS_NO_LIBS) -fpic -shared $< -o $@
 
 $(BCS): %/test.bc: %/test.c
-	$(X64CC) -emit-llvm -frecord-command-line -c -o $@ $<
+	$(X64CC) $(shell ./extract-cflags.sh $<) -c -o $@ $<
+
+llvm-bc/%.bc: llvm-bc/%.c
```
RyanGlScott, line 206: Should `STANDALONE_BCS` be used instead of `llvm-bc/%.bc` here?

langston-barrett: I don't think so
```
make: *** No rule to make target `llvm-bc/%.c', needed by `llvm-bc/declare-in-override.bc'.  Stop.
```

RyanGlScott: Oops, I overlooked the "standalone" part of that name :)

grease-cli/tests/llvm-bc/declare-in-override.c
langston-barrett: Good idea, ported the script to Python and added this feature.

PR #259, opened 2025-06-27
Add basic shape inference from DWARF

2over12:


grease-x86/src/Grease/Macaw/Arch/X86.hs
```diff
@@ -39,6 +41,10 @@ import What4.Interface qualified as W4
 
 type instance ArchReloc X86.X86_64 = EE.X86_64_RelocationType
 
+-- | x64 System-V assumed
+reglist :: [Some.Some X86.X86Reg]
+reglist = Some.Some <$> [X86.RDI, X86.RSI, X86.RDX, X86.RCX, X86.R8, X86.R9]
```
RyanGlScott: You should be able to simplify this definition by using [`x86ArgumentRegs`](https://github.com/GaloisInc/macaw/blob/c240264e41829894a231ae01cafc35607f641816/x86/src/Data/Macaw/X86/X86Reg.hs#L474-L475) from `macaw-x86`.

grease-cli/src/Grease/Main.hs
```diff
@@ -269,6 +273,33 @@ withMemOptions opts k =
           }
    in k
 
+loadDwarfpreconditions ::
```
RyanGlScott: Style nit pick: use lowerCamelCase:

```suggestion
loadDwarfPreconditions ::
```

grease-cli/src/Grease/Main.hs
```diff
@@ -269,6 +273,33 @@ withMemOptions opts k =
...
+  ExtShape ext ~ PtrShape ext w =>
+  Mem.HasPtrWidth w =>
+  Symbolic.SymArchConstraints arch =>
+  -- | Is DWARF enabled
+  Bool ->
```
RyanGlScott: If I understand correctly, then if this is `False`, then this will basically just become the identity function, which feels a bit strange design-wise. I would prefer to have this function unconditionally load DWARF preconditions, and then have callers decide whether to apply the function or not...

grease-cli/src/Grease/Main.hs
```diff
@@ -598,8 +629,19 @@ simulateMacawCfg la bak fm halloc macawCfgConfig archCtx simOpts setupHook mbCfg
...
+  -- so we find the segment that corresponds by checking for all segments if segindex->memsegment -> the segment of the target addr
+  let segIndexToSegment = MM.memSegmentIndexMap memory
+  let targetSegment = fst <$> ((\seg -> find (\(_, seg2) -> seg == seg2) (Map.toList segIndexToSegment)) =<< (MM.segoffSegment <$> mbCfgAddr))
+  let mbSegment = (\bin -> find (\phdr -> (Just $ Elf.phdrSegmentIndex phdr) == targetSegment) $ headerPhdrs bin) =<< mcElf macawCfgConfig
+  let mbNewAddr =
+        ( \(seghdr, entAddr) ->
+            (MM.memWordValue $ MM.segoffOffset entAddr) + fromIntegral (Elf.phdrSegmentVirtAddr seghdr)
+        )
+          <$> ((,) <$> mbSegment <*> mbCfgAddr)
+  dwarfedArgs <- loadDwarfpreconditions (simEnableDWARFPreconditions simOpts) argNames initArgs_ macawCfgConfig archCtx mbNewAddr
```
RyanGlScott: ...specifically, here would be a good place to control whether to call this function or not. Something like:

```hs
  dwarfedArgs <-
    if simEnableDWARFPreconditions simOpts
      then loadDwarfPreconditions argNames initArgs_ macawCfgConfig archCtx mbNewAddr
      else pure initArgs_
```

grease/src/Grease/GitRev.hs
```diff
@@ -1,6 +1,6 @@
 {-# LANGUAGE TemplateHaskell #-}
 
--- | These are placed in their own module to minimize the cost of recompilation
+-- \| These are placed in their own module to minimize the cost of recompilation
```
RyanGlScott: Why this change?

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,206 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+   in (C.Some <$> pshape)
+pointerShapeOfDwarf _ _ _ _ = Nothing
+
+-- Is there really nothing available that looks like this?
+takeJust :: (a -> Maybe b) -> [a] -> [b]
+takeJust _ [] = []
+takeJust f (h : tl) =
+  case f h of
+    Nothing -> []
+    Just e -> e : takeJust f tl
```
RyanGlScott: This is [`Data.Maybe.mapMaybe`](https://hackage.haskell.org/package/base-4.21.0.0/docs/Data-Maybe.html#v:mapMaybe).

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,206 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+              C.Some r <- pure reg
+              shp <- shapeFromVar aContext r sub var
+              pure (Text.pack $ coerce $ mkRegName r, shp)
+        )
+        args
+    named = Map.fromList $ ascParams
+   in
+    ParsedShapes{_getParsedShapes = named}
+
+fromDwarfInfo :: (Symbolic.SymArchConstraints arch, ExtShape ext ~ PtrShape ext wptr, Mem.HasPtrWidth wptr) => ArchContext arch -> Word64 -> [Data.Macaw.Dwarf.CompileUnit] -> Maybe (ParsedShapes ext)
```
RyanGlScott: Style suggestion: this line is quite long, and it would be helpful to split it up into multiple lines for the sake of readability. Something like this, perhaps:

```hs

fromDwarfInfo ::
  ( Symbolic.SymArchConstraints arch
  , ExtShape ext ~ PtrShape ext wptr
  , Mem.HasPtrWidth wptr
  ) =>
  ArchContext arch ->
  Word64 ->
  [Data.Macaw.Dwarf.CompileUnit] ->
  Maybe (ParsedShapes ext)
```

And similarly for the other newly introduced functions in this module.

grease-cli/src/Grease/Main.hs
```diff
@@ -598,8 +629,19 @@ simulateMacawCfg la bak fm halloc macawCfgConfig archCtx simOpts setupHook mbCfg
   let mdEntryAbsAddr = fmap (segoffToAbsoluteAddr memory) mbCfgAddr
   initArgs_ <- minimalArgShapes bak archCtx mdEntryAbsAddr
   let argNames = fmapFC (Const . getValueName) rNameAssign
+  -- Is there a better way to do this resolution? We need the pc absent any position independent loading
+  -- so we find the segment that corresponds by checking for all segments if segindex->memsegment -> the segment of the target addr
```
RyanGlScott: If I understand correctly, you are trying to take `mbCfgAddr :: Maybe (ArchSegmentOff arch)` and convert it to a `Word64` such that it does _not_ have `grease`'s constant load offset applied. (That is, you'd want something like `0x1234` rather than `0x10001234`.) Is that right? If so, I think the most straightforward way to accomplish this would be to do the following:

1. Use [`segoffToAbsoluteAddr`](https://github.com/GaloisInc/grease/blob/5802e9b03d848611eeba0ce1fa49093db7e32f8e/grease/src/Grease/Utility.hs#L97-L114) to convert this to a `MemWord`.
2. Use [`memWordValue`](https://github.com/GaloisInc/macaw/blob/c240264e41829894a231ae01cafc35607f641816/base/src/Data/Macaw/Memory.hs#L305) to get the underlying `Word64` from the `MemWord`.
3. Subtract [`loadOffset loadOpts`](https://github.com/GaloisInc/macaw/blob/c240264e41829894a231ae01cafc35607f641816/base/src/Data/Macaw/Memory/LoadCommon.hs#L31) from the resulting address to remove the constant load offset.

grease-x86/src/Grease/Macaw/Arch/X86.hs
2over12: I dont think so, we need them in ABI order, and only pointer/int regs right?

grease/src/Grease/Shape.hs
2over12: That isnt quite what we want right? because we want to stop after a Nothing in order to avoid bad types. Point being if we cant build a member for a dwarf item because of some mistmatch e.g. dwarf info that cause it to span two regs then we would be offset and the next shape would hit the wrong reg

grease-x86/src/Grease/Macaw/Arch/X86.hs
RyanGlScott: I'm not quite sure what you mean here. These are in fact in ABI order, unless I'm missing something.

2over12: Ah never mind i misread 👍

grease/src/Grease/Shape.hs
RyanGlScott: Ah, I missed the fact that this stops at the first `Nothing`. (That's what I get for only reading the type signature.)

In that case, I think including this definition is fine. You might want to leave a comment on what it's doing, however (if for no other reason than to avoid confusing this with `mapMaybe`, which is a pretty common function of that type).

grease-cli/src/Grease/Main.hs
2over12: must this always work? i could imagine PIE situations where a segment gets loaded somewhere else for whatever reason but yeah im looking to map this to the DWARF pc for the target CU which is going to be whatever relocatable object it's ins virtual address without any load base

grease-cli/src/Grease/Main.hs
2over12: imo it's doing the same thing as `loadInitialPreconditions` where each loader either maintains the args or updates them with new info, but i dont have strong feelings either way tbh

RyanGlScott: Fair enough, I didn't realize that there was prior art. I'm content with copying what `loadInitialPreconditions` does.

grease-cli/src/Grease/Main.hs
RyanGlScott: Admittedly, I'm not sure off the top of my head, but I _think_ this should work even in position-independent contexts. Come to think of it, it may the case that `segoffToAbsoluteAddr` already removes the constant load offset—I'm not sure. Worth double-checking me here.

2over12: got it, ill take a look, the situation I was thinking of is if you have a target entrypoint inside of a shared library dep, then if you would load both in macaw, one of them must not be directly at the base. But come to think of it, idk if we do that in grease currently

RyanGlScott: Indeed, `grease` doesn't support shared library entrypoints, so that part shouldn't matter here.

2over12:
Need to add handling of recursive types

grease-cli/src/Grease/Cli.hs
```diff
@@ -185,6 +187,8 @@ simOpts = do
             <> Opt.metavar "FILE"
             <> Opt.help "Initial precondition for use in refinement"
         )
+  simEnableDWARFPreconditions <-
+    Opt.switch (Opt.long "enable-dwarf-preconditions" <> Opt.help "enables dwarf parsing based initial shapes. manual shapes supersede these shapes.")
```
langston-barrett, line 200: What do you think about enabling this by default? Feels like it might be nice out of the box behavior to me.

grease-cli/src/Grease/Main.hs
```diff
@@ -269,6 +273,33 @@ withMemOptions opts k =
...
+  IO (Shape.ArgShapes ext NoTag tys)
+loadDwarfPreconditions dwarfPrecs argNames initShapes macawCfgConfig archContext targetAddr =
+  let dwarfPrs = do
+        addr <- targetAddr
+        elf <- snd . Elf.getElf <$> mcElf macawCfgConfig
+        let (_, cus) = dwarfInfoFromElf elf
+        shps <- Shape.fromDwarfInfo archContext addr cus
+        let repl = Shape.replaceShapes argNames initShapes shps
+        either (const Nothing) Just repl
+   in pure $ (if dwarfPrecs then fromMaybe initShapes dwarfPrs else initShapes)
```
langston-barrett: Could we simplify by inlining the `fromMaybe`?
```suggestion
        let repl = Shape.replaceShapes argNames initShapes shps
        either (const initShapes) id repl
   in pure $ (if dwarfPrecs then dwarfPrs else initShapes)
```

langston-barrett: Also, the parens are unnecessary around the `if`

grease-cli/src/Grease/Main.hs
langston-barrett: > Indeed, grease doesn't support shared library entrypoints, so that part shouldn't matter here.

Doesn't it?

```
echo 'int main() { return 0; }' > test.c
clang -shared test.c
cabal run exe:grease -- --symbol main a.out

Finished analyzing 'main'.
```

grease-cli/src/Grease/Main.hs
```diff
@@ -598,8 +629,19 @@ simulateMacawCfg la bak fm halloc macawCfgConfig archCtx simOpts setupHook mbCfg
   let mdEntryAbsAddr = fmap (segoffToAbsoluteAddr memory) mbCfgAddr
   initArgs_ <- minimalArgShapes bak archCtx mdEntryAbsAddr
   let argNames = fmapFC (Const . getValueName) rNameAssign
+  -- Is there a better way to do this resolution? We need the pc absent any position independent loading
+  -- so we find the segment that corresponds by checking for all segments if segindex->memsegment -> the segment of the target addr
+  let segIndexToSegment = MM.memSegmentIndexMap memory
+  let targetSegment = fst <$> ((\seg -> find (\(_, seg2) -> seg == seg2) (Map.toList segIndexToSegment)) =<< (MM.segoffSegment <$> mbCfgAddr))
+  let mbSegment = (\bin -> find (\phdr -> (Just $ Elf.phdrSegmentIndex phdr) == targetSegment) $ headerPhdrs bin) =<< mcElf macawCfgConfig
```
langston-barrett: Might be more readable using `do` notation instead of bind+lambda.

grease-cli/src/Grease/Main.hs
```diff
@@ -598,8 +629,19 @@ simulateMacawCfg la bak fm halloc macawCfgConfig archCtx simOpts setupHook mbCfg
...
+  let mbNewAddr =
+        ( \(seghdr, entAddr) ->
```
langston-barrett: nit: `lowerCamelCase`
```suggestion
        ( \(segHdr, entAddr) ->
```

grease-cli/tests/Makefile
```diff
@@ -95,12 +95,15 @@ SHARED_LIBRARY_PPC32_EXES = $(addsuffix /test.ppc32.elf,$(SHARED_LIBRARY_DIRS))
 # that links against the shared library. Note that we use the file extension
 # .elf rather than .so in order to make it easier for the grease test suite to
 # discover the shared libraries.
-STANDALONE_SHARED_LIBRARY_DIRS = sanity/pass/simulate_from_shared_library
+STANDALONE_SHARED_LIBRARY_DIRS = sanity/pass/simulate_from_shared_library
```
langston-barrett: Nit: Why the whitespace change?

grease-cli/tests/sanity/pass/dwarf-shape-recursive/test.c
```diff
@@ -0,0 +1,19 @@
...
+  int sum = 0;
+  while (curr) {
+    sum += curr->x;
+    curr = curr->next;
+  }
+  return sum;
+}
+
+// all: flags {"--symbol", "foo", "--enable-dwarf-preconditions",
+// "--no-heuristics"} x64: go(prog) x64: ok()
```
langston-barrett: This looks a bit odd, perhaps try:
```suggestion
// all: flags {"--symbol", "foo"}
// all: flags {"--enable-dwarf-preconditions", "--no-heuristics"}
// x64: go(prog)
// x64: ok()
```

grease-x86/src/Grease/Macaw/Arch/X86.hs
```diff
@@ -39,6 +41,10 @@ import What4.Interface qualified as W4
 
 type instance ArchReloc X86.X86_64 = EE.X86_64_RelocationType
 
+-- | x64 System-V assumed
+reglist :: [Some.Some X86.X86Reg]
+reglist = Some.Some <$> X86.x86ArgumentRegs
```
langston-barrett: Please use camelCase

grease/src/Grease/Macaw/Arch.hs
```diff
@@ -245,6 +246,9 @@ data ArchContext arch = ArchContext
   -- Currently, x86-64 is the only supported architecture that does something
   -- non-trivial here. All other architectures can simply leave the stack
   -- pointer by implementing @'_archOffsetStackPointerPostCall' = 'pure'@.
+  , _archABIParams :: [C.Some (MC.ArchReg arch)]
+  -- ^ The ordered integer and pointer ABI registers for the Architecture. In the future
```
langston-barrett: Why capitalize "architecture"? If it's a link, surround in single quotes

grease/src/Grease/Macaw/Arch.hs
```diff
@@ -245,6 +246,9 @@ data ArchContext arch = ArchContext
...
+  -- this should handle things like stack params, and multiple calling conventions.
```
langston-barrett: Let's make a numbered TODO that references a GitHub issue describing what remains to be done here.

grease/src/Grease/Macaw/PLT.hs
```diff
@@ -1,7 +1,7 @@
 {-# LANGUAGE ExplicitNamespaces #-}
 {-# LANGUAGE OverloadedStrings #-}
 
--- | See @doc/shared-libraries.md@ for user-facing documentation about PLT
+-- \| See @doc/shared-libraries.md@ for user-facing documentation about PLT
```
langston-barrett: Why did this get escaped?

grease/src/Grease/Macaw/Syscall.hs
```diff
@@ -1,6 +1,6 @@
 {-# LANGUAGE ExplicitNamespaces #-}
 
--- | Functionality for converting 'Stubs.Syscall's into functions
+-- \| Functionality for converting 'Stubs.Syscall's into functions
```
langston-barrett: Same q

grease/src/Grease/Shape.hs
```diff
@@ -28,29 +29,52 @@ module Grease.Shape (
   minimalShapeWithPtrs',
   traverseShapeWithType,
   tagWithType,
+  ParsedShapes (..),
```
langston-barrett, line 34: FYI you can split the Haddocks into sections the same way you've done in the module itself like so:
```suggestion
  -- * Replacing
  ParsedShapes (..),
```

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+      [ "Type mismatch for"
+      , PP.pretty (typeMismatchName tm) PP.<> ":"
+      , "expected:"
+      , PP.viaShow (expectedType tm)
+      , "but found:"
+      , PP.viaShow (foundType tm)
+      ]
+
+newtype ParsedShapes ext
+  = ParsedShapes {_getParsedShapes :: Map.Map Text (C.Some (Shape ext NoTag))}
```
langston-barrett, line 353: Could use a Haddock

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+                    Left $
+                      TypeMismatch
+                        { typeMismatchName = nm
+                        , expectedType = C.Some ty
+                        , foundType = C.Some ty'
+                        }
+      Nothing -> Right s
+
+isInSubProg :: Word64 -> Subprogram -> Bool
+isInSubProg w sub =
```
langston-barrett: The `Word64` could use a comment or a newtype to make it clear what it represents

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+  let entryMatch = (==) w <$> MDwarf.subEntryPC sub
+      def = fromMaybe False
+   in def
+        ( do
+            sdef <- MDwarf.subDef sub
+            lpc <- MDwarf.subLowPC sdef
+            hpc <- MDwarf.subHighPC sdef
+            let res = w >= lpc && w < (lpc + hpc)
+            pure $ res
```
langston-barrett: tiny tiny style nit: This could probably just be inlined
```suggestion
            pure $ w >= lpc && w < (lpc + hpc)
```

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+  do
+    let mp = MDwarf.subTypeMap sprog
+    (mbtypeApp, _) <- Map.lookup vrTy mp
+    rightToMaybe mbtypeApp
+
+constructPtrTarget :: HasPtrWidth w => Subprogram -> MDwarf.TypeApp -> Maybe (PtrTarget w NoTag)
+constructPtrTarget sprog tyApp =
+  PtrTarget Nothing <$> shapeSeq tyApp
+ where
+  -- TODO: are these always byte fields?
```
langston-barrett: Not sure I understand the question but let me know if there's anything I can clarify or help with here!

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+  ishape w = Just $ Seq.singleton $ Initialized NoTag (toBytes w)
+  padding :: (Integral a) => a -> MemShape w NoTag
+  padding w = Uninitialized (toBytes w)
+  -- if we dont know where to place members we have to fail/just place some bytes
+  buildMember :: Word64 -> MDwarf.Member -> Maybe (Word64, Seq.Seq (MemShape w NoTag))
+  buildMember loc mem =
+    ( \memLoc ->
+        do
+          padd <- Just (if memLoc == loc then Seq.empty else Seq.singleton (padding $ memLoc - loc))
+          membershape <- shapeOfTyApp $ MDwarf.memberType mem
```
langston-barrett: camelCase please
```suggestion
          memberShape <- shapeOfTyApp $ MDwarf.memberType mem
```

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+constructPtrTarget sprog tyApp =
+  PtrTarget Nothing <$> shapeSeq tyApp
+ where
+  -- TODO: are these always byte fields?
+  ishape w = Just $ Seq.singleton $ Initialized NoTag (toBytes w)
+  padding :: (Integral a) => a -> MemShape w NoTag
+  padding w = Uninitialized (toBytes w)
+  -- if we dont know where to place members we have to fail/just place some bytes
+  buildMember :: Word64 -> MDwarf.Member -> Maybe (Word64, Seq.Seq (MemShape w NoTag))
+  buildMember loc mem =
```
langston-barrett: "mem" seems a little ambiguous ("memory"), perhaps "member"?
```suggestion
  buildMember loc member =
```

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+    ( \memLoc ->
+        do
+          padd <- Just (if memLoc == loc then Seq.empty else Seq.singleton (padding $ memLoc - loc))
+          membershape <- shapeOfTyApp $ MDwarf.memberType mem
+          let memByteSize = sum $ memShapeSize ?ptrWidth <$> membershape
+          let nextLoc = memLoc + fromIntegral memByteSize
+          Just (nextLoc, padd Seq.>< membershape)
+    )
+      =<< MDwarf.memberLoc mem
```
langston-barrett: Why not:
```suggestion
  memLoc <- MDwarf.memberLoc mem
  padd <- Just (if memLoc == loc then Seq.empty else Seq.singleton (padding $ memLoc - loc))
  membershape <- shapeOfTyApp $ MDwarf.memberType mem
  let memByteSize = sum $ memShapeSize ?ptrWidth <$> membershape
  let nextLoc = memLoc + fromIntegral memByteSize
  Just (nextLoc, padd Seq.>< membershape)
```

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+  shapeOfTyApp :: MDwarf.TypeRef -> Maybe (Seq.Seq (MemShape w NoTag))
+  shapeOfTyApp x = shapeSeq =<< extractType sprog x
+
+  shapeSeq :: MDwarf.TypeApp -> Maybe (Seq.Seq (MemShape w NoTag))
+  shapeSeq (MDwarf.UnsignedIntType w) = ishape w
+  shapeSeq (MDwarf.SignedIntType w) = ishape w
+  shapeSeq MDwarf.SignedCharType = ishape (1 :: Int)
+  shapeSeq MDwarf.UnsignedCharType = ishape (1 :: Int)
+  -- Need modification to DWARF to collect counts properly + evaluate ops for ub
```
langston-barrett: What does "evaluate ops for ub" mean?

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+  shapeSeq (MDwarf.ArrayType _elTy _ub) = Nothing
+  -- need to compute padding between els, infront of els and at end
```
langston-barrett: What is the distinction between padding "between" elements and padding "in front" of each element?
```suggestion
  -- need to compute padding between elements, in front of each element and at the end
```

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+  shapeSeq (MDwarf.StructType sdecl) =
+    let structSize = MDwarf.structByteSize sdecl
+     in do
+          (currLoc, seqs) <-
+            foldM
+              ( \(currloc, seqs) mem ->
```
langston-barrett:
```suggestion
              ( \(currLoc, seqs) mem ->
```

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+                  do
+                    (nextLoc, additionalShapes) <- buildMember currloc mem
+                    Just (nextLoc, seqs Seq.>< additionalShapes)
+              )
+              (0, Seq.empty)
+              (MDwarf.structMembers sdecl)
+          endPad <- Just (if currLoc >= structSize then Seq.empty else Seq.singleton (padding $ structSize - currLoc))
```
langston-barrett: Why not
```suggestion
          let endPad = if currLoc >= structSize then Seq.empty else Seq.singleton (padding $ structSize - currLoc)
```

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+  ArchContext arch ->
+  MC.ArchReg arch tp ->
+  Subprogram ->
+  MDwarf.TypeApp ->
+  Maybe (C.Some (PtrShape ext w NoTag))
+pointerShapeOfDwarf _ r _ (MDwarf.SignedIntType _) = intPtrShape r
+pointerShapeOfDwarf _ r _ (MDwarf.UnsignedIntType _) = intPtrShape r
+pointerShapeOfDwarf _ _ sprog (MDwarf.PointerType _ tyRef) =
+  let mshape = constructPtrTarget sprog =<< extractType sprog =<< tyRef
+      pshape = ShapePtr NoTag (Offset (toBytes (0 :: Int))) <$> mshape
```
langston-barrett: camelCase please

langston-barrett: I think `Bytes` might be `Num`, in which case you can do this:
```suggestion
  let mshape = constructPtrTarget sprog =<< extractType sprog =<< tyRef
      pshape = ShapePtr NoTag (Offset 0) <$> mshape
```

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+extractType :: Subprogram -> MDwarf.TypeRef -> Maybe MDwarf.TypeApp
+extractType sprog vrTy =
+  do
+    let mp = MDwarf.subTypeMap sprog
+    (mbtypeApp, _) <- Map.lookup vrTy mp
+    rightToMaybe mbtypeApp
+
+constructPtrTarget :: HasPtrWidth w => Subprogram -> MDwarf.TypeApp -> Maybe (PtrTarget w NoTag)
+constructPtrTarget sprog tyApp =
+  PtrTarget Nothing <$> shapeSeq tyApp
```
langston-barrett: Should we consider using `ptrTarget`, or will `shapeSeq` be guaranteed to create "normalized" `PtrTarget`s?

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+  , Mem.HasPtrWidth wptr
+  , Symbolic.SymArchConstraints arch
+  ) =>
+  ArchContext arch ->
+  MC.ArchReg arch tp ->
+  Subprogram ->
+  MDwarf.Variable ->
+  Maybe (C.Some (Shape ext NoTag))
+shapeFromVar arch buildingForReg sprog vr =
+  C.mapSome ShapeExt <$> (pointerShapeOfDwarf arch buildingForReg sprog =<< extractType sprog =<< MDwarf.varType vr)
```
langston-barrett: Maybe use `do` notation to avoid the long line?

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+
+shapeFromDwarf :: (Symbolic.SymArchConstraints arch, ExtShape ext ~ PtrShape ext wptr, Mem.HasPtrWidth wptr) => ArchContext arch -> Subprogram -> ParsedShapes ext
+shapeFromDwarf aContext sub =
+  let
+    abiregs = aContext Lens.^. archABIParams
```
langston-barrett: camelCase please

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
 instance (MC.PrettyF tag, PrettyExt ext tag) => PP.Pretty (ArgShapes ext tag tys) where
   pretty (ArgShapes regs) =
     MC.foldlFC (\doc rShape -> PP.vcat [doc, PP.pretty rShape]) "" regs
+
+-- * Replacing
```
langston-barrett: I wonder if we shouldn't add this content to a module under `Grease.Macaw`, considering that it is binary-specific.

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+    args = (zip abiregs $ snd <$> (toAscList $ subParamMap sub))
+    ascParams =
+      takeJust
+        ( \(reg, var) ->
+            do
+              C.Some r <- pure reg
+              shp <- shapeFromVar aContext r sub var
+              pure (Text.pack $ coerce $ mkRegName r, shp)
+        )
+        args
```
langston-barrett: In situations like this, I like to name the function parameter so that the overall computation is clear, e.g.
```
  let f reg var = ...
  in takeJust (uncurry f) args
```

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+    named = Map.fromList $ ascParams
```
langston-barrett: Unnecessary `$`

langston-barrett: Also, maybe just inline

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+   in
+    ParsedShapes{_getParsedShapes = named}
+
+fromDwarfInfo ::
+  (Symbolic.SymArchConstraints arch, ExtShape ext ~ PtrShape ext wptr, Mem.HasPtrWidth wptr) =>
+  ArchContext arch ->
+  Word64 ->
+  [Data.Macaw.Dwarf.CompileUnit] ->
+  Maybe (ParsedShapes ext)
+fromDwarfInfo aContext addr cus =
```
langston-barrett: Needs Haddock

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+                    List.find
+                      ( \x ->
+                          let rs = cuRanges x
+                           in let isInCU =
+                                    any
+                                      ( \range ->
+                                          let begin = rangeBegin range
+                                              end = rangeEnd range
+                                           in begin <= mval && mval < end
+                                      )
```
langston-barrett: Perhaps name this something like `inRange`

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+                                      rs
+                               in isInCU
+                      )
```
langston-barrett: Perhaps name this something like `isInCUs`

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+                      cus
+               in let targetSubProg = (\x -> List.find (isInSubProg mval) (cuSubprograms x)) =<< targetCu
```
langston-barrett: Again, `do` notation could help here

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +338,230 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+                   in ( shapeFromDwarf
+                          aContext
+                          <$> targetSubProg
+                      )
+        )
+   in res
```
langston-barrett: Inline?

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +340,260 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+                  (MDwarf.structMembers sdecl)
+              endPad <- lift $ Just (if currLoc >= structSize then Seq.empty else Seq.singleton (padding $ structSize - currLoc))
+              lift $ Just $ (seqs Seq.>< endPad)
+          )
+  shapeSeq (MDwarf.PointerType _ maybeRef) =
+    let mshape = constructPtrMemShapeFromRef sprog =<< (lift $ maybeRef)
+     in Seq.singleton <$> mshape
+  shapeSeq _ = lift $ Nothing
+
+type VisitState = Map.Map MDwarf.TypeRef Int
```
langston-barrett: FWIW I personally tend to avoid type synonyms - it's just one more thing to keep track of when reading the code. In a situation like this, I'd probably make this a `newtype` and add dedicated helpers like `visitType :: MDwarf.TypeRef -> StateT VisitState m ()`

grease-cli/src/Grease/Cli.hs
2over12: I think it could have surprising behavior in some cases that lead to over-restricted/unexpected shapes so im not positive default is the answer

grease/src/Grease/Options.hs
```diff
@@ -104,6 +112,8 @@ data SimOpts
   , simEnableDWARFPreconditions :: Bool
   -- ^ Enables parsing DWARF to extract initial shape types. This option is
   -- superseded by `simInitialPreconditions` and does not do anything in LLVM mode.
+  , simTypeUnrollingBound :: TypeUnrollingBound
+  -- ^ Bounds the number of times a recursive pointer will be visited when building shapes from DWARF (bounded to 3 recursive pointers by default).
```
langston-barrett, line 117: Maybe:
```suggestion
  -- ^ Bounds the number of times a recursive type will be visited when building shapes from DWARF (bounded to 3 recursive pointers by default).
```
Also, I think there's a way to get `optparse-applicative` to show the default, rather than hardcoding it in the comment.

grease/src/Grease/Shape.hs
```diff
@@ -487,17 +493,17 @@ nullPtr =
...
-constructPtrMemShapeFromRef :: HasPtrWidth w => Subprogram -> MDwarf.TypeRef -> StateT VisitState Maybe (MemShape w NoTag)
-constructPtrMemShapeFromRef sprog ref =
+constructPtrMemShapeFromRef :: HasPtrWidth w => TypeUnrollingBound -> Subprogram -> MDwarf.TypeRef -> StateT VisitState Maybe (MemShape w NoTag)
+constructPtrMemShapeFromRef tyUnrollBound sprog ref =
   do
     mp <- get
     let ct = fromMaybe 0 $ Map.lookup ref mp
     _ <- put (Map.insert ref (ct + 1) mp)
-    if ct > 3
+    if ct > coerce tyUnrollBound
```
langston-barrett: I think this would be more clear just pattern-matching on `TypeUnrollingBound`, which is also zero-cost.

grease/src/Grease/Shape.hs
```diff
@@ -476,7 +482,7 @@ constructPtrTarget sprog tyApp =
               lift $ Just $ (seqs Seq.>< endPad)
           )
   shapeSeq (MDwarf.PointerType _ maybeRef) =
-    let mshape = constructPtrMemShapeFromRef sprog =<< (lift $ maybeRef)
+    let mshape = constructPtrMemShapeFromRef tyUnrollBound sprog =<< (lift $ maybeRef)
```
langston-barrett:
```suggestion
    let mshape = constructPtrMemShapeFromRef tyUnrollBound sprog =<< lift maybeRef
```

grease/src/Grease/Shape.hs
```diff
@@ -487,17 +493,17 @@ nullPtr =
...
     mp <- get
     let ct = fromMaybe 0 $ Map.lookup ref mp
     _ <- put (Map.insert ref (ct + 1) mp)
-    if ct > 3
+    if ct > coerce tyUnrollBound
       then
         pure $ nullPtr
       else
-        let mshape = constructPtrTarget sprog =<< (lift $ extractType sprog ref)
+        let mshape = constructPtrTarget tyUnrollBound sprog =<< (lift $ extractType sprog ref)
```
langston-barrett:
```suggestion
        let mshape = constructPtrTarget tyUnrollBound sprog =<< extractType sprog ref
```

grease/src/Grease/Options.hs
RyanGlScott: > Also, I think there's a way to get `optparse-applicative` to show the default, rather than hardcoding it in the comment.

Use [`showDefault`](https://hackage-content.haskell.org/package/optparse-applicative-0.19.0.0/docs/Options-Applicative.html#v:showDefault) to do this.

grease/src/Grease/Shape.hs
2over12: on the dwarf side we dont parse dwarf count and the ub is raw DWARF_ops that we would have to evaluate to actually get an upper bound (depending on how the array is declared). im not planning on handling this in this PR probably

grease/src/Grease/Shape.hs
2over12: can fix, the confusion is probably coming from each, i was just marking that we need to make sure to buffer the first element by padding (potentially) in between each element, and then the remainder

grease/src/Grease/Shape.hs
langston-barrett: Oh I didn't realize that "ub" stands for "upper bound"! (It's a bit ambiguous because it could also stand for "undefined behavior".) Perhaps we should file an issue describing what remains to be done and refer to it from here?

grease-cli/src/Grease/Main.hs
2over12: i dont think this works that well with it sitting in the Maybe monad

grease/src/Grease/Shape.hs
2over12: i was just noting to make sure i always call this from places were dwarf is in bytes not bits

grease/src/Grease/Options.hs
2over12: showDefault 👍, on pointer vs type, pointer is careful word choice because this comes into affect by counting recursive pointer visits not the struct for instance. E.g. it will replace the third visit of a pointer with a null

grease-cli/src/Grease/Main.hs
```diff
@@ -269,6 +272,35 @@ withMemOptions opts k =
...
+  Bool ->
+  -- | How much to unroll recursive types
+  TypeUnrollingBound ->
+  -- | Argument names
+  Ctx.Assignment (Const.Const String) tys ->
+  -- | Initial arguments
+  Shape.ArgShapes ext NoTag tys ->
+  -- | Macaw CFG for DWARF
+  MacawCfgConfig arch ->
+  -- | Arch context for abi info
```
langston-barrett:
```suggestion
  -- | Architecture-specific information, needed for ABI info
```

grease-cli/src/Grease/Main.hs
```diff
@@ -269,6 +272,35 @@ withMemOptions opts k =
...
    in k
 
+loadDwarfPreconditions ::
+  ExtShape ext ~ PtrShape ext w =>
+  Mem.HasPtrWidth w =>
+  Symbolic.SymArchConstraints arch =>
+  -- | Is DWARF enabled
+  Bool ->
+  -- | How much to unroll recursive types
+  TypeUnrollingBound ->
```
langston-barrett: I definitely appreciate the docs - this one is slightly redundant because it is implied by the type itself.

grease-cli/src/Grease/Main.hs
```diff
@@ -598,8 +630,23 @@ simulateMacawCfg la bak fm halloc macawCfgConfig archCtx simOpts setupHook mbCfg
...
   let argNames = fmapFC (Const . getValueName) rNameAssign
+  -- Is there a better way to do this resolution? We need the pc absent any position independent loading
+  -- so we find the segment that corresponds by checking for all segments if segindex->memsegment -> the segment of the target addr
+  let segIndexToSegment = MM.memSegmentIndexMap memory
+  let targetSegment = fst <$> ((\seg -> find (\(_, seg2) -> seg == seg2) (Map.toList segIndexToSegment)) =<< (MM.segoffSegment <$> mbCfgAddr))
+  let mbSegment =
+        ( do
+            bin <- mcElf macawCfgConfig
+            find (\phdr -> (Just $ Elf.phdrSegmentIndex phdr) == targetSegment) $ headerPhdrs bin
+        )
```
langston-barrett: Unnecessary parens

grease-cli/src/Grease/Main.hs
```diff
@@ -598,8 +630,23 @@ simulateMacawCfg la bak fm halloc macawCfgConfig archCtx simOpts setupHook mbCfg
...
+  let mbNewAddr =
+        ( \(segHdr, entAddr) ->
+            (MM.memWordValue $ MM.segoffOffset entAddr) + fromIntegral (Elf.phdrSegmentVirtAddr segHdr)
+        )
+          <$> ((,) <$> mbSegment <*> mbCfgAddr)
```
langston-barrett:
```suggestion
  let mbNewAddr = do
        segHdr <- mbSegment
        entAddr <- mbCfgAddr
        Just (MM.memWordValue $ MM.segoffOffset entAddr) + fromIntegral (Elf.phdrSegmentVirtAddr segHdr)
```

grease-cli/src/Grease/Main.hs
```diff
@@ -598,8 +630,23 @@ simulateMacawCfg la bak fm halloc macawCfgConfig archCtx simOpts setupHook mbCfg
...
+  dwarfedArgs <- loadDwarfPreconditions (simEnableDWARFPreconditions simOpts) (simTypeUnrollingBound simOpts) argNames initArgs_ macawCfgConfig archCtx mbNewAddr
```
langston-barrett: This line is pretty long, maybe let-bind the sim* options to short names?

grease-cli/tests/sanity/pass/dwarf-shape-recursive/test.c
```diff
@@ -0,0 +1,19 @@
+
```
langston-barrett, line 7: Could use a comment explaining what is being tested

grease-cli/tests/sanity/pass/dwarf-shape-recursive/test.c
```diff
@@ -0,0 +1,19 @@
...
+  while (curr) {
+    sum += curr->x;
+    curr = curr->next;
+  }
+  return sum;
+}
+
+// all: flags {"--symbol", "sum_list", "--enable-dwarf-preconditions", "--no-heuristics"}
+// x64: go(prog) 
+// x64: ok()
```
langston-barrett: Perhaps add some tests with different values for recursive type unrollings?

langston-barrett: Also FWIW it might be nice to directly test the generated shapes, along the lines of https://github.com/GaloisInc/grease/pull/254/files#diff-f49829d9043ff039dfb0971ac8c722f5efb92871f4a6ab08e91082b9dbaf2e90R109

grease/src/Grease/Macaw/Arch.hs
```diff
@@ -245,6 +246,10 @@ data ArchContext arch = ArchContext
   -- Currently, x86-64 is the only supported architecture that does something
   -- non-trivial here. All other architectures can simply leave the stack
   -- pointer by implementing @'_archOffsetStackPointerPostCall' = 'pure'@.
+  , _archABIParams :: [C.Some (MC.ArchReg arch)]
+  -- ^ The ordered integer and pointer ABI registers for the architecture. I
+  -- TODO: In the future this should handle things like stack params https://github.com/GaloisInc/grease/issues/260
+  -- TODO: In the future this should handle multiple calling conventions. https://github.com/GaloisInc/grease/issues/261
```
langston-barrett: https://galoisinc.github.io/grease/dev/style.html#numbered-todos
```suggestion
  -- TODO(#260): In the future this should handle things like stack params.
  -- TODO(#261): In the future this should handle multiple calling conventions.
```
We have a script that checks for stale TODOs in CI, but it only works for this format. Plus, this is more concise.

grease/src/Grease/Macaw/PLT.hs
```diff
@@ -112,15 +112,18 @@ pltStubSymbols pltStubInfo loadOptions ehi =
...
 
   withinSection ::
     (Ord (Elf.ElfWordType w), Num (Elf.ElfWordType w)) =>
-    MM.MemWord w -> Elf.ElfSection (Elf.ElfWordType w) -> Bool
+    MM.MemWord w ->
+    Elf.ElfSection (Elf.ElfWordType w) ->
+    Bool
   withinSection addr section =
     let addr' :: Elf.ElfWordType w
         addr' = fromIntegral addr
```
langston-barrett, line 129: Why did formatting change in this file? Is there something wrong with our Formolu setup?

grease/src/Grease/Options.hs
```diff
@@ -25,6 +27,12 @@ import Grease.Macaw.PLT
 import Grease.Requirement (Requirement)
 import Grease.Solver (Solver (..))
 
+newtype TypeUnrollingBound = TypeUnrollingBound Int
+  deriving Show
```
langston-barrett, line 31:
```suggestion
-- | See 'simTypeUnrollingBound'
newtype TypeUnrollingBound = TypeUnrollingBound Int
  deriving Show
```

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +340,270 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+                    Left $
+                      TypeMismatch
+                        { typeMismatchName = nm
+                        , expectedType = C.Some ty
+                        , foundType = C.Some ty'
+                        }
+      Nothing -> Right s
+
+isInSubProg ::
+  -- | the program counter from the base of the target ELF
```
langston-barrett: I'm not sure how to understand this - Can you say more?

grease/src/Grease/Shape.hs
langston-barrett: I see that you resolved this conversation, but I'm still thinking that this might be helpful

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +340,270 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+  HasPtrWidth w =>
+  TypeUnrollingBound ->
+  Subprogram ->
+  MDwarf.TypeApp ->
+  StateT VisitState Maybe (PtrTarget w NoTag)
+constructPtrTarget tyUnrollBound sprog tyApp =
+  ptrTarget Nothing <$> shapeSeq tyApp
+ where
+  ishape w = lift $ Just $ Seq.singleton $ Initialized NoTag (toBytes w)
+  padding :: (Integral a) => a -> MemShape w NoTag
```
langston-barrett: Unnecessary parens

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +340,270 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+  padding w = Uninitialized (toBytes w)
+  -- if we dont know where to place members we have to fail/just place some bytes
+  buildMember :: HasPtrWidth w => Word64 -> MDwarf.Member -> StateT VisitState Maybe (Word64, Seq.Seq (MemShape w NoTag))
+  buildMember loc member =
+    do
+      memLoc <- lift $ MDwarf.memberLoc member
+      padd <- lift $ Just (if memLoc == loc then Seq.empty else Seq.singleton (padding $ memLoc - loc))
+      memberShape <- shapeOfTyApp $ MDwarf.memberType member
+      let memByteSize = sum $ memShapeSize ?ptrWidth <$> memberShape
```
langston-barrett: Since this is a sequence, perhaps
```suggestion
      memberShapes <- shapeOfTyApp $ MDwarf.memberType member
      let memByteSize = sum $ memShapeSize ?ptrWidth <$> memberShapes
```

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +340,270 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+
+  shapeSeq :: (HasPtrWidth w) => MDwarf.TypeApp -> StateT VisitState Maybe (Seq.Seq (MemShape w NoTag))
+  shapeSeq (MDwarf.UnsignedIntType w) = ishape w
+  shapeSeq (MDwarf.SignedIntType w) = ishape w
+  shapeSeq MDwarf.SignedCharType = ishape (1 :: Int)
+  shapeSeq MDwarf.UnsignedCharType = ishape (1 :: Int)
+  -- Need modification to DWARF to collect DW_TAG_count and properly evaluate dwarf ops for upper bounds in subranges
+  -- https://github.com/GaloisInc/grease/issues/263
+  shapeSeq (MDwarf.ArrayType _elTy _ub) = lift $ Nothing
+  -- need to compute padding between elements, in-front of the first element, and at end
```
langston-barrett: Just for my own edification: I didn't realize that there could be padding before the first element of a struct. In what situations does that arise?

grease/src/Grease/Shape.hs
langston-barrett: I see that you resolved this - Is it because don't agree? If so, can you say why?

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +340,270 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+  do
+    mp <- get
+    let ct = fromMaybe 0 $ Map.lookup ref mp
+    _ <- put (Map.insert ref (ct + 1) mp)
+    if ct > coerce tyUnrollBound
+      then
+        pure $ nullPtr
+      else
+        let memShape = constructPtrTarget tyUnrollBound sprog =<< (lift $ extractType sprog ref)
+         in (\x -> pure $ Pointer NoTag (Offset 0) x) =<< memShape
```
langston-barrett: Perhaps:
```suggestion
      else do
        ty <- lift $ extractType sprog ref
        ptrTgt <- constructPtrTarget tyUnrollBound sprog ty
        pure (Pointer NoTag (Offset 0) ptrTgt)
```

grease/src/Grease/Shape.hs
langston-barrett: Looks like the Haddock for fromDwarfInfo expands on this - perhaps refer to that?

grease-cli/src/Grease/Main.hs
langston-barrett: Would it perhaps be worth wrapping this logic up as a function that lives beside loadDwarfPreconditions? It could even return a newtype to make it clear how the two are connected.

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +340,270 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+            let rs = cuRanges x
+             in let isInCU =
+                      any
+                        ( \range ->
+                            let begin = rangeBegin range
+                                end = rangeEnd range
+                             in begin <= addr && addr < end
+                        )
+                        rs
+                 in isInCU
```
langston-barrett: Sorry, my previous suggestion was unclear. I was suggesting the following:
```suggestion
             in let isInCU range =
                      let begin = rangeBegin range
                          end = rangeEnd range
                       in begin <= addr && addr < end
                 in any isInCU rs
```

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +340,270 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+                  Nothing ->
+                    Left $
+                      TypeMismatch
+                        { typeMismatchName = nm
+                        , expectedType = C.Some ty
+                        , foundType = C.Some ty'
+                        }
+      Nothing -> Right s
+
+isInSubProg ::
```
langston-barrett: This is defined pretty far away from its (single) use in fromDwarfInfo - Could it perhaps live in a where-clause on that function?

grease/src/Grease/Shape.hs
2over12: i personally wouldnt be a big fan of 
```
case tyUnrollBound of
 TypeUnrollingBound(x) | x > 3 -> ...
 _ -> ...
```

a bit verbose for my taste but lmk

grease/src/Grease/Shape.hs
2over12: i have no idea if this is possible or exists somewhere. more an artifact of the fact that dwarf would allow such a case to occur simply by declaring no member at offset 0 and we have to do "something"

grease/src/Grease/Shape.hs
langston-barrett: What about pattern-matching on it in the argument list? Or adding an accessor `getTypeUnrollingBound` and using it here?

Also, I don't think the `x > 3` would be necessary in either case?

grease/src/Grease/Shape.hs
langston-barrett: Fair enough, perhaps worth saying that in the comment.

grease/src/Grease/Macaw/PLT.hs
2over12: both are valid under formulo maybe i hit enter by accident im not sure?

2over12: idk im unsure actually

langston-barrett: Okay, let's just keep an eye on it.

grease/src/Grease/Shape.hs
2over12: err `ct > x` but fine ill match it in the parameter

grease-cli/src/Grease/Main.hs
```diff
@@ -269,6 +273,61 @@ withMemOptions opts k =
           }
    in k
 
+-- \| Find the pc for DWARF which is relative to the ELF absent its base load address.
```
langston-barrett: Seems the haddock for this was escaped?

grease-cli/src/Grease/Main.hs
```diff
@@ -269,6 +273,61 @@ withMemOptions opts k =
...
+-- We find this address by finding the segment that corresponds to the target address
+--  by checking for all ELF segments (mapped to Macaw segments)
+-- and finding the ELF segment that corresponds to the Macaw segment for the target address.
```
langston-barrett: nit: Some weird whitespace and line breaks here

grease-cli/tests/sanity/pass/dwarf-shape-recursive/test.c
```diff
@@ -0,0 +1,33 @@
...
+
+// all: flags {"--symbol", "sum_list", "--enable-dwarf-preconditions", "--no-heuristics", "--type-unrolling-bound", "12"}
+// x64: go(prog) 
+// x64: ok()
+
+// all: flags {"--symbol", "sum_list", "--enable-dwarf-preconditions", "--no-heuristics", "--type-unrolling-bound", "0"}
+// x64: go(prog) 
+// x64: ok()
```
langston-barrett: The assertions here don't seem to meaningfully differ from the test for the default value, so I think the test could still pass even if this flag didn't do anything. It would be nice to test the generated shapes, as is done here:

https://github.com/GaloisInc/grease/pull/254/files#diff-f49829d9043ff039dfb0971ac8c722f5efb92871f4a6ab08e91082b9dbaf2e90R64

grease-cli/tests/sanity/pass/dwarf-shapes/test.c
```diff
@@ -0,0 +1,14 @@
...
+
+struct bar {
+  int x;
+};
+
+int foo(int x, int *y, struct bar *z) { return x + *y + z->x; }
+
+// all: flags {"--symbol", "foo", "--enable-dwarf-preconditions", "--no-heuristics"}
+// x64: go(prog)
+// x64: ok()
```
langston-barrett, line 14: Most of our tests use the entrypoint `test`. Since `foo` isn't a significant name, perhaps let's use `test` here too

grease/src/Grease/Macaw/Arch.hs
```diff
@@ -245,6 +246,10 @@ data ArchContext arch = ArchContext
   -- Currently, x86-64 is the only supported architecture that does something
   -- non-trivial here. All other architectures can simply leave the stack
   -- pointer by implementing @'_archOffsetStackPointerPostCall' = 'pure'@.
+  , _archABIParams :: [C.Some (MC.ArchReg arch)]
+  -- ^ The ordered integer and pointer ABI registers for the architecture. I
+  -- TODO(260): In the future this should handle things like stack params.
+  -- TODO(261): In the future this should handle multiple calling conventions.
```
langston-barrett:
```suggestion
  -- TODO(#260): In the future this should handle things like stack params.
  -- TODO(#261): In the future this should handle multiple calling conventions.
```

grease/src/Grease/Options.hs
langston-barrett: We should remove this from the comment if it is now handled by `showDefault`.

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +342,267 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+          let endPad = if endLoc >= structSize then Seq.empty else Seq.singleton (padding $ structSize - endLoc)
+          lift $ Just $ (seqs Seq.>< endPad)
+  shapeSeq (MDwarf.PointerType _ maybeRef) =
+    let mshape = constructPtrMemShapeFromRef tyUnrollBound sprog =<< lift maybeRef
+     in Seq.singleton <$> mshape
+  shapeSeq _ = lift $ Nothing
+
+type VisitState = Map.Map MDwarf.TypeRef Int
+
+nullPtr :: HasPtrWidth w => (MemShape w NoTag)
```
langston-barrett: Unnecessary parens

grease/src/Grease/Shape.hs
```diff
@@ -314,3 +342,267 @@ deriving instance (ShowExt ext tag, ShowF tag) => Show (ArgShapes ext tag tys)
...
+nullPtr =
+  let zbyt = TaggedByte{taggedByteTag = NoTag, taggedByteValue = 0}
+   in Exactly (take (fromInteger $ CT.intValue ?ptrWidth `div` 8) $ repeat $ zbyt)
+
+constructPtrMemShapeFromRef :: HasPtrWidth w => TypeUnrollingBound -> Subprogram -> MDwarf.TypeRef -> StateT VisitState Maybe (MemShape w NoTag)
+constructPtrMemShapeFromRef bound@(TypeUnrollingBound tyUnrollBound) sprog ref =
+  do
+    mp <- get
+    let ct = fromMaybe 0 $ Map.lookup ref mp
+    _ <- put (Map.insert ref (ct + 1) mp)
```
langston-barrett: I still think that this would make a nice helper function (perhaps in a `where` clause) but nbd

langston-barrett (approved):
Looks great! I think the main thing I'd like to see is moving this code to its own module `Grease.Macaw.Dwarf`; This would help because:

- It would make it more obvious that this is all specific to binaries
- Generally, more modules enables faster compilatoin (they are compiled separately, and compilation tends to be superlinear)
- It would provide a common home for `addressToELFAddress`, `loadDwarfPreconditions`, and `fromDwarfInfo`
- It would allow all of the above to hide their implementation details / helper types and functions from unrelated code
- It would provide a natural place to put a Haddock describing the process at a high level
- It would reduce the size of `Main` which is already a quite large module

grease/src/Grease/Shape.hs
langston-barrett: Appreciate your flexibility here. FWIW, `coerce` is fairly uncommon, and it's usually used in circumstances where there would be a performance penalty to using the newtype accessor/pattern match (e.g., `fmap getTypeUnrollingBound someList`).

grease/src/Grease/Macaw/Dwarf.hs
```diff
@@ -0,0 +1,305 @@
...
+    let seg = MM.segoffSegment entAddr
+    let targetSegment = fst <$> find (\(_, seg2) -> seg == seg2) (Map.toList segIndexToSegment)
+    let mbSegment = find (\phdr -> (Just $ Elf.phdrSegmentIndex phdr) == targetSegment) $ Elf.headerPhdrs bin
+    let mbNewAddr =
+          do
+            segHdr <- mbSegment
+            pure $ (MM.memWordValue $ MM.segoffOffset entAddr) + fromIntegral (Elf.phdrSegmentVirtAddr segHdr)
+    mbNewAddr
+
+loadDwarfPreconditions ::
```
langston-barrett, line 75: Looks like the Haddock for this was misplaced?

grease/src/Grease/Macaw/Dwarf.hs
```diff
@@ -0,0 +1,305 @@
+{-# LANGUAGE ImplicitParams #-}
+{-# LANGUAGE TypeFamilies #-}
+
+-- | Dwarf shape parsing and initialization
```
langston-barrett: Thanks for moving this into its own module, this is really nice!

It's probably worth including a bit more content here in the module-level Haddock explaining the purpose of this module. Even something as simple as "Bootstrap the refinement loop by constructing initial 'Shape's in a type-directed way from DWARF debug info" would be helpful.

langston-barrett (approved):

langston-barrett:
Hmm, idk why GHA hasn't run the `docs` job. Perhaps we should just disable that branch protection rule for the moment?

langston-barrett:
Okay, changed the ruleset.

PR #265, opened 2025-07-02
Balanced unrolling of recursive pointers in DWARF types

### LINKED ISSUE #264: DWARF Shapes: Separate recursive type instantiation to better handle trees etc
Currently the count for each recursive type ref is stored globally in the state. These should be partitioned by member we visit so that we arent biased and doing a DFS instantiation. If we had a tree structure it would be weighted only on the left hand side.

2over12:
Closes #264

### DIFF
--- a/grease/src/Grease/Macaw/Dwarf.hs
+++ b/grease/src/Grease/Macaw/Dwarf.hs
@@ -210,7 +210,7 @@ pointerShapeOfDwarf ::
 pointerShapeOfDwarf _ _ r _ (MDwarf.SignedIntType _) = intPtrShape r
 pointerShapeOfDwarf _ _ r _ (MDwarf.UnsignedIntType _) = intPtrShape r
 pointerShapeOfDwarf _ tyUnrollBound _ sprog (MDwarf.PointerType _ tyRef) =
-  let memShape = constructPtrTarget tyUnrollBound sprog Map.empty =<< (extractType sprog =<< tyRef)
+  let memShape = constructPtrTarget tyUnrollBound sprog Map.empty =<< extractType sprog =<< tyRef
       pointerShape = ShapePtr NoTag (Offset 0) <$> memShape
    in (C.Some <$> pointerShape)
 pointerShapeOfDwarf _ _ _ _ _ = Nothing


grease-cli/tests/sanity/pass/dwarf-shape-recursive-tree/test.c
```diff
@@ -0,0 +1,34 @@
+/* Copyright (c) Galois, Inc. 2025 */
+
+// Tests that heuristics are unneeded when using dwarf populated shapes to
+// create a memory precondition that allows sum_list to execute.
+// Tests that the recursive type instantiation is bounded
```
langston-barrett, line 5: Looks like this was just copied from the other test?

grease-cli/tests/sanity/pass/dwarf-shape-recursive-tree/test.c
```diff
@@ -0,0 +1,34 @@
...
+  struct tree *rhs;
+};
+
+struct tree {
+  int x;
+  struct treenode *next;
+};
+
+// getting this to only be safe when balanced is a bit annoying
+// we always assume init by one
```
langston-barrett, line 19: I'm not exactly sure what I should take away from this comment.

grease-cli/tests/sanity/pass/dwarf-shape-recursive-tree/test.c
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  return val;
+}
+
+// Checks that we have a 34 null pointer which is on the RHS and larger than would be the case without even unrolling
+
+// all: flags {"--symbol", "sum_tree", "--enable-dwarf-preconditions", "--no-heuristics"} 
+// x64: go(prog)
+// x64: check("000034: XX XX XX XX ## ## ## ## 00 00 00 00 00 00 00 00")
```
langston-barrett, line 34: I believe you that this test is correct, but it's a bit hard to tell from the content of the assertion itself. I would recommend including more of the shape as is done here: https://github.com/GaloisInc/grease/pull/254/files#diff-f49829d9043ff039dfb0971ac8c722f5efb92871f4a6ab08e91082b9dbaf2e90R64

You can just copy-paste the shape printout from `tests/sanity/.../...recursive-tree/test.out`, you don't have to write it yourself.

langston-barrett (approved):
Very nice, I like that this actually *simplifies* the DWARF shape inference code.

Issue #192, opened 2025-05-27
Concretize `Bool`


langston-barrett:
Consider this function:
```
(defun @test ((p (Ptr 64)) (b Bool)) Unit
  (start start:
    (let sz (bv 64 1))
    (let a (alloca none sz))
    (branch b if: else:))
  (defblock if:
    (load none i8 a)
    (return ()))
  (defblock else:
    (load none i8 a)
    (return ())))
```
```
cabal run exe:grease -- test.llvm.cbl

Finished analyzing 'test'. Likely bug: uninitialized stack read at test.llvm.cbl:7:5
Allocated at test.llvm.cbl:4:12

Concretized arguments:

%0: 0000000000000000
```
Where's the concretization for `b`?

(Note that this only affects S-expression programs, as there is no way for a LLVM bitcode module or ELF binary to create a Crucible `Bool`, only `(Ptr 1)`s.)

PR #269, opened 2025-07-02
llvm: Support floating-point types

### LINKED ISSUE #268: Support floating-point values
GREASE does not support floating-point values:
```
(defun @test ((f (FP Float))) Unit
  (start start:
    (return ())))
```
```
cabal run exe:grease -- float.llvm.cbl

grease: Could not determine minimal shape for argument
```
(commit: fa62160)

We should add `Shape`s for floating-point, so that GREASE supports executing more LLVM code out of the box.

langston-barrett:
Fixes #268.

### DIFF
--- a/grease-cli/tests/llvm/types.llvm.cbl
+++ b/grease-cli/tests/llvm/types.llvm.cbl
@@ -1,0 +1,31 @@
+; Copyright (c) Galois, Inc. 2025
+
+; Test that GREASE handles all the Crucible types targeted by Crucible-LLVM.
+
+;; flags {"--symbol", "test"}
+;; go(prog)
+
+(defun @test
+  ((x0 Bool)
+   (x1 (Ptr 1))
+   (x2 (Ptr 8))
+   (x3 (Ptr 16))
+   (x4 (Ptr 32))
+   (x5 (Ptr 64))
+   (x6 (Ptr 128))
+   (x7 (Ptr 256))
+   (x8 (Ptr 512))
+   (x9 (FP Half))
+   (x10 (FP Float))
+   (x11 (FP Double))
+   (x12 (FP Quad))
+   (x13 (FP DoubleDouble))
+   (x14 (FP X86_80))
+   (x15 (Struct Bool Unit))
+   (x16 Unit))
+  Unit
+  (start start:
+    (return ())))
+
+; TODO(#268): Support floats
+;; check "Exception: Could not determine minimal shape for argument"
--- a/grease-cli/tests/llvm/types.llvm.cbl
+++ b/grease-cli/tests/llvm/types.llvm.cbl
@@ -27,5 +27,24 @@
   (start start:
     (return ())))
 
-; TODO(#268): Support floats
-;; check "Exception: Could not determine minimal shape for argument"
+;; check [[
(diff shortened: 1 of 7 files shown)


grease-cli/tests/llvm/types.llvm.cbl
```diff
@@ -0,0 +1,50 @@
...
+;; %5: 000000+0000000000000000
+;; %6: XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX
+;; %7: XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX
+;; %8: XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX XX
+;; %9: Half
+;; %10: Float
+;; %11: Double
+;; %12: Quad
+;; %13: DoubleDouble
+;; %14: X86_80
```
RyanGlScott, line 45: I suppose we might actually want to print these values in the output some day... but I'm fine with not doing that for now.

grease-cli/tests/llvm/types.llvm.cbl
```diff
@@ -0,0 +1,50 @@
...
+   (x2 (Ptr 8))
+   (x3 (Ptr 16))
+   (x4 (Ptr 32))
+   (x5 (Ptr 64))
+   (x6 (Ptr 128))
+   (x7 (Ptr 256))
+   (x8 (Ptr 512))
+   (x9 (FP Half))
+   (x10 (FP Float))
+   (x11 (FP Double))
```
RyanGlScott, line 20: TIL that this is how these types are represented in `crucible-syntax`. We should update [these claims](https://github.com/GaloisInc/grease/blob/0a3846ff9bd0947d7ab27a058c4bd343fe8e01aa/doc/builtins.md#L257-L258) in the documentation, which I think are incorrect in light of this.

grease-cli/tests/llvm/types.llvm.cbl
langston-barrett: There's no real associated value, just the shape. And the shape doesn't hold any information, it just indicates that we should make a fully symbolic float there.

grease-cli/tests/llvm/types.llvm.cbl
langston-barrett: Good call, done

grease-cli/tests/llvm/types.llvm.cbl
RyanGlScott: What happens if we have a counterexample with a concrete floating-point value? Will its value actually get printed then, or will we just use `Float`/`Double`/etc. as a placeholder? (I'm fine with either, I just wanted to understand how it currently behaves.)

langston-barrett: The value will not be printed, the type will be printed as a placeholder. Same as #192.

PR #273, opened 2025-07-08
Support loading initial preconditions from JSON

### LINKED ISSUE #36: Support JSON shape file input
@langston-barrett says:

-----

The shape file format was intended for human consumption and production, but it has found use in the Ghidra plugin. We should also support natively ingesting shapes as JSON.

langston-barrett:
Fixes #36.

Needs:

- [x] Parsing of structs
- [x] Parsing of (and tests for) nontrivial pointers
- [x] Mention in the docs

### DIFF
--- a/grease-cli/tests/x86/extra/json-shapes.json
+++ b/grease-cli/tests/x86/extra/json-shapes.json
@@ -1,0 +1,5 @@
+{
+  "rax": { "type": "ext", "val": { "tag": null, "type": "bv", "width": 64 } },
+  "rcx": { "type": "ext", "val": { "tag": null, "type": "bv-lit", "width": 64, "val": 42 } },
+  "rdx": { "type": "ext", "val": { "tag": null, "type": "ptr", "offset": 4, "target": [{ "type": "uninit", "uninit": 4 }, { "type": "init", "tag": null, "init": 4 }] } }
+}
--- a/grease-cli/tests/x86/extra/json-shapes.json
+++ b/grease-cli/tests/x86/extra/json-shapes.json
@@ -2,7 +2,6 @@
   "rax": {
     "type": "ext",
     "val": {
-      "tag": null,
       "type": "bv",
       "width": 64
     }
@@ -10,7 +9,6 @@
   "rcx": {
     "type": "ext",
     "val": {
-      "tag": null,
       "type": "bv-lit",
       "width": 64,
       "val": 42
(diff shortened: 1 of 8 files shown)


grease-cli/tests/x86/extra/json-shapes.json
```diff
@@ -0,0 +1,5 @@
+{
```
2over12, line 1: nit is this the result of prettying this json? i would expect jq to pull the fields down but nbd either way

grease/src/Grease/Shape.hs
```diff
@@ -397,3 +411,84 @@ replaceShapes names (ArgShapes args) (ParsedShapes replacements) =
...
+      "unit" -> Some . ShapeUnit <$> (parseTag =<< tg)
+      _ -> fail ("Unknown Shape type: " ++ Text.unpack ty)
+ where
+  doParseExt v = do
+    Some s <- parseExt =<< v .: "val"
+    pure (Some (ShapeExt s))
+
+  parseFloat tg v = do
+    Some fi <- parseFloatRepr =<< v .: "info"
+    tag <- parseTag =<< tg
```
2over12: do we even want to be parsing the tag, right now it could be anything and we will const it, seems a bit odd to me, it's not clear when we would ever want things tagged?

grease-cli/tests/x86/extra/json-shapes.json
langston-barrett: No, I can prettyprint it if you like!

grease/src/Grease/Shape.hs
langston-barrett: Are you suggesting we just fix it to `NoTag` everywhere? I think that would be reasonable, but since I already implemented the more general version it seems wasteful to remove it at this point. This way the parser is a bit more robust to future uses, though perhaps slightly more complex.

2over12: yeah i guess that's fair but imo if there's no tag we shouldnt be requiring "tag": 1 so imo when the type is NoTag ideally we wouldnt be fetching the tg at all

2over12: it's a bit unclear how to do this easily if you want to maintain the ability to parse the tag if you want to and make the types line up nicely, a bit hard to abstract over

2over12: a possible way i guess would be to pass the full object to parseTag rather than the tag, then the NoTag parser can elect not to access the tag field?

langston-barrett: Good call, thanks! Done.

PR #274, opened 2025-07-09
Bump Crucible submodule, use new symbolic string-loading functions

langston-barrett:
Brings in https://github.com/GaloisInc/macaw/pull/513 and https://github.com/GaloisInc/crucible/pull/1460, removes some hand-rolled string loading functionality.

### DIFF
--- a/grease/src/Grease/Macaw/Memory.hs
+++ b/grease/src/Grease/Macaw/Memory.hs
@@ -23,8 +23,8 @@ import Data.ByteString qualified as BS
 import Data.Macaw.CFG qualified as MC
 import Data.Macaw.Memory qualified as MM
 import Data.Macaw.Symbolic qualified as Symbolic
-import Data.Macaw.Symbolic.Backend qualified as Symbolic
 import Data.Macaw.Symbolic.Concretize qualified as Symbolic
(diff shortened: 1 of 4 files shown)


langston-barrett:
Ok, I generalized the API exported by Crucible-LLVM and wrapped it in macaw-symbolic, and this appears to be working as expected.

grease/src/Grease/Macaw/Memory.hs
```diff
@@ -47,13 +47,12 @@ loadString ::
   , Mem.HasLLVMAnn sym
   , ?memOpts :: Mem.MemOptions
   ) =>
+  MM.AddrWidthRepr (MC.ArchAddrWidth arch) ->
```
RyanGlScott: Why do we need to pass this in explicitly? Shouldn't we be able to just call `let addrWidth = MC.addrWidthRepr ?ptrWidth`, as was done before?

langston-barrett: Good point, and actually the situation is better than that: We don't even need to plumb `ArchInfo` around anymore, because we were only using it for pointer width (which is implicit) and endianness, which is not actually necessary for loading bytes one at a time. Good observation, this let me remove that parameter in a lot of places!

PR #275, opened 2025-07-14
Add some docs and support for ghidra script type stubs

2over12:
Potentially in the future we want to move to a gradle setup with a plugin bound to a ghidra install dir. I tend to find that type of setup more maintainable for larger projects etc but for the moment this adds some docs on how to setup pyi stubs for ghidra

I am glad upstream now distributes these rather than having to generate them yourself

### DIFF
--- a/doc/dev/ghidra-script-dev.md
+++ b/doc/dev/ghidra-script-dev.md
@@ -12,3 +12,7 @@ pip install ghidra-stubs
 ```
 
 Editors that support type stubs will be able to use this virtualenv to provide partial autocomplete support.
+
+## Linting
+
+See the [linting section of the dev docs](dev.md) for discussion of how to lint this script with ruff.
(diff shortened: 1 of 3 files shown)


doc/dev/ghidra-script-dev.md
```diff
@@ -0,0 +1,14 @@
+# Ghidra Script Development with Type Stubs
```
thebendavis, line 1: Thanks for adding these docs - please link this file, perhaps from `SUMMARY.md`, so it shows up on https://galoisinc.github.io/grease/

thebendavis (approved):
Nice - thanks!

doc/dev/ghidra-script-dev.md
thebendavis: Hmm, actually `dev.md` already has some Ghidra/Python docs (`ruff` notes), maybe we should refactor and group Haskell dev vs Ghidra/Python dev information, or something else? I don't have a strong opinion, other than since there isn't much "working on ghidra plugin stuff" info ideally it's not spread too far apart as otherwise it'd be easy to miss.

2over12: I think maybe im going to preference adding the ghidra script dev to the summary then linking to the linting docs from there. I think the ruff stuff belongs as part of the linting stuff but also want it to be visible from the ghidra notes

thebendavis: Works for me - thanks!

PR #277, opened 2025-07-14
Fix type stub import in jython

2over12:


ghidra_scripts/grease.py
```diff
@@ -35,7 +35,17 @@
...
 import tempfile
-import typing
+
+
+try:
+    import typing
+except ImportError:
+
+    class FakeObj:
+        TYPE_CHECKING = False
```
RyanGlScott, line 48: My understanding is that `TYPE_CHECKING` is a special constant that is assumed to be true by Python typecheckers. Is this correct? If so, it would be worth leaving a comment stating why we both to set this to `False`.

thebendavis (approved):

Issue #283, opened 2025-07-18
Print pointers with empty memshapes differently


2over12:
Right now we will print a pointer to an empty memshape as:
```
rdi: 000008+0000000000000000

000008: 
```

But then there actually isnt an object allocated for 0008 because the ArgShape actually initializes rdi to a llvm bv pointer. I think this is a bit confusing for users (and me since I got fooled by it for a second 😆 )

since rdi is treated effectively as a symbolic bv ptr maybe it would be better to print it how those are printed?

This behavior can be observed with the program from https://github.com/GaloisInc/grease/issues/279, where an error will occur in the 0x0 allocation despite rdi being initialized as:

```
rdi: 000008+0000000000000000

000008: 
```

PR #284, opened 2025-07-21
Use macaw memory model errors for refinement

### LINKED ISSUE #279: False positive with must-fail heuristic on Macaw pointer dereferences
GREASE gives a pretty poor result on the following very simple Macaw S-expression program that dereferences `rdi + 4`. This is ultimately due to our lack of fine-grained tracking of errors arising from the Macaw memory model (https://github.com/GaloisInc/macaw/issues/429).
```
(defun @test ((regs X86Regs)) X86Regs
  (start start:
    (let ptr0 (get-reg rdi regs))
    (let offset (bv 64 4))
    (let ptr1 (pointer-add ptr0 offset))
    (let _ (pointer-read (Bitvector 8) le ptr1))
    (return regs)))
```
```
cabal run exe:grease -- --symbol test bad.x64.cbl

Finished analyzing 'test'. Likely bug: unavoidable error (safety condition is unsatisfiable) at internal
internal: error: in test
PointerRead outside of static memory range (known BlockID 0): bvSum crdi_offset@69:bv 0x4:[64]

Concretized arguments:

rcx: 0000000000000000
rdx: 0000000000000000
rsi: 0000000000000000
rdi: 0000000000000000
r8: 0000000000000000
r9: 0000000000000000
r10: 0000000000000000
```

(commit: 0f0c1541f2b8d6f84355382b6baec0e140e095ba)

2over12:
Closes #279. This in theory should fix 279 once we add a ptr heuristic for the unmapped error

- [x] Add ptr conversion heuristic for the unmapped global error (or do something else per https://github.com/GaloisInc/grease/issues/283)
- [x] Fix bug where top level callback is used (this param def is called https://github.com/GaloisInc/grease/blob/027e05f613b401fbe8bb7c59ca4bbf8a99eef77c/grease-cli/src/Grease/Main.hs#L553 instead of https://github.com/GaloisInc/grease/blob/027e05f613b401fbe8bb7c59ca4bbf8a99eef77c/grease-cli/src/Grease/Main.hs#L690
- [x] Add docs and regression test


grease-cli/tests/x86/write-to-fixed-address-in-rax.x64.cbl
```diff
@@ -0,0 +1,15 @@
...
+;; go(prog)
+
+(defun @test ((regs X86Regs)) X86Regs
+  (start start:
+    (let rax-ptr (get-reg rax regs))
+    (let zero64 (bv 64 0))
+    (pointer-write (Bitvector 64) le rax-ptr zero64)
+    (return regs)))
+    
+;; no_heuristic()
```
langston-barrett, line 15: Should there be a TODO associated with this? Seems bad that we can't handle it?

grease-cli/tests/x86/write-to-fixed-address-in-rax.x64.cbl
```diff
@@ -0,0 +1,15 @@
+; Copyright (c) Galois, Inc. 2025
+
+; A test case that writes to a pointer stored in %rax.
+
+;; flags {"--symbol", "test", "--initial-precondition", "tests/x86/extra/make-rax-deadbeef.txt"}
```
langston-barrett: Could use `--arg-u64 rax:0x0bad1deadeadbeef` instead, to make this test self-contained?

2over12:
> A piece im not totally sold on is the following:
> 
> I made the refinement tests now pass without losing mustfail by adding heuristic handling for unmapped memory error. I am a bit worried about this heuristic being a bit too broad and refining literals. The only reason this [test](https://github.com/GaloisInc/grease/blob/030a508162e33127737ea7166a7949899bc86fa3/grease-cli/tests/x86/write-to-fixed-address-in-memory.x64.cbl#L6) passes is because the bv lit is not annotated so the heuristics arent run.
> 
> Im a bit worried that by broadly running the heuristics against any UnmappedMemory error we could end up refining a literal or something a long those lines but i dont see a way to make this happen. We are implicitly avoiding such cases by not having the pointer literal annotated

I think short term this is ok. We should also open a PR that only labels mustfail if all bugs are unfixable since mustfail still is not accurate. in the future i think making initial states OX would be better

grease/src/Grease/Refine/Diagnostic.hs
```diff
@@ -124,8 +126,9 @@ instance PP.Pretty Diagnostic where
           , PP.indent 4 (C.ppSimError $ lp ^. W4.labeledPredMsg)
           , -- , PP.indent 4 (W4.ppExpr $ lp ^. W4.labeledPred)
             case minfo of
+              Just (MacawMemError (UnmappedGlobalMemoryAccess _)) -> "<Unmapped memory>"
```
langston-barrett: Why the angle brackets? Also, should we perhaps print the bitvector? In fact, should there maybe be a `Pretty` instance upstream, so that we could just call `pretty` here?

grease/src/Grease/Concretize.hs
```diff
@@ -133,17 +134,29 @@ data ConcretizedData sym ext argTys
...
+  concretize :: forall sym t st fs. (WI.IsExprBuilder sym, sym ~ W4.ExprBuilder t st fs) => sym -> W4.GroundEvalFn t -> e sym -> IO (e sym)
+
+instance Concretize ErrorDescription where
+  concretize sym (W4.GroundEvalFn gFn) (CrucibleLLVMError bb cs) = do
+    bb' <- Mem.concBadBehavior sym gFn bb
+    pure (CrucibleLLVMError bb' cs)
+  concretize sym (W4.GroundEvalFn gFn) (MacawMemError memerr) = do
+    (UnmappedGlobalMemoryAccess ptrVal) <- pure memerr
+    cptr <- Mem.concPtr sym gFn ptrVal
+    pure $ MacawMemError (UnmappedGlobalMemoryAccess cptr)
```
langston-barrett: Why a typeclass instead of a simple function `concretizeErrorDescription`? Such a function could live in [ErrorDescription.hs](https://github.com/GaloisInc/grease/pull/284/files#diff-1a24746c2d30c235e89c047b70d5f63ed6b4d26bc185ca1061548650d1d5fdf7)

grease/src/Grease/ErrorDescription.hs
```diff
@@ -0,0 +1,11 @@
...
+
+import Data.Macaw.Symbolic.Memory qualified as MSM
+import Lang.Crucible.LLVM.Errors qualified as CLLVM
+import Lang.Crucible.LLVM.MemModel.CallStack qualified as LLCS
+
+-- | An error either from the underlying LLVM memory model or
+-- from Macaw.
+data ErrorDescription sym
+  = CrucibleLLVMError (CLLVM.BadBehavior sym) LLCS.CallStack
+  | MacawMemError (MSM.MacawError sym)
```
langston-barrett, line 47: Perhaps worth a `Pretty` instance?

grease/src/Grease/Heuristic.hs
```diff
@@ -425,32 +441,39 @@ pointerHeuristics ::
...
-                Mem.UnwritableRegion -> onePtrHeuristics dst
-                _ -> do
-                  r1 <- onePtrHeuristics dst
-                  r2 <- onePtrHeuristics src
-                  pure (mergeResultsOptimistic r1 r2)
-        Nothing -> pure Unknown
+  [ f
   ]
+ where
+  f bak anns initMem obligation minfo argNames args =
```
langston-barrett: This is a bit odd, why did this change from a lambda to being in a `where`-clause?

grease/src/Grease/MustFail.hs
```diff
@@ -118,9 +112,9 @@ checkOneMustFail ::
   ) =>
   bak ->
   -- | Predicates for which no heuristic succeeded
-  [(C.ProofObligation sym, Maybe (Mem.BadBehavior sym))] ->
+  [(C.ProofObligation sym, Maybe (ErrorDescription sym))] ->
   IO Bool
 checkOneMustFail bak failed =
-  if List.any (Tuple.uncurry excludeMustFail) failed
+  if List.any (Tuple.uncurry $ excludeMustFail) failed
```
langston-barrett: Let's revert this

grease/src/Grease/Refine.hs
```diff
@@ -219,6 +222,20 @@ combiner = C.Combiner $ \mr1 mr2 -> do
         ProveNoHeuristic errs2 ->
           pure (failed (ProveNoHeuristic (errs1 <> errs2)))
 
+type MacawAssertionCallback sym = sym -> W4.Pred sym -> MSM.MacawError sym -> IO (W4.Pred sym)
+type LLVMMemModelCallback sym = LLCS.CallStack -> Mem.BoolAnn sym -> CLLVM.BadBehavior sym -> IO ()
+
+buildErrMaps :: forall t st fs bldr. (bldr ~ (W4.ExprBuilder t st fs)) => IO (IORef (Map.Map (Nonce t C.BaseBoolType) (ErrorDescription bldr)), LLVMMemModelCallback bldr, MacawAssertionCallback bldr)
```
langston-barrett: - Could use a Haddock
- You probably don't need to explicitly quantify over all the type variables
- There are superfluous parens in the constraint
- I get nervous once we get up to 3-tuples. Might be worth making a record type that this function returns,  especially since you've already gone through the trouble of naming the types of the callbacks.
- The standard name for what you've called `bldr` is `sym`

grease/src/Grease/Refine.hs
```diff
@@ -289,7 +306,7 @@ consumer bak anns execResult la heuristics argNames argShapes bbMap initState =
               RefinedPrecondition fc' -> pure $ ProveRefine fc'
               Unknown -> runHeuristics hs fc
           runHeuristics [] _ =
-            let err = fmap Tuple.snd minfo
+            let err = minfo
```
langston-barrett: Inline?

grease/src/Grease/Heuristic.hs
2over12: this is because we need a type signature on applyToErr to abstract out the two calls to applyMemoryHeuristics, let me know if you have a better way but that ends up in a situation where we now want to bind that applyErr under the lambda with a type sig hence the where clause

grease-cli/src/Grease/Main.hs
```diff
@@ -1255,10 +1253,9 @@ simulateLlvmCfg la simOpts bak fm halloc llvmCtx llvmMod initMem setupHook mbSta
       let typeCtx = llvmCtx ^. Trans.llvmTypeCtx
       let dl = TCtx.llvmDataLayout typeCtx
       (args, setupMem, setupAnns) <- setup la bak dl valueNames argTys argShapes initMem
-      bbMapRef <- liftIO (newIORef Map.empty)
-      let ?recordLLVMAnnotation =
-            \callStack (Mem.BoolAnn ann) bb ->
-              modifyIORef bbMapRef $ Map.insert ann (callStack, bb)
+      ErrorCallbacks{errorMap = bbMapRef, llvmErrCallback = recordLLVMAnnotation, macawAssertionCallback = processMacawAssert} <- liftIO buildErrMaps
```
langston-barrett: Long line, consider just using the record accessors below

grease/src/Grease/ErrorDescription.hs
```diff
@@ -0,0 +1,25 @@
...
+import Data.Macaw.Symbolic.Memory (MacawError (UnmappedGlobalMemoryAccess))
+import Data.Macaw.Symbolic.Memory qualified as MSM
+import Lang.Crucible.LLVM.Errors qualified as CLLVM
+import Lang.Crucible.LLVM.Errors qualified as Mem
+import Lang.Crucible.LLVM.MemModel.CallStack qualified as LLCS
+import Lang.Crucible.LLVM.MemModel.Pointer qualified as Mem
+import What4.Expr qualified as W4
+import What4.Interface qualified as WI
+
+concretizeErrorDescription :: (WI.IsExprBuilder sym, sym ~ W4.ExprBuilder t st fs) => sym -> W4.GroundEvalFn t -> ErrorDescription sym -> IO (ErrorDescription sym)
```
langston-barrett: Long line, maybe break it up?

grease/src/Grease/ErrorDescription.hs
```diff
@@ -0,0 +1,25 @@
...
+concretizeErrorDescription sym (W4.GroundEvalFn gFn) (CrucibleLLVMError bb cs) = do
+  bb' <- Mem.concBadBehavior sym gFn bb
+  pure (CrucibleLLVMError bb' cs)
+concretizeErrorDescription sym (W4.GroundEvalFn gFn) (MacawMemError memerr) = do
+  (UnmappedGlobalMemoryAccess ptrVal) <- pure memerr
+  cptr <- Mem.concPtr sym gFn ptrVal
+  pure $ MacawMemError (UnmappedGlobalMemoryAccess cptr)
```
langston-barrett: If you used a lambda-case on the final argument, you wouldn't need to repeat the function and parameter names.

grease/src/Grease/Refine.hs
```diff
@@ -219,6 +223,29 @@ combiner = C.Combiner $ \mr1 mr2 -> do
...
 
+data ErrorCallbacks sym t
+  = ErrorCallbacks
+  { llvmErrCallback :: LLCS.CallStack -> Mem.BoolAnn sym -> CLLVM.BadBehavior sym -> IO ()
+  , macawAssertionCallback :: sym -> W4.Pred sym -> MSM.MacawError sym -> IO (W4.Pred sym)
+  , errorMap :: IORef (Map.Map (Nonce t C.BaseBoolType) (ErrorDescription sym))
+  }
+
+-- | Builds a mapping from assertions to errors if those assertions are unprovable
+-- there are two types of errors 'CLLVM.BadBehavior' which describes errors emitted from the llvm memory model/semantics
```
langston-barrett:
```suggestion
-- | Builds a mapping from assertions to errors if those assertions are unprovable.
--
-- There are two types of errors: 'CLLVM.BadBehavior' which describes errors emitted from the LLVM memory model/semantics
```

grease-cli/src/Grease/Main.hs
2over12: id prefer not to repeat that structure across the multiple times we use that function, the other option would be to CPS the function but i dont really want to force that on the callees

2over12: i split the lines so they arent as long lmk

grease/src/Grease/Refine/Diagnostic.hs
```diff
@@ -76,7 +74,7 @@ data Diagnostic where
     -- | Goal that failed (the negation of this predicate was satisfiable)
     W4.LabeledPred (W4.Pred sym) C.SimError ->
     -- | Description of the problem, if available
-    Maybe (Mem.CallStack, Mem.BadBehavior sym) ->
+    Maybe (Grease.ErrorDescription.ErrorDescription sym) ->
```
langston-barrett: Why fully qualified?

```suggestion
    Maybe (ErrorDescription sym) ->
```

grease/src/Grease/ErrorDescription.hs
```diff
@@ -0,0 +1,68 @@
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+module Grease.ErrorDescription (ErrorDescription (..), concretizeErrorDescription, ppMacawError, ppErrorDesc) where
```
langston-barrett: Long line alert 🚨

grease/src/Grease/ErrorDescription.hs
```diff
@@ -0,0 +1,68 @@
...
+data ErrorDescription sym
+  = CrucibleLLVMError (CLLVM.BadBehavior sym) LLCS.CallStack
+  | MacawMemError (MSM.MacawError sym)
+
+ppDelimitedObj :: [PP.Doc a] -> PP.Doc a
+ppDelimitedObj lst = PP.sep $ "(" : lst ++ [")"]
+
+-- TODO(#310) Pretty print macaw error using a macaw implementation
+ppMacawError :: W4.IsExpr (W4.SymExpr sym) => MSM.MacawError sym -> PP.Doc a
+ppMacawError (MSM.UnmappedGlobalMemoryAccess ptrVal) = PP.sep ["(", "UnmappedGlobalMemoryAccess", CLLVM.ppPtr ptrVal, ")"]
```
langston-barrett: Long line alert 🚨

grease/src/Grease/ErrorDescription.hs
```diff
@@ -0,0 +1,68 @@
...
+
+ppErrorDesc :: W4.IsExpr (W4.SymExpr sym) => ErrorDescription sym -> PP.Doc a
+ppErrorDesc =
+  \case
+    (MacawMemError mmErr) -> ppDelimitedObj ["MacawMemErr", ppMacawError mmErr]
+    (CrucibleLLVMError bb callStack) ->
```
langston-barrett: Unnecessary parens here in the patterns

grease/src/Grease/ErrorDescription.hs
```diff
@@ -0,0 +1,68 @@
...
+ppDelimitedObj lst = PP.sep $ "(" : lst ++ [")"]
+
+-- TODO(#310) Pretty print macaw error using a macaw implementation
+ppMacawError :: W4.IsExpr (W4.SymExpr sym) => MSM.MacawError sym -> PP.Doc a
+ppMacawError (MSM.UnmappedGlobalMemoryAccess ptrVal) = PP.sep ["(", "UnmappedGlobalMemoryAccess", CLLVM.ppPtr ptrVal, ")"]
+
+ppErrorDesc :: W4.IsExpr (W4.SymExpr sym) => ErrorDescription sym -> PP.Doc a
+ppErrorDesc =
+  \case
+    (MacawMemError mmErr) -> ppDelimitedObj ["MacawMemErr", ppMacawError mmErr]
```
langston-barrett: This is for displaying to users, right? I don't think it will be helpful for them to know that the error comes from  Macaw, perhaps just:
```suggestion
    MacawMemError mmErr -> ppMacawError mmErr
```

grease/src/Grease/ErrorDescription.hs
langston-barrett: Generally, `Doc`s are for showing to users. Can we describe this error in a way that's more accessible to users? Perhaps

```suggestion
ppMacawError (MSM.UnmappedGlobalMemoryAccess ptrVal) =
  "Read or write of an unmapped pointer:" PP.<+> CLLVM.ppPtr ptrVal
```

grease/src/Grease/ErrorDescription.hs
```diff
@@ -0,0 +1,68 @@
...
+    (CrucibleLLVMError bb callStack) ->
+      let ppCs = Mem.ppCallStack callStack
+       in PP.vcat $
+            [PP.indent 4 (Mem.ppBB bb)]
+              List.++
+              -- HACK(crucible#1112): No Eq on Doc, no access to cs
```
langston-barrett: What's the TODO here? That issue is closed: https://github.com/GaloisInc/crucible/issues/1112

(This is probably just a leftover TODO)

grease/src/Grease/ErrorDescription.hs
```diff
@@ -0,0 +1,68 @@
...
+              if Mem.null callStack
+                then []
+                else ["in context:", PP.indent 2 ppCs]
+
+instance W4.IsExpr (W4.SymExpr sym) => Show (ErrorDescription sym) where
+  show :: WI.IsExpr (WI.SymExpr sym) => ErrorDescription sym -> String
+  show = show . ppErrorDesc
```
langston-barrett: Do we need a `Show` instance? The intended semantics of `Show` are to provide a machine-readable description (via `Read`). (Though it is commonly misused in our codebases for human-readable prints.)

grease/src/Grease/ErrorDescription.hs
```diff
@@ -0,0 +1,68 @@
...
+
+instance W4.IsExpr (W4.SymExpr sym) => PP.Pretty (ErrorDescription sym) where
+  pretty :: WI.IsExpr (WI.SymExpr sym) => ErrorDescription sym -> PP.Doc ann
+  pretty = PP.viaShow
```
langston-barrett: Why not just
```suggestion
  pretty = ppErrorDesc
```
I think using `viaShow` will prevent this from composing nicely into larger documents (because it will become just a unstructured `String`).

grease/src/Grease/Refine.hs
```diff
@@ -219,6 +223,33 @@ combiner = C.Combiner $ \mr1 mr2 -> do
...
+buildErrMaps :: sym ~ W4.ExprBuilder t st fs => IO (ErrorCallbacks sym t)
+buildErrMaps = do
+  bbMapRef <- newIORef Map.empty
+  let recordLLVMAnnotation = \callStack (Mem.BoolAnn ann) bb ->
+        modifyIORef bbMapRef $ Map.insert ann (CrucibleLLVMError bb callStack)
+  let processMacawAssert = \sym p err -> do
+        (ann, p') <- W4.annotateTerm sym p
+        _ <- modifyIORef bbMapRef $ Map.insert ann (MacawMemError err)
+        pure p'
+  pure ErrorCallbacks{errorMap = bbMapRef, llvmErrCallback = recordLLVMAnnotation, macawAssertionCallback = processMacawAssert}
```
langston-barrett: Long line alert 🚨

grease/src/Grease/Heuristic.hs
langston-barrett: Can you `let`-bind `applyToErr` with a type signature inwide the `BBMemoryError` case, just before `onePointerHeuristics`?

If not, I'd recommend lifting `f` to a top-level function called `pointerHeuristic` and making `pointerHeuristics` a singleton list around it.

2over12: it's not really possible because we need it in MacawMemError as well, you can probably bind a lambda at the top level for the whole function but im just gonna lift f

langston-barrett (approved):

PR #285, opened 2025-07-23
Initialize pointer-sized arguments to bitvectors

### LINKED ISSUE #279: False positive with must-fail heuristic on Macaw pointer dereferences

### LINKED ISSUE #283: Print pointers with empty memshapes differently

langston-barrett:
Previously, we initialized these as pointers (`ShapePtr`) to empty allocations (i.e., empty `PtrTarget`s). However, the code in `Setup` went on to treat such pointers to empty allocations as plain bitvectors (i.e., `LLVMPointer`s with block number 0), which led to some confusing output seen in issue #283.

This sidesteps the confusion by never creating such zero-size pointers in the first place. Instead, initialize pointer-size registers to be symbolic bitvectors (`ShapePtrBV`) from the start. If they are treated as pointers during execution, then make them pointers to (at first) one uninitialized byte.

This doesn't yet fix #279, which will still require #284.

Needs:

- [x] Fix tests
- [x] Adjust note about empty pointer targets

Fixes #283, or at least ameliorates the problem.


grease/src/Grease/Heuristic.hs
```diff
@@ -121,6 +122,12 @@ refinePtrArg ::
   IO (HeuristicResult ext argTys)
 refinePtrArg la args modify sel =
   case args ^. selectArg sel of
+    ShapeExt (ShapePtrBV _tag w) | Just Refl <- testEquality w ?ptrWidth -> do
+      let pt = ptrTarget Nothing Seq.empty
+      doLog la $ Diag.HeuristicPtrTarget pt
+      pt' <- modify pt
+      let args' = args & selectArg sel .~ ShapeExt (ShapePtr NoTag (Offset 0) pt')
```
2over12, line 129: Im trying to work out if this matches up with the "heuristics wont make empty memshapes" piece. Im assuming that the modify provided is always supposed to grow this target allocation, is that always true?

2over12 (approved):

grease/src/Grease/Heuristic.hs
langston-barrett: Yeah, in practice this is always one of the `.*(g|G)rowPtrTarget.*` functions.

Issue #228, opened 2025-06-12
cli: Add flags for specifying initial register values


langston-barrett:
It's a relatively common pattern to write a shapes DSL file that just specifies that one register contains a buffer of a fixed size, and another register contains a literal bitvector that is exactly the length of the buffer. For such simple situations, the shapes DSL is overkill.

Instead, we could provide command-line options to customize the initial content of registers, e.g.,

- `--reg-buf-uninit rdi:10` would put a pointer to an uninitialized buffer of length 10 in `rdi`
- `--reg-buf-init rdi:10` would do the same, but with an initialized buffer
- `--reg-int rsi:10` would put a concrete bitvector with value 10 in `rsi`

Going further, we might want:

- `--reg-str "rdi:some string here"`: put a pointer to a buffer holding the concrete bytes of the given string in `rdi`
- `--reg-buf-bytes "rdi:00 ff ef ..."`: put a pointer to a buffer with the given bytes in `rdi`
- `--arg-* 0:*` (where `*` is replaced by all of the things after `reg-` in the above options) to ask GREASE to do the ABI-specific mapping of arguments to registers, so that e.g., `--arg-int 0:1` would put `1` in `rdi` on x86_64 with SysV.

This would provide a smooth ramp-up of complexity and customizability from specifying no additional arguments to GREASE, to providing some of these CLI opts, to specifying a shapes DSL file, to specifying a startup override.

langston-barrett:
This feature would also be a good target for a Ghidra UI element that sets some of these same parameters. This would add a new level of customizability / power to the Ghidra plug-in, without forcing users to exit their Ghidra-based workflows to manually edit a shapes file.

PR #291, opened 2025-07-24
cli: Add groups of command-line options

langston-barrett:
We have 27 options and some of them fall into recognizable groups. Let's start organizing them, especially since I'm about to add a bunch more in #228. A disadvantage is that all groups appear after all non-grouped options, but I don't really want to add a "default" group, especially since it'd be easy to forget it when adding a new option and then have a lone flag that appears before all groups.

### DIFF
--- a/grease-cli/src/Grease/Cli.hs
+++ b/grease-cli/src/Grease/Cli.hs
@@ -57,14 +57,13 @@ boundedEnumMetavar _ = Opt.metavar $ varShowS ""
 
 entrypointParser :: Opt.Parser Entrypoint
 entrypointParser =
-  inGroup addressParser
-    Opt.<|> inGroup symbolParser
-    Opt.<|> inGroup coreDumpParser
-    Opt.<|> inGroup addressStartupOvParser
-    Opt.<|> inGroup symbolStartupOvParser
+  Opt.parserOptionGroup "Entrypoint options" $
+    addressParser
+      Opt.<|> symbolParser
+      Opt.<|> coreDumpParser
+      Opt.<|> addressStartupOvParser
+      Opt.<|> symbolStartupOvParser
  where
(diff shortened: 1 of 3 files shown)


langston-barrett:
Okay sorry for the churn but this should be ready now.

cabal.project
```diff
@@ -1,3 +1,6 @@
+-- https://github.com/kquick/tasty-sugar/issues/4
+allow-newer: tasty-sugar:optparse-applicative
```
langston-barrett: This has been fixed: https://github.com/kquick/tasty-sugar/issues/4

grease-cli/src/Grease/Cli.hs
```diff
@@ -57,12 +57,14 @@ boundedEnumMetavar _ = Opt.metavar $ varShowS ""
...
-  addressParser
-    Opt.<|> symbolParser
-    Opt.<|> coreDumpParser
-    Opt.<|> addressStartupOvParser
-    Opt.<|> symbolStartupOvParser
+  inGroup addressParser
+    Opt.<|> inGroup symbolParser
+    Opt.<|> inGroup coreDumpParser
+    Opt.<|> inGroup addressStartupOvParser
+    Opt.<|> inGroup symbolStartupOvParser
```
RyanGlScott: Possibly silly `optparse-applicative` question: would it be equivalent to write `Opt.parserOptionGroup "Entrypoint options" (addressParser <|> symbolParser <|> ...)` instead? This would be shorter and avoid duplication.

RyanGlScott (approved):
TIL about `parserOptionGroup`!

grease-cli/src/Grease/Cli.hs
langston-barrett: Yes, it appears that that works!

PR #292, opened 2025-07-25
Documentation and refactoring re: specifying initial preconditions

langston-barrett:
Just some documentation and refactoring in advance of #228.

### DIFF
--- a/grease-cli/src/Grease/Main.hs
+++ b/grease-cli/src/Grease/Main.hs
@@ -553,6 +553,7 @@ macawInitArgShapes la bak archCtx simOpts macawCfgConfig argNames mbCfgAddr = do
   initArgs_ <- minimalArgShapes bak archCtx mdEntryAbsAddr
   let shouldUseDwarf = simEnableDebugInfoPreconditions simOpts
   let getDwarfArgs = do
+        -- Maybe
         elfHdr <- mcElf macawCfgConfig
         addr <- mbCfgAddr
         loadDwarfPreconditions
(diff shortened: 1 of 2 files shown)


grease-cli/src/Grease/Main.hs
```diff
@@ -516,6 +516,56 @@ initialLlvmFileSystem halloc sym simOpts = do
...
+  -- | If simulating a binary, this is 'Just' the address of the user-requested
+  -- entrypoint function. Otherwise, this is 'Nothing'.
+  Maybe (MC.ArchSegmentOff arch) ->
+  IO (ArgShapes (Symbolic.MacawExt arch) NoTag (Symbolic.CtxToCrucibleType (Symbolic.ArchRegContext arch)))
+macawInitArgShapes la bak archCtx simOpts macawCfgConfig argNames mbCfgAddr = do
+  let memory = mcMemory macawCfgConfig
+  let mdEntryAbsAddr = fmap (segoffToAbsoluteAddr memory) mbCfgAddr
+  initArgs_ <- minimalArgShapes bak archCtx mdEntryAbsAddr
+  let shouldUseDwarf = simEnableDebugInfoPreconditions simOpts
+  let getDwarfArgs = do
```
RyanGlScott, line 555: It might be worth adding a type signature here to make it more obvious that the `do`-notation is for the `Maybe` monad.

langston-barrett: Adding a full type signature would have required using `ScopedTypeVariables`, so I added a comment next to the `do` instead.

PR #293, opened 2025-07-25
cli: Add flags for specifying initial register values

### LINKED ISSUE #228: cli: Add flags for specifying initial register values

langston-barrett:
Fixes #228. Based on #292 and #291. Needs:

- [ ] Some subset of:
  - [ ] `--arg-buf-init`
  - [ ] `--arg-int-32`
  - [ ] `--arg-int-64`
  - [ ] `--arg-str`
  - [ ] `--arg-buf-bytes`
- [ ] A test for Macaw S-expression CFGs

### DIFF
--- a/grease-cli/src/Grease/Main.hs
+++ b/grease-cli/src/Grease/Main.hs
@@ -516,6 +516,56 @@ initialLlvmFileSystem halloc sym simOpts = do
   (fs, gs, ov) <- CLLVM.SymIO.initialLLVMFileSystem halloc sym ?ptrWidth fileContents mirroredOutputs C.emptyGlobals
   pure (fileContents, fs, gs, ov)
 
+-- | Compute the initial 'ArgShapes' for a Macaw CFG.
+--
+-- Sources argument shapes from:
+--
+-- 1. A default initial shape for each register (via 'minimalArgShapes')
+-- 2. DWARF debug info (via 'loadDwarfPreconditions') if
+--    'simEnableDebugInfoPreconditions' is 'True'
+-- 3. A shape DSL file (via 'loadInitialPreconditions')
+--
+-- Later steps override earlier ones.
+macawInitArgShapes ::
+  ( C.IsSymBackend sym bak
+  , C.IsSyntaxExtension (Symbolic.MacawExt arch)
+  , Symbolic.SymArchConstraints arch
+  , Mem.HasPtrWidth (MC.ArchAddrWidth arch)
+  , MM.MemWidth (MC.ArchAddrWidth arch)
+  , Integral (Elf.ElfWordType (MC.ArchAddrWidth arch))
+  , Show (ArchReloc arch)
+  , ?memOpts :: Mem.MemOptions
+  , ?parserHooks :: CSyn.ParserHooks (Symbolic.MacawExt arch)
+  ) =>
+  GreaseLogAction ->
+  bak ->
+  ArchContext arch ->
+  SimOpts ->
+  MacawCfgConfig arch ->
+  Ctx.Assignment (Const String) (Symbolic.CtxToCrucibleType (Symbolic.ArchRegContext arch)) ->
+  -- | If simulating a binary, this is 'Just' the address of the user-requested
+  -- entrypoint function. Otherwise, this is 'Nothing'.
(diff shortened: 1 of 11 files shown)


grease-cli/src/Grease/Main.hs
```diff
@@ -1157,28 +1179,32 @@ simulateMacaw la halloc elf loadedProg mbPltStubInfo archCtx txtBounds simOpts p
 --
 -- 1. A default initial shape for each register (via 'minimalShapeWithPtrs')
 -- 2. DWARF debug info (via 'GLD.diArgShapes') if
---    'simEnableDebugInfoPreconditions' is 'True'
+--    'initPrecondUseDebugInfo' is 'True'
 -- 3. A shape DSL file (via 'loadInitialPreconditions')
+-- 4. Simple shapes from the CLI (via 'useSimpleShapes')
 --
 -- Later steps override earlier ones.
```
2over12, line 1186: Do we document the ordering between initial shapes and args somewhere?

2over12 (approved):
This lgtm and a big user improvement for simple cases

grease-cli/src/Grease/Main.hs
langston-barrett: Can you say more about your question? In `doc/refinement.md`, we make the same statement that is made here.

2over12: e.g. for dwarf we specify https://github.com/GaloisInc/grease/blob/f9beb3c6da86b73f19572d010b91b98802ad08da/grease-cli/src/Grease/Cli.hs#L122, hopefully a user doesnt usually do stuff like this but might be nice to mention in the help menu that simple shapes will override all

langston-barrett: Yeah, so that is covered here in the docs: https://github.com/GaloisInc/grease/pull/293/files#diff-98ed49ecb35aed8906c90016ab14e6bc0a78b5a3f1380206dac478144d1a4224R66

I will add some content to `--help` to point to users to the docs to understand the precedence order here.

langston-barrett: Clarified, here's what it looks like now:
```
Initial precondition options
  --use-debug-info-types   Use types in debug info to infer initial
                           preconditions. Can be (partially) overridden
                           --initial-precondition, --arg-*, and/or
                           --*-startup-override.
  --type-unrolling-bound ARG
                           Number of recursive pointers to visit during DWARF
                           shape building (default: 3)
  --initial-precondition FILE
                           Initial precondition for use in refinement. Can be
                           (partially) overridden by --arg-* and/or
                           --*-startup-override.
  --arg-buf-uninit ARG:N   initialize argument ARG to a pointer to an
                           uninitialized buffer of N bytes
  --arg-buf-init ARG:N     initialize argument ARG to a pointer to an
                           initialized buffer of N symbolic bytes
  --arg-u32 ARG:N          initialize argument ARG to an unsigned 32-bit integer
  --arg-u64 ARG:N          initialize argument ARG to an unsigned 64-bit integer
```

Issue #304, opened 2025-07-31
doc: Revise Ghidra batch plugin docs


langston-barrett:
The [page on the Ghidra batch plugin](https://galoisinc.github.io/grease/ghidra-batch-plugin.html) makes references like "this plugin" and describes commands that should be run in the `./ghidra-batch-plugin` directory, without noting such. This is because it was previously the README in that directory, but was migrated to `doc/` without wording changes. We should revise this so that it fits in more naturally with the rest of the documentation.

PR #305, opened 2025-07-31
Specify where to run commands for ghidra-plugin install

2over12:


langston-barrett (approved):
This is a good change, though we'll need a bit more work to fully solve #304.

If you wouldn't mind, I'd appreciate you giving this doc a close read and thinking about how to fit it in with the rest of the documentation. For instance:

1. "GREASE" can be removed from the title, because it appears at the top of each documentation page anyway
2. We tend to use sentence-case for titles in the docs, e.g., "Comparison with other tools..." as opposed to "Comparison With Other Tools..."
3. The link to the installation instructions should be a intra-doc link (specified by a path) rather than a URL
4. There's no need to duplicate the build instructions now that this is part of the docs that contain those
5. The section on formatting should be part of the dev docs, because it isn't user-facing

I'd be fine with merging this as-is if you'd prefer to do that in a follow-up or just defer for future work.

PR #317, opened 2025-08-11
macaw: Address overrides

### LINKED ISSUE #321: macaw: Overrides that run at arbitrary addresses
From #313:

> It would be nice for GREASE to support custom assertions that denote a bug. Ideally, these would be highly expressive, and able to be attached to (a set of) addresses.  \[In particular, these addresses might not be function entrypoints.\] When GREASE reaches that address, it would evaluate the custom assertion expression, and determine whether the computed predicate is satisfiable. If so, GREASE would consider it a possible bug.

One option would be to reuse the existing S-expression syntax. The user could provide an S-expression file that takes the current register state as a parameter and possibly calls the builtins described in #320. The user could provide the address and the path to the S-expression file via a command-line argument.

More generally, it would be nice for overrides that run at a particular address ("address overrides") to be able to arbitrarily inspect and modify the current state (e.g., registers, memory).

Implementation-wise, this could probably happen by overriding the behavior of the [`MacawInstructionStart`](https://github.com/GaloisInc/grease/blob/6149ec2894c52d881c92cc44d14573fc534bf68b/grease/src/Grease/Macaw/SimulatorHooks.hs#L154) statement.

One complication is that the Crucible simulator can't just be coaxed into making a call to an arbitrary CFG regardless of its state. We may need to execute these address overrides in a complete new simulator, inheriting the globals (i.e., LLVM memory) and register state from the parent, and mapping back if possible. It is likely possible to re-import changed globals, but I'm not at all sure that it would be possible to change register values.

langston-barrett:
Fixes #321 .

### DIFF
--- a/doc/overrides.md
+++ b/doc/overrides.md
@@ -286,10 +286,10 @@ Note that:
 
 ## Address overrides
 
-Users may supply *address overrides* via `--address-override`, which are
-overrides that run when symbolic execution reaches a particular address (i.e.,
-instruction pointer value). Address overrides are only available when executing
-binaries.
+Users may supply *address overrides* via `--addr-override`. Address overrides
+are overrides that run when symbolic execution reaches a particular address
+(i.e., instruction pointer value). Address overrides are only available when
+executing binaries.
 
 An address override takes the register struct (e.g., `X86Regs`) as an input
 and returns `Unit`. When control reaches the address in question, the override


langston-barrett:
Example:
```
(defun @print ((regs X86Regs)) Unit
  (start start:
    (println "HERE!")
    (return ())))
```
```
cabal run exe:grease -- --addr-override 0x401020:print.x64.cbl grease-cli/tests/refine/pos/malloc-free/test.x64.elf --symbol test

HERE!
All goals passed!
Not testing any requirements beyond memory safety
Finished analyzing 'test'.
```

grease/src/Grease/Macaw/Overrides/Address.hs
```diff
@@ -0,0 +1,198 @@
...
+-- Instead, they construct a new backend and simulator, and run there. This is
+-- because Crucible can't just start running a CFG from an arbitrary state.
+runAddressOverride ::
+  ( sym ~ W4.ExprBuilder t st fs
+  , Symbolic.SymArchConstraints arch
+  ) =>
+  C.CrucibleState p sym (Symbolic.MacawExt arch) rtp blocks r (Ctx.EmptyCtx Ctx.::> Symbolic.ArchRegStruct arch) ->
+  C.SomeCFG (Symbolic.MacawExt arch) (Ctx.EmptyCtx Ctx.::> Symbolic.ArchRegStruct arch) C.UnitType ->
+  IO ()
+runAddressOverride crucState someCfg = do
```
2over12, line 141: So if im reading this right we construct a new simstate and just pass the regs in as an arg, I guess this means we can write assertions over reg state but not memory state unfortunately, probably need an issue for that at some point since that may become important?

langston-barrett: We should also be able to write assertions over memory state. The LLVM memory model is a Crucible global variable, and the globals are copied as-is into the new simulator state in `toInitialState`: https://github.com/GaloisInc/grease/pull/317/files#diff-af0cd58b814678c7303457fc970cc51a0ba492caa89d1a71a03459600e38498dR126

PR #318, opened 2025-08-11
Print post-simulation memory

### LINKED ISSUE #315: Print the final LLVM memory
We currently print LLVM memory in the debug log as it exists *before* execution, but not *after*. It would be nice to have the ability to view the post-state.

langston-barrett:
Fixes #315.

See it in action with:
```
cabal run exe:grease -- grease-cli/tests/llvm/free.llvm.cbl -v

-- snip --
Base memory
  Allocations:
    Free annotation Nonce {indexValue = 213} 183 grease-cli/tests/llvm/free.llvm.cbl:10:5
-- snip --
```

### DIFF
--- a/grease-cli/src/Grease/Main.hs
+++ b/grease-cli/src/Grease/Main.hs
@@ -760,6 +760,7 @@ macawInitState ::
   SimOpts ->
   Macaw.SetupHook sym arch ->
   Symbolic.MemPtrTable sym (MC.ArchAddrWidth arch) ->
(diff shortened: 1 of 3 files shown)


RyanGlScott (approved):
Given that we already call `ppMem` in another diagnostic message (in `Grease.Setup.Diagnostic`), do you think it is worth prefacing each message with something like "Pre-simulation memory" and "Post-simulation memory"? Or is it pretty obvious which is which if you see the surrounding context?

Issue #327, opened 2025-08-12
Can't `close` stdin


langston-barrett:
This fails:
```
; Copyright (c) Galois, Inc. 2025

; Ensure that `close`ing `stdin` works.

;; flags {"--symbol", "test"}
;; go(prog)

; int close(int filedes)
(declare @close ((filedes (Ptr 32))) (Ptr 32))

(defun @test () Unit
  (start start:
    (let stdin (ptr 32 0 (bv 32 0)))
    (let closed (funcall @close stdin))
    (let closed-off (ptr-offset 32 closed))
    (assert! (not (equal? closed-off (bv 32 0xffffffff))) "close() returned -1!")
    (return ())))

;; ok()
```
```
cabal run exe:grease -- tests/llvm/close-stdin.llvm.cbl

Finished analyzing 'test'. Likely bug: unavoidable error (safety condition is unsatisfiable) at tests/llvm/close-stdin.llvm.cbl:16:5
tests/llvm/close-stdin.llvm.cbl:16:5: error: in test
close() returned -1!
```
But it works on my machine ™️:
```c
#include <unistd.h>
int main() { close(0); }
```
```sh
clang -fsanitize=undefined test.c -o test
./test
echo $?
0
```

langston-barrett:
The upstream crux-llvm test succeeds:

https://github.com/GaloisInc/crucible/blob/331c83b37bc3e375d8cdc64f285417eb38dcd695/crux-llvm/test-data/golden/stdio/close-read-stdin.c#L9-L10

This may be because they explicitly initialize stdin here:

https://github.com/GaloisInc/crucible/blob/331c83b37bc3e375d8cdc64f285417eb38dcd695/crux-llvm/src/Crux/LLVM/Simulate.hs#L217

PR #329, opened 2025-08-12
cli: Add `--sym-stdin`, always initialize `stdin`

### LINKED ISSUE #324: cli: Flag to specify a number of symbolic bytes for `stdin`
We should add a flag `--sym-stdin N` that initializes the `stdin` of the symbolic filesystem with `N` symbolic bytes.

langston-barrett:
Fixes #324, #327.


langston-barrett:
Agreed! Mind if I do that in a follow-up? #330

PR #333, opened 2025-08-13
ci: Support GHC 9.10.2, drop GHC 9.4.8

### LINKED ISSUE #44: Support GHC 9.10
@langston-barrett says:

-----

Prerequisite: https://github.com/GaloisInc/stubs/issues/51

langston-barrett:
Fixes #44.


langston-barrett:
Ah, alas:
```
src/Grease/Heuristic/Diagnostic.hs:4:14: error: [GHC-46537]
Error:     Unsupported extension: TypeAbstractions
  |
4 | {-# LANGUAGE TypeAbstractions #-}
  |              ^^^^^^^^^^^^^^^^
```
https://downloads.haskell.org/ghc/9.10.1-alpha2/docs/users_guide/using-warnings.html?highlight=warnings#ghc-flag--Wdeprecated-type-abstractions

This warning was introduced in GHC 9.10 and I wanted to turn it into an error, but it appears that actually fixing the error requires GHC 9.8. I'll just put it in the block of `-Werror` warnings for GHC 9.12, then.

RyanGlScott:
Alternatively, I think you could just guard the use of `{-# LANGUAGE TypeAbstractions #-}` behind the appropriate CPP.

langston-barrett:
Eh, I don't think it's worth the effort / complication. The point of this warning is just to make us aware that we need to make a change in the future, so I think we have successfully handled it (in some sense) when we have a concrete plan for resolving the situation it warns about. In this case, that's fixing it when we do the GHC 9.12 upgrade.

RyanGlScott (approved):

PR #336, opened 2025-08-13
Path splitting, `getopt_long`, and unsupported relocations

### LINKED ISSUE #323: Path splitting
GREASE should support a mode that performs [path splitting](https://hackage-content.haskell.org/package/crucible-0.7.2/docs/Lang-Crucible-Simulator-PathSplitting.html). In this mode, GREASE would always branch concretely, and never mux states at CFG merge points. Upon completion of a branch (or in the future, perhaps even just upon encountering a safety assertion) GREASE would perform abduction. If a new precondition is found, the remaining states would be discarded and execution would start again under that precondition.

There's some resaon to suspect that this might be better than GREASE's default (SSE) mode. First of all, the safety assertions will be simpler, making them easier for the heuristics to inspect and for SMT solvers to dispatch. Secondly, if GREASE encounters an error it doesn't know how to solve, that might become apparent sooner rather than later. If this experiment turns out well, we might consider splitting paths by default.

In the far future, we should consider a more intelligent strategy that alternates between path splitting and state merging depending on its own heuristics, like ["veritesting"](https://dl.acm.org/doi/10.1145/2568225.2568293).

langston-barrett:
Fixes #323. Needs:

- [x] Log messages for completion of each path
- [x] Test case(s)

### DIFF
--- a/doc/shared-libraries.md
+++ b/doc/shared-libraries.md
@@ -49,4 +49,10 @@ name of the PLT stub, it is fine to leave `<NAME>` empty.) Reverse engineering
 suites such as [Ghidra](https://ghidra-sre.org/) can be effective at
 determining PLT stub information.
 
+## Skipping unsupported relocations
+
+The flag `--skip-unsupported-relocs` induces GREASE to populate unsupported
+relocations with fresh, symbolic data rather than throwing an exception and
+halting analysis.
+
 <!-- Copyright (c) Galois, Inc. 2024. -->


langston-barrett:
This is failing on a simple test case with
```
cabal run exe:grease -- --symbol test grease-cli/tests/llvm/uninit-stack-conditional.llvm.cbl --path-strategy Dfs


Path ended at grease-cli/tests/llvm/uninit-stack-conditional.llvm.cbl:12:5 with result success
grease: Global variable for BoundedExecFrameData not initialized
CallStack (from HasCallStack):
  error, called at src/Lang/Crucible/Simulator/BoundedExec.hs:186:25 in crucible-0.7.2.0.99-inplace:Lang.Crucible.Simulator.BoundedExec
```

langston-barrett:
Huh, apparently I wrote in override in such a way that it typechecks with GHC 9.10, but not with any earlier GHCs!

doc/usage.md
```diff
@@ -71,4 +71,14 @@ by loading a file at a fixed offset. The `--raw-binary` flag enables this mode.
...
 
+## Path strategies
+
+GREASE supports a few different *path strategies*, which indicate how to handle
+branching. A strategy can be specified using `--path-strategy`.
+
+- SSE stands for *static symbolic execution*. In this mode, GREASE always
+  merges paths at control-flow join points.
+- DFS stands for *depth-first search*. In this mode, GREASE never merges paths,
+  and explores paths in a depth-first traversal.
```
RyanGlScott: I take that "SSE" and "DFS" are meant to be the literal strings that you pass to `--path-strategy`? If so, it might be worth rendering these words in monospace font to make the connection to `--path-strategy` (also rendered in monospace font) a bit clearer.

doc/usage.md
```diff
@@ -71,4 +71,14 @@ by loading a file at a fixed offset. The `--raw-binary` flag enables this mode.
 is loaded at offset `0x0`, which is likely not ideal for many binaries so an offset can be set with `--load-base BASE_ADDR`.
 There are no symbols in a raw binary, so address entrypoints are the only valid entrypoints (i.e. `--address`).
 
+## Path strategies
+
+GREASE supports a few different *path strategies*, which indicate how to handle
+branching. A strategy can be specified using `--path-strategy`.
+
+- SSE stands for *static symbolic execution*. In this mode, GREASE always
+  merges paths at control-flow join points.
```
RyanGlScott: Mention that SSE is the default strategy.

grease-cli/tests/llvm/uninit-stack-conditional.llvm.cbl
```diff
@@ -0,0 +1,24 @@
...
+    (let a (alloca none sz))
+    (branch b if: else:))
+  (defblock if:
+    (return ()))
+  (defblock else:
+    (load none i8 a)
+    (return ())))
+
+;; flags {"--symbol", "test"}
+;; flags {"--path-strategy", "Sse"}
```
RyanGlScott: Is `--path-strategy` case-insensitive, or do you literally have to type it as "`Sse`"? If it's the latter, then I would propose that we do make it case-insensitive, especially since the docs don't mention this as a requirement.

grease-cli/src/Grease/Cli.hs
```diff
@@ -332,6 +340,12 @@ simOpts = do
                     ]
                 )
           )
+  simSkipUnsupportedRelocs <-
+    fmap GO.SkipUnsupportedRelocs $
+      Opt.switch
+        ( Opt.long "skip-unsupported-relocs"
+            <> Opt.help "Populate unsupported relocations with symbolic data"
```
RyanGlScott, line 347: Do you have a test case that makes key use of this feature that we could add?

grease/src/Grease/Macaw.hs
```diff
@@ -373,7 +375,8 @@ memConfigInitial bak arch ptrTable relocs =
       -- way to hook each read without needing to re-implement all of the logic
       -- for MacawReadMem/MacawCondReadMem, which is quite involved.
       Symbolic.concreteImmutableGlobalRead = \memRep ptr -> do
-        assertRelocSupported arch ptr relocs
+        Monad.unless (Opts.getSkipUnsupportedRelocs skipUnsupportedRelocs) $
+          assertRelocSupported arch ptr relocs
```
RyanGlScott, line 380: It might be worth updating the comment [here](https://github.com/GaloisInc/grease/blob/b70bd70b5849e03d785f844be8fa1fa6a8784f62/grease/src/Grease/Macaw.hs#L131-L138) to mention `--skip-unsupported-relocs`.

grease/src/Grease/Options.hs
```diff
@@ -35,6 +37,11 @@ import Grease.Shape.Simple (SimpleShape)
 import Grease.Solver (Solver (..))
 import Lang.Crucible.Utils.Timeout (Timeout)
 
+data PathStrategy
+  = Dfs
+  | Sse
```
RyanGlScott: Please spell out what these acronyms stand for in Haddock comments.

grease/src/Grease/Refine.hs
```diff
@@ -346,24 +352,35 @@ execCfg ::
...
-  X.finally
-    ( do
-        r <- C.executeCrucible execFeats' initialState
-        o <- C.getProofObligations bak
-        pure (r, o)
+  IO
+    ( C.ExecResult p sym ext (C.RegEntry sym ret)
+    , C.ProofObligations sym
+    , Seq (C.WorkItem p sym ext (C.RegEntry sym ret))
     )
```
RyanGlScott, line 365: This return type is getting pretty large—worth leaving a Haddock comment to explain what each element of the tuple represents?

grease/src/Grease/Refine.hs
```diff
@@ -432,16 +449,76 @@ execAndRefine ::
...
+  let refineOne initSt = do
+        (execResult, goals, remaining) <- execCfg bak strat execFeats' initSt
+        doLog la (Diag.ExecutionResult memVar execResult)
+        bbMap <- readIORef bbMapRef
+        refineResult <- proveAndRefine bak solver solverTimeout anns execResult la heuristics argNames argShapes initState bbMap goals
+        case strat of
+          Opts.Dfs -> do
+            loc <- W4.getCurrentProgramLoc (C.backendGetSym bak)
+            doLog la (Diag.RefinementFinishedPath loc (shortResult refineResult))
+          _ -> pure ()
```
RyanGlScott: Nit pick: spell out this pattern so that we `-Wincomplete-patterns` will warn us in case we ever add more path strategies:

```suggestion
          Opts.Sse -> pure ()
```

grease-cli/tests/llvm/uninit-stack-conditional.llvm.cbl
langston-barrett: Made lower-case and clarified in the docs.

grease-cli/src/Grease/Cli.hs
langston-barrett: Not a standalone one, no. But execution of `bc` from `main` quickly grinds to a halt without this:
```
cabal run exe:grease -- bc --symbol main --path-strategy dfs --no-heuristics --sym-stdin 8

Analyzing from entrypoint 'main'
Skipped call to a PLT stub at address 0x10002360: isatty
grease: Attempted to read from an unsupported relocation type (R_X86_64_COPY) at address 0x100172e0.
This may be due to a PLT stub that grease did not detect.
If so, try passing --plt-stub <ADDR>:<NAME>, where
<ADDR> and <NAME> can be obtained by disassembling the
relevant PLT section of the binary
(.plt, .plt.got, .plt.sec, etc.).

HasCallStack backtrace:
  finally, called at src/Lang/Crucible/Backend/Online.hs:228:5 in crucible-0.7.2.0.99-inplace:Lang.Crucible.Backend.Online
```

RyanGlScott: Ah, so this is really about `COPY` relocations in particular. That makes sense. It's rather difficult to handle those correctly in `grease`, given that it doesn't support shared libraries (where the data that otherwise would be copied over lives).

Perhaps this trick of using `--skip-unsupported-relocs` to skip over `COPY` relocations should be mentioned in the docs?

RyanGlScott (approved):
LGTM modulo https://github.com/GaloisInc/grease/pull/336#discussion_r2279034710

PR #338, opened 2025-08-14
UX improvements to Ghidra plugin

2over12:
- [ ] Show mustfail with multiple locations at entrypoint
- [ ] Option for overrides file

### DIFF
--- a/ghidra-plugin/src/main/scala/grease/GreaseWrapper.scala
+++ b/ghidra-plugin/src/main/scala/grease/GreaseWrapper.scala
@@ -135,7 +135,11 @@ case class GreaseConfiguration(
 case class GreaseException(val msg: String) extends Exception
 
 object GreaseResult {
-  def parseBatch(batch: String, addrs: AddressingMode): Option[PossibleBug] = {
+  def parseBatch(
+      ent: Address,
+      batch: String,
+      addrs: AddressingMode
+  ): Option[PossibleBug] = {
     val js = ujson.read(batch)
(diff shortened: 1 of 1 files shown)


ghidra-plugin/src/main/scala/grease/GreaseAnalyzer.scala
```diff
@@ -214,6 +187,14 @@ class GreaseAnalyzer
       null,
       "Uses the image base of this Ghidra binary as the load base when in raw mode (this is the default, disable to use custom image base)"
     )
+
+    options.registerOption(
+      GreaseAnalyzer.OVERRIDE_FILE,
+      OptionType.FILE_TYPE,
+      null,
+      null,
+      "Override file to use in GREASE"
```
langston-barrett: Maybe worth mentioning that this is in YAML format? And/or mentioning that it will be passed to `--overrides`?

ghidra-plugin/src/main/scala/grease/GreaseWrapper.scala
```diff
@@ -177,10 +232,83 @@ case class PossibleBug(
...
+      ParsedBatch.addComment(
+        s"Concretized invalidating shape ${pred.concShapes}",
+        prog,
+        pred.loc
+      )
+
+    ParsedBatch.greaseBookmark(
+      prog,
+      inFunc,
+      "Failed to Refine",
```
langston-barrett:
```suggestion
      "Failed to refine",
```

Issue #343, opened 2025-08-15
Add a test-case for the "heartbleed" vulnerability


langston-barrett:
With a little prompting, GREASE can find the "Heartbleed" vulnerability, as distilled into this file:

https://github.com/GaloisInc/BESSPIN-Tool-Suite/blob/c0843e65eabb6c1bdd9238a49ecb69c404d52fa8/docs/cwesEvaluation/poc-exploits/Heartbleed/heartbleed.c

<details>
<summary>Contents</summary>
    
```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>

// Macros defined in OpenSSL library file 'ssl/ssl_locl.h'.
#define n2s(c,s) ((s=(((unsigned int)(c[0]))<< 8)| \
                     (((unsigned int)(c[1]))    )),c+=2)
#define s2n(s,c) ((c[0]=(unsigned char)(((s)>> 8)&0xff), \
                   c[1]=(unsigned char)(((s)    )&0xff)),c+=2)

// Macros defined in OpenSSL library file 'ssl/ssl3.h'.
#define TLS1_HB_REQUEST  1
#define TLS1_HB_RESPONSE 2

#define POC_HEARTBLEED_REAL_PAYLOAD_LEN 2
#define POC_HEARTBLEED_NOMINAL_PAYLOAD_LEN 20

// Fill in the 'buf' array with 'num' random numbers.
void RAND_pseudo_bytes(unsigned char *buf, size_t num) {
  for (size_t i = 0; i < num; ++i) {
    buf[i] = rand();
  }
}


// Original function implemented in OpenSSL library file 'ssl/t1_lib.c'.
int tls1_process_heartbeat(unsigned char *message) {
  unsigned char *p = &message[0], *pl;
  unsigned short hbtype;
  unsigned int payload;
  unsigned int padding = 16; /* Use minimum padding */

  /* Read type and payload length first */
  hbtype = *p++;
  n2s(p, payload);
  pl = p;

  if (hbtype == TLS1_HB_REQUEST) {
    unsigned char *buffer, *bp;

    /* Allocate memory for the response, size is 1 bytes
     * message type, plus 2 bytes payload length, plus
     * payload, plus padding
     */
    buffer = (unsigned char *) malloc (1 + 2 + payload + padding);
    if (buffer == NULL) {
      printf("Failed to allocate space for the response buffer\n");
      return -1;
    }
    bp = buffer;

    /* Enter response type, length and copy payload */
[...]

PR #345, opened 2025-08-15
Perform path-satisfiability checking

### LINKED ISSUE #344: Path satisfiability checking
We should add support for [path satisfiability checking](https://hackage-content.haskell.org/package/crucible-0.7.2/docs/Lang-Crucible-Simulator-PathSatisfiability.html#v:pathSatisfiabilityFeature). It should probably be on by default.

langston-barrett:
Fixes #344, #343.


langston-barrett:
This improves GREASE's output on the Heartbleed example (#343), since it no longer reports a "possible bug" that's further down an infeasible path. I also had to enable must-fail even in the presence of `--no-heuristics`. Otherwise the "multi-must-fail" heuristic fires, but just on the single predicate (which would otherwise be must-fail). This is unnecessarily confusing. I think must-fail despite `--no-heuristics` is fine, because the real aim of `--no-heuristics` is just to not refine the precondition (IMO). All together, this gives us a good result on the Heartbleed example (#343):
```
cabal run exe:grease -- --symbol tls1_process_heartbeat heartbleed --initial-precondition shapes.txt --no-heuristics

Analyzing from entrypoint 'tls1_process_heartbeat'
Finished analyzing 'tls1_process_heartbeat'. Likely bug: unavoidable error (safety condition is unsatisfiable) at 0x100012db
The region wasn't allocated or wasn't large enough
Performing a memory copy of 0xffff:[64] bytes
  Destination: (6, 0x3:[64])
  Source:      (annotation Nonce {indexValue = 194} 3, 0x3:[64])
```

langston-barrett:
Nice, with these changes GREASE doesn't even need a full shapes DSL precondition, it can find the bug just with `rdi` pointing to 19 initialized bytes:
```
cabal run exe:grease -- --symbol tls1_process_heartbeat heartbleed --no-heuristics --arg-buf-init rdi:19

Finished analyzing 'tls1_process_heartbeat'. Possible bug(s):

At 0x100012db:
Mem copy failed
Details:
  Invalid copy source
Concretized arguments:

rcx: 0000000000000000
rdx: 0000000000000000
rsi: 0000000000000000
rdi: 000000+0000000000000000
r8: 0000000000000000
r9: 0000000000000000
r10: 0000000000000000

000000: 01 80 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
```
That is a valid Heartbleed payload.

langston-barrett:
Enabling `assertThenAssume` with path-sat turned out to require a new heuristic relating to `free` for all the tests to pass 🤷

PR #349, opened 2025-09-04
doc: Describe how to report test suite coverage

langston-barrett:
It's surprisingly difficult!

I'm pleased and surprised to report that our coverage numbers are actually not that bad:

- 46% of "top level definitions"
- 55% of "alternatives"
- 73% of "expressions"

### DIFF
--- a/doc/dev/tests.md
+++ b/doc/dev/tests.md
@@ -124,3 +124,55 @@ structure to that of rustc, so almost all of the advice there applies *mutatis
 mutandis*.
 
 [helpful guidance]: (https://github.com/rust-lang/rust/blob/3350c1eb3fd8fe1bee1ed4c76944d707bd256876/src/doc/rustc-dev-guide/src/tests/best-practices.md)
+
+## Coverage
+
+To enable collection of coverage information, add the following to
+`cabal.project.local`. This will collect coverage information for the `grease`
+and `grease-cli` packages, but *not* for any of their dependencies.
+
+```cabal
+coverage: False
+package *
+  coverage: False
+package grease
+  coverage: True
+package grease-cli
+  coverage: True
+```
+
+Unfortunately, you'll have to force Cabal to recompile GREASE and all of its
+dependencies for these changes to take effect.
+
+```sh
+cabal clean
+cabal build
+```
+
+Now, run the test suite:
+
+```sh
+cd grease-cli
+cabal run test:grease-tests
+```
+
+You should see `grease-tests.tix` in your working directory. If not, you may
+have forgotten to run `cabal clean`.
+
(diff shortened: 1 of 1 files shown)


doc/dev/tests.md
```diff
@@ -124,3 +124,54 @@ structure to that of rustc, so almost all of the advice there applies *mutatis
...
+
+```cabal
+coverage: False
+package *
+  coverage: False
+package grease
+  coverage: True
+package grease-cli
+  coverage: True
+```
```
RyanGlScott, line 142: Worth a comment saying what this project configuration is doing—namely, it's measuring coverage in `grease` itself, but _not_ in any of its dependencies.

doc/dev/tests.md
```diff
@@ -124,3 +124,54 @@ structure to that of rustc, so almost all of the advice there applies *mutatis
...
+export GHCVER=$(ghc --numeric-version)
+hpc markup \
+  grease-cli/grease-tests.tix \
+  --hpcdir=${PWD}/dist-newstyle/build/aarch64-osx/ghc-${GHCVER}/grease-0.1.0.0/build/extra-compilation-artifacts/hpc/dyn/mix \
+  --hpcdir=${PWD}/dist-newstyle/build/aarch64-osx/ghc-${GHCVER}/grease-cli-0.1.0.0/build/extra-compilation-artifacts/hpc/dyn/mix \
+  --hpcdir=${PWD}/dist-newstyle/build/aarch64-osx/ghc-${GHCVER}/grease-cli-0.1.0.0/x/grease/build/grease/grease-tmp/extra-compilation-artifacts/hpc/vanilla/mix \
+  --hpcdir=${PWD}/dist-newstyle/build/aarch64-osx/ghc-${GHCVER}/grease-cli-0.1.0.0/t/grease-tests/build/grease-tests/grease-tests-tmp/extra-compilation-artifacts/hpc/vanilla/mix \
+  --srcdir=grease \
+  --srcdir=grease-cli
+```
```
RyanGlScott, line 178: This is just complicated enough that I wonder if we shouldn't package it up into a script of some sort. This would have the advantage that it would be easier to users to run the command locally, and we could potentially make it more robust across different GHC versions by scripting the logic used to find the `mix` directories. (No need to actually write this additional scripting as a part of this PR, but it might be nice to do something akin to SAW's [`compute-coverage.sh`](https://github.com/GaloisInc/saw-script/blob/17e5d4d74b60cb27bbcf95c00f1ac62b37773232/compute-coverage.sh) at a later date.)

langston-barrett: Agreed! https://github.com/GaloisInc/grease/issues/350

PR #355, opened 2025-09-05
ci: Output HTML coverage report to job summary page

langston-barrett:
Adds a coverage overview to the "job summary" page, see [this run](https://github.com/GaloisInc/grease/actions/runs/17498093231). This overview is not *optimal*, because I'm just `cat`ing an entire HTML document (with a header, styling, etc.) into a file that gets displayed as GitHub-flavored Markdown. This results in (1) some garbage at the top, (2) blank "progress" bars, and (3) dead links. I personally think all this is fine, although we could consider doing some light Python scripting to extract just the table. We could do that up-front or an a follow-up.

It would also be nice to package up the whole report as an artifact that developers can download, I'll do that next.

See:

- https://github.blog/news-insights/product-news/supercharging-github-actions-with-job-summaries/
- https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#adding-a-job-summary

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -112,7 +112,10 @@ jobs:
 
     - name: Report coverage
       shell: bash
-      run: ./scripts/coverage.sh
+      run: |
+        ./scripts/coverage.sh markup
+        cat hpc_index.html >> "${GITHUB_STEP_SUMMARY}"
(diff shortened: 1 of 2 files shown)


langston-barrett:
It was real easy to implement the table extraction, so I went ahead and did that part.

PR #358, opened 2025-09-09
Resolve forward declarations in address overrides

### LINKED ISSUE #357: Resolve forward declarations in address overrides
I neglected to implement the logic to resolve forward declarations that appear in [address overrides](https://github.com/GaloisInc/grease/blob/da4089043f2580b5b3d22ee3db0c5ba5695c8d39/doc/overrides.md#address-overrides), meaning that such overrides are incapable of using built-ins such as [`fresh-bytes` or `read-c-string`](https://github.com/GaloisInc/grease/blob/da4089043f2580b5b3d22ee3db0c5ba5695c8d39/doc/builtins.md#s-expression-specific-overrides) (which must be explicitly `@declare`d).

For example, if I replace the content of [`addr-override.x64.cbl`]() with
```
(declare @read-c-string ((x (Ptr 64))) (String Unicode))
(defun @nope ((regs X86Regs)) Unit
  (start start:
    (let rax (get-reg rax regs))
    (funcall @read-c-string rax)
    (return ())))
```
Then
```
cabal run exe:grease -- --symbol test --addr-override 0x401000:tests/refine/bug/extra/addr-override.x64.cbl tests/refine/pos/compare_to_null/test.x64.elf
```
yields
```
grease: Could not resolve function: read-c-string
Called at: tests/refine/bug/extra/addr-override.x64.cbl:5:5
In nope at tests/refine/bug/extra/addr-override.x64.cbl:5:5
```

langston-barrett:
Fixes #357.

### DIFF
--- a/grease/src/Grease/Macaw/Overrides/Address.hs
+++ b/grease/src/Grease/Macaw/Overrides/Address.hs
@@ -12,6 +12,7 @@ module Grease.Macaw.Overrides.Address (
   addressOverrideParser,
   AddressOverrides,
   loadAddressOverrides,
+  registerAddressOverrideHandles,
   maybeRunAddressOverride,
 ) where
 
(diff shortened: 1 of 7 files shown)


grease/src/Grease/Macaw/Overrides/Address.hs
```diff
@@ -67,8 +78,86 @@ addressOverrideParser = do
...
+    [] ->
+      userErr ("Expected to find a function named `" <> fnNameText <> "`")
+    _ : _ ->
+      userErr ("Override contains multiple `" <> fnNameText <> "` functions")
+
+-- | Does a function have the same name as the @.cbl@ file in which it is
+-- defined? That is, is a function publicly visible from the @.cbl@ file?
+isPublicCblFun :: W4.FunctionName -> LCCR.AnyCFG ext -> Bool
+isPublicCblFun fnName (LCCR.AnyCFG cfg) =
+  C.handleName (LCCR.cfgHandle cfg) == fnName
```
RyanGlScott, line 160: This is exactly the same as the function of the same name in `Grease.LLVM.Overrides.Skip`, correct? Can the two functions be de-duplicated somehow?

langston-barrett: Yes, and we should go further and deduplicate some other related logic that isn't yet in its own function(s): #362.

PR #360, opened 2025-09-10
Expose more Macaw CFG type-checking functionality

langston-barrett:
Follow-up to #359, expose lower-level functionality with a structured error type to make it more usable by downstream clients.

### DIFF
--- a/grease/src/Grease/Macaw/Entrypoint.hs
+++ b/grease/src/Grease/Macaw/Entrypoint.hs
@@ -47,7 +47,8 @@ checkMacawEntrypointCfgsSignatures archCtx entrypointCfgs = do
     pure entrypointCfgs
   let entryName = W4.functionName (C.handleName (C.Reg.cfgHandle entrypointCfg0))
   let name = Text.concat ["CFG `", entryName, "`"]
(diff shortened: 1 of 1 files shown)


grease/src/Grease/Macaw/Entrypoint.hs
```diff
@@ -68,44 +78,82 @@ checkMacawEntrypointCfgsSignatures archCtx entrypointCfgs = do
...
+        (Symbolic.MacawExt arch)
+        (Ctx.EmptyCtx Ctx.::> Symbolic.ArchRegStruct arch)
+        (Symbolic.ArchRegStruct arch)
+    ) ->
+  IO
+    ( C.Reg.SomeCFG
+        (Symbolic.MacawExt arch)
+        (Ctx.EmptyCtx Ctx.::> Symbolic.ArchRegStruct arch)
+        (Symbolic.ArchRegStruct arch)
+    )
```
RyanGlScott: Should this be `IO a`?

langston-barrett: Yes, good suggestion!

PR #364, opened 2025-09-10
ci: Upload HTML coverage report as an artifact

### LINKED ISSUE #356: ci: Upload HTML coverage report as an artifact
It would be nice to upload the full HTML coverage report generated during each CI build as a downloadable artifact, so that developers can easily access coverage information without having to recompile GREASE from scratch. This would also enable us to compare coverage across short time-scales (e.g., less than the artifact retention period) or between different branches.

langston-barrett:
...so that developers can download and inspect it.

Fixes #356.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -123,6 +123,14 @@ jobs:
         printf '```\n\n' >> "${GITHUB_STEP_SUMMARY}"
         printf '</details>\n' >> "${GITHUB_STEP_SUMMARY}"
 
+        ./scripts/coverage.sh markup --destdir=coverage-report
+
+    - name: Create coverage report artifact
+      uses: actions/upload-artifact@v4
+      with:
+        name: coverage-report-${{ matrix.ghc }}-${{ matrix.cache-key-prefix }}
+        path: coverage-report/
+
     - name: Test grease-diagrams
       # TODO: Make sure there's no diff
       run: cabal run exe:grease-diagrams -- --output doc/mem.svg --width 400 --height 400


langston-barrett:
Here it is in action: https://github.com/GaloisInc/grease/actions/runs/17625231487?pr=364

.github/workflows/ci.yaml
```diff
@@ -123,6 +123,14 @@ jobs:
         printf '```\n\n' >> "${GITHUB_STEP_SUMMARY}"
         printf '</details>\n' >> "${GITHUB_STEP_SUMMARY}"
 
+        ./scripts/coverage.sh markup --destdir=coverage-report
+
+    - name: Create coverage report artifact
+      uses: actions/upload-artifact@v4
+      with:
+        name: coverage-report-${{ matrix.ghc }}-${{ matrix.cache-key-prefix }}
```
RyanGlScott, line 131: A very minor thing, but for jobs without a `cache-key-prefix`, this results in artifact names like `coverage-report-9.10.2-` with a trailing hyphen. I don't suppose there is an easy way to remove the trailing hyphen?

RyanGlScott (approved):
LGTM, but one quibble below.

.github/workflows/ci.yaml
langston-barrett: Yeah, we could presumably do something like `${{ matrix.ghc }}${{ matrix.cache-key-prefix == "" && "" || join(["-", matrix.cache-key-prefix ]) }}`. I think that's kind of ugly. I don't have any real preference between an ugly filename and ugly logic in the workflow, let me know if you do! 🤷

RyanGlScott: I don't have a strong preference either. If it's ugly to fix, then I'm content with keeping it the way it is.

PR #367, opened 2025-09-11
Allow address overrides to abort

langston-barrett:
Previously, GREASE added an extraneous extra failing proof obligation if an override aborted with `AssertionFailure`. This led to some confusing behavior around concretization, where data was concretized for the proof obligation `false`. Instead, simply re-raise *any* abort reason that comes from an address override.

### DIFF
--- a/grease/src/Grease/Macaw/Overrides/Address.hs
+++ b/grease/src/Grease/Macaw/Overrides/Address.hs
@@ -266,13 +266,13 @@ runAddressOverride crucState someCfg = do
     toInitialState crucState C.UnitRepr $
       C.runOverrideSim C.UnitRepr $
         C.regValue <$> C.callCFG cfg regs
-  C.withBackend (C.execStateContext initState) $ \bak -> do
+  C.withBackend (C.execStateContext initState) $ \_bak -> do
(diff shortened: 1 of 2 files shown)


grease/src/Grease/Macaw/Overrides/Address.hs
```diff
@@ -266,13 +266,13 @@ runAddressOverride crucState someCfg = do
     toInitialState crucState C.UnitRepr $
       C.runOverrideSim C.UnitRepr $
         C.regValue <$> C.callCFG cfg regs
-  C.withBackend (C.execStateContext initState) $ \bak -> do
+  C.withBackend (C.execStateContext initState) $ \_bak -> do
```
langston-barrett, line 269: (note: this is necessary to bring some constraints into scope, even if it isn't used)

RyanGlScott: In that case, you might be able to get away with using `ctxSolverProof` instead of `withBackend`.

langston-barrett: I'm going to defer this because this PR is to unblock some time-sensitive downstream work!

PR #372, opened 2025-09-12
Refactoring LLVM S-expression override code

langston-barrett:
Like #371, but for (non-address) S-expression overrides.

### DIFF
--- a/grease/src/Grease/LLVM/Overrides/SExp.hs
+++ b/grease/src/Grease/LLVM/Overrides/SExp.hs
@@ -67,8 +67,8 @@ instance PP.Pretty LLVMSExpOverrideError where
             , PP.pretty err
             ]
 
--- | A 'CLLVM.SomeLLVMOverride' universally quantified over @p@ and @sym@.
-data AnyLLVMOverride
+-- | A 'CLLVM.SomeLLVMOverride' quantified at a higher rank over @p@ and @sym@.
+newtype AnyLLVMOverride
   = AnyLLVMOverride (forall p sym. CLLVM.SomeLLVMOverride p sym CLLVM.LLVM)
 
 -- | An LLVM function override, corresponding to a single S-expression file.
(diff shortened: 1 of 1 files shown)


grease/src/Grease/LLVM/Overrides/SExp.hs
```diff
@@ -108,30 +112,70 @@ acfgToSomeLLVMOverride ::
   FilePath {- The file which defines the CFG's function.
               This is only used for error messages. -} ->
   C.Reg.AnyCFG CLLVM.LLVM ->
-  IO (CLLVM.SomeLLVMOverride p sym CLLVM.LLVM)
+  Either LLVMSExpOverrideError AnyLLVMOverride
```
RyanGlScott: Given that this now returns an `AnyLLVMOverride` instead of a `SomeLLVMOverride`, the name of this function (`acfgToSomeLLVMOverride`) should be updated accordingly.

grease/src/Grease/LLVM/Overrides/SExp.hs
```diff
@@ -29,77 +30,80 @@ import Lang.Crucible.LLVM.Syntax (emptyParserHooks, llvmParserHooks)
...
+      UnsupportedType path err ->
+        PP.nest 2 $
+          PP.vcat
+            [ "Unsupported type in override in file" PP.<+> PP.pretty path
+            , PP.pretty err
+            ]
+
+-- | A 'CLLVM.SomeLLVMOverride' universally quantified over @p@ and @sym@.
+data AnyLLVMOverride
+  = AnyLLVMOverride (forall p sym. CLLVM.SomeLLVMOverride p sym CLLVM.LLVM)
```
RyanGlScott:
```suggestion
newtype AnyLLVMOverride
  = AnyLLVMOverride (forall p sym. CLLVM.SomeLLVMOverride p sym CLLVM.LLVM)
```

grease/src/Grease/LLVM/Overrides/SExp.hs
```diff
@@ -29,77 +30,80 @@ import Lang.Crucible.LLVM.Syntax (emptyParserHooks, llvmParserHooks)
...
+            , "in" PP.<+> PP.pretty path
+            ]
+      UnsupportedType path err ->
+        PP.nest 2 $
+          PP.vcat
+            [ "Unsupported type in override in file" PP.<+> PP.pretty path
+            , PP.pretty err
+            ]
+
+-- | A 'CLLVM.SomeLLVMOverride' universally quantified over @p@ and @sym@.
```
RyanGlScott: Strictly speaking, it's not universally quantifying over `p` and `sym`, since that would suggest that they are bound by the data type itself. I think it's more accurate to say that they are quantified in a higher rank.

grease/src/Grease/LLVM/Overrides/SExp.hs
langston-barrett: Thanks, good catch!

PR #381, opened 2025-09-18
Abort path instead of throwing an exception on unsupported relocations

langston-barrett:
Most unsupported features (e.g., instructions without semantics) cause individual symex paths to abort. This is nice because we can continue to explore other paths, and GREASE might even abduce a precondition that avoids the problematic path after all. Additionally, we can continue to analyze other functions in the binary.

Unsupported relocations were previously handled by throwing an uncaught exception. This commit brings them in line with those other unsupported features for all of the reasons described above.

This commit also causes GREASE to be more conservative about what it calls a refinement success in a few suspicious cases. This will hopefully allow us to avoid some more bugs along the lines of the one fixed in #380.

### DIFF
--- a/grease-cli/tests/llvm/exit.llvm.cbl
+++ b/grease-cli/tests/llvm/exit.llvm.cbl
@@ -10,4 +10,4 @@
 	(let p (ptr 32 0 (bv 32 0)))
     (funcall h p)
     (return ())))
-;; ok()
+;; check "Program exited"
--- a/grease/src/Grease/Heuristic/Result.hs
+++ b/grease/src/Grease/Heuristic/Result.hs
@@ -20,10 +20,13 @@ import Prettyprinter qualified as PP
 
 -- | There is no way to proceed with refinement, for some explicit reason
 data CantRefine
-  = MissingFunc (Maybe String)
+  = Exhausted String
+  | Exit (Maybe Int)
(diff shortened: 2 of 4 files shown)


grease/src/Grease/Heuristic/Result.hs
```diff
@@ -20,22 +20,29 @@ import Prettyprinter qualified as PP
 
 -- | There is no way to proceed with refinement, for some explicit reason
 data CantRefine
-  = MissingFunc (Maybe String)
+  = Exhausted String
+  | Exit (Maybe Int)
+  | MissingFunc (Maybe String)
```
RyanGlScott: Please document what these `Maybe` fields are.

grease-cli/tests/llvm/exit.llvm.cbl
```diff
@@ -10,4 +10,4 @@
 	(let p (ptr 32 0 (bv 32 0)))
     (funcall h p)
     (return ())))
-;; ok()
+;; check "Program exited"
```
RyanGlScott, line 13: Is it worth checking the exit code as well here?

langston-barrett: I would, but I guess Crucible-LLVM treats this as `EarlyExit`, which doesn't have an associated code.

PR #384, opened 2025-09-18
Concretize fresh bytes

2over12:


grease/src/Grease/Syntax/Overrides.hs
```diff
@@ -79,21 +89,29 @@ freshBytes ::
...
+  , sym ~ WE.ExprBuilder scope st fs
+  , bak ~ C.OnlineBackend solver scope st fs
+  , WPO.OnlineSolver solver
   ) =>
+  bak ->
   LCS.RegValue' sym (LCT.StringType WI.Unicode) ->
   LCS.RegValue' sym (LCT.BVType w) ->
   LCS.OverrideSim p sym ext r args ret (LCS.RegValue sym (LCT.VectorType (LCT.BVType 8)))
-freshBytes name0 bv0 =
+freshBytes bak name0 bv0 =
```
langston-barrett, line 102: You don't need to pass `bak` around everywhere, you can use [`ovrWithBackend`](https://hackage-content.haskell.org/package/crucible-0.7.2/docs/Lang-Crucible-Simulator-OverrideSim.html#v:ovrWithBackend).

2over12: i dont think you can since you need onlinesolver for the concretization stuff right? the backend wont have the correct constraints

PR #392, opened 2025-09-24
Rename `grease-cli` to `grease-exe`, split out CLI library

### LINKED ISSUE #250: Split out `Grease.Cli` from the rest of `grease-cli`
Currently, `grease-cli` roughly consists of two parts:

1. `Grease.Cli`, which defines an `optparse-applicative`–based CLI parser for the `grease` executable
2. `Grease.Main`, which implements the rest of the logic used in the `grease` executable

Currently, these are both included in the same library, but I think it would make sense to split out (1) into its own, separate library. The issue is primarily one of dependencies: (2) requires far more dependencies than (1), as (2) must depend on all of the `macaw` backends. This means that it is currently impossible for downstream dependencies to depend on (1) for the CLI parsing logic without bringing in a whole slew of unneeded dependencies.

I propose that we split up the `grease-cli` into two libraries: `grease-cli` for (1) and `grease-exe` for (2). The `grease` executable would then depend on both the `grease-cli` and `grease-exe` libraries. I'm unsure if we want to split these up into two separate `.cabal` packages or if it would suffice to have them live as public sub-libraries in the same `.cabal` package. (If we do the latter, should the `.cabal` package be named `grease-cli.cabal` or `grease-exe.cabal`?)

langston-barrett:
Fixes #250.


cabal/coverage.cabal.project
```diff
@@ -9,5 +9,5 @@ package *
   coverage: False
 package grease
   coverage: True
-package grease-cli
+package grease-exe
   coverage: True
```
RyanGlScott, line 15: Surely we also want coverage for `grease-cli` here?

doc/dev/dev.md
RyanGlScott, line 1: GitHub won't let me highlight it, but this part of the file also needs to be updated:

https://github.com/GaloisInc/grease/blob/3483aaa11b3cb88a5761ae4020b48823bbf6b0bb/doc/dev/dev.md?plain=1#L150-L158

RyanGlScott (approved):

PR #393, opened 2025-09-24
cli: Improvements around symbolic I/O

### LINKED ISSUE #390: cli: Add flag for specifying symbolic files
Along the lines of https://github.com/GaloisInc/grease/issues/228 and https://github.com/GaloisInc/grease/issues/324, it would be nice to have a CLI flag that allows users to directly specify the paths and sizes of symbolic files, rather than having to create an `--fs-root` on disk. I imagine this might look like the following: `--sym-file SIZE:PATH`.

At that point, we'll have three flags related to symbolic I/O, so we might consider grouping them into their own struct, and adding an option group as is done [here](https://github.com/GaloisInc/grease/blob/9186c27c4f40c8bb5ec5898f4cb0d6e4f24f524d/grease-cli/src/Grease/Cli.hs#L158) for options related to the initial precondition.

This will involve editing the following block of code:

https://github.com/GaloisInc/grease/blob/9186c27c4f40c8bb5ec5898f4cb0d6e4f24f524d/grease-cli/src/Grease/Main.hs#L515-L548

It would be good to factor that chunk into a module in the `grease` library, perhaps `Grease.SymIO` that takes the new struct as a parameter (rather than all of `SimOpts`).

In the future, we should also probably consider adding flags for initializing various network interfaces with symbolic data.

langston-barrett:
Fixes #390.

### DIFF
--- a/grease/src/Grease/SymIO.hs
+++ b/grease/src/Grease/SymIO.hs
@@ -40,14 +40,14 @@ initialLlvmFileSystem ::
   sym ->
   GO.FsOpts ->
   IO (InitializedFs sym ptrW)
-initialLlvmFileSystem halloc sym simOpts = do
+initialLlvmFileSystem halloc sym fsOpts = do
   fileContents_ <-
-    case GO.fsRoot simOpts of
+    case GO.fsRoot fsOpts of
       Nothing -> pure SymIO.emptyInitialFileSystemContents
       Just rootDir -> SymIO.Loader.loadInitialFiles sym rootDir
-  fileContentsWithStdin <- withSymStdin (GO.fsStdin simOpts) fileContents_
+  fileContentsWithStdin <- withSymStdin (GO.fsStdin fsOpts) fileContents_
   fileContentsWithFiles <-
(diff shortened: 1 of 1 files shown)


grease/src/Grease/SymIO.hs
```diff
@@ -0,0 +1,85 @@
...
+-- | Initialize the symbolic file system.
+initialLlvmFileSystem ::
+  ( C.IsSymInterface sym
+  , Mem.HasPtrWidth ptrW
+  ) =>
+  C.HandleAllocator ->
+  sym ->
+  GO.FsOpts ->
+  IO (InitializedFs sym ptrW)
+initialLlvmFileSystem halloc sym simOpts = do
```
RyanGlScott: Perhaps this should be named "`fsOpts`" instead of "`simOpts`s", given that it's of type `FsOpts` instead of `SimOpts`.

RyanGlScott (approved):

grease/src/Grease/SymIO.hs
langston-barrett: Good catch, thanks!

PR #398, opened 2025-09-25
Refactoring in `Grease.{Main,Refine}`

### LINKED ISSUE #397: refactor: De-callback-ify the refinement loop
GREASE's `refinementLoop` takes [a callback](https://github.com/GaloisInc/grease/blob/a8353a47fb2edc5106dc5c1128828359413e3671/grease/src/Grease/Refine.hs#L591) as an argument. In `Grease.Main`, `refinmentLoop` is called once in `simulateMacawCfg`:

https://github.com/GaloisInc/grease/blob/a8353a47fb2edc5106dc5c1128828359413e3671/grease-exe/src/Grease/Main.hs#L833

and once in `simulateLlvmCfg`:

https://github.com/GaloisInc/grease/blob/a8353a47fb2edc5106dc5c1128828359413e3671/grease-exe/src/Grease/Main.hs#L1403

The callbacks in both instances perform a lot of similar work:

1. They call `setup` to generate initial argument values and memory
2. They initialize:
  - the LLVM filesystem (`initialLlvmFileSystem`)
  - error maps (`buildErrMaps`)
  - `ToConcretize`
  - `Conc.InitialState`
3. They pass all of that data to `execAndRefine`

The parts where they really differ are:

- Creation of an initial Crucible state (`initState`), which depends on the different language extensions.
- The Macaw call to `execAndRefine` catches missing semantics exceptions

We could probably do a lot to share more code between these two, ideally with the end result of substantially shrinking or even eliminating the callback.

langston-barrett:
I'm not anywhere near satisfied, but I think that this is a few steps in the right direction. Fixes #397, at least enough for now.

### DIFF
--- a/grease/src/Grease/Refine.hs
+++ b/grease/src/Grease/Refine.hs
@@ -123,6 +123,7 @@ import Grease.Concretize.ToConcretize qualified as ToConc
 import Grease.Cursor qualified as Cursor
 import Grease.Cursor.Pointer qualified as PtrCursor
 import Grease.Diagnostic
+import Grease.ExecutionFeatures (boundedExecFeats)
 import Grease.Heuristic
 import Grease.Options (BoundsOpts)
 import Grease.Options qualified as Opts
@@ -565,7 +566,7 @@ execAndRefine bak _fm la memVar refineData bbMapRef execData = do
       ProveCantRefine (Timeout{}) -> "symex timeout"
       ProveCantRefine (Unsupported{}) -> "unsupported feature"
 
--- TODO: Fold into `refinementLoop`?
+-- | Run 'Setup.setup' then 'execAndRefine'. Usually passed to 'refinementLoop'.
 refineOnce ::
   ( Mem.HasPtrWidth wptr
   , C.IsSyntaxExtension ext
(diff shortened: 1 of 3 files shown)


grease/src/Grease/Refine.hs
```diff
@@ -568,6 +555,81 @@ execAndRefine bak solver _fm la memVar anns heuristics argNames argShapes initSt
       ProveCantRefine (Timeout{}) -> "symex timeout"
       ProveCantRefine (Unsupported{}) -> "unsupported feature"
 
+-- TODO: Fold into `refinementLoop`?
+refineOnce ::
```
2over12, line 570: Id probably tack on a doc here since it is exported and clients need to know to pass it to the loop

grease/src/Grease/Refine.hs
```diff
@@ -568,6 +555,81 @@ execAndRefine bak solver _fm la memVar anns heuristics argNames argShapes initSt
...
+  (toConc, globals1) <- liftIO $ ToConc.newToConcretize halloc (GSIO.initFsGlobals initFs_)
+  let initFs = initFs_{GSIO.initFsGlobals = globals1}
+  st <- mkInitState toConc setupMem initFs args
+  let concInitState =
+        Conc.InitialState
+          { Conc.initStateArgs = args
+          , Conc.initStateFs = GSIO.initFsContents initFs
+          , Conc.initStateMem = initMem
+          }
+  let boundsOpts = Opts.simBoundsOpts simOpts
```
2over12, line 625: it's maybe a bit odd the one can setup refineonce with features that dont implement the bounds declared here, maybe either just pass the timeout or just setup the bounds features here?

2over12 (approved):

grease/src/Grease/Refine.hs
langston-barrett: Good call, done!

grease/src/Grease/Refine.hs
langston-barrett: Whoops, yep!

PR #399, opened 2025-09-26
Testing GHC 9.10.2 panic

langston-barrett:
Do not merge! Saw a panic locally, seeing if I can reproduce in CI.


langston-barrett:
Diagnosed in GHC, so we can safely delete this branch.

PR #400, opened 2025-09-26
Only consider bugs reachabile if all other goals succeed

langston-barrett:
Previously, it was possible for refinement to report a likely bug (`ProveBug`) even if there were other errors along the path that would actually prevent that bug from being reachable in practice. Fix this by only considering a bug reachable if all other goals pass.


grease/src/Grease/Refine.hs
```diff
@@ -214,12 +215,20 @@ combiner = C.Combiner $ \mr1 mr2 -> do
...
   case C.subgoalResult r1 of
     -- can't refine further, no matter what the other result would be
-    ProveBug{} -> pure r1
     ProveCantRefine{} -> pure r1
     -- if we find a refinement, don't bother with other goals
     ProveRefine{} -> pure r1
     -- if this goal succeeded, continue with the others
     ProveSuccess -> mr2
+    -- a bug is only reachable if there are no other errors
+    ProveBug{} -> do
```
2over12, line 224: Should we be returning r1 in the case below here for ProveBug under ProveNoHeuristic wouldnt that mean we end up in a case where we have some unsolved goals and then we return a bug anyways?

2over12:
An additional comment is in this new world i think we have to be careful about what it means for 2 bugs to exist. Im not gonna block the PR on this but downstream dependents might need to think about this. If we have two provebugs which is reached first/actually consistent. A prove bug might mean downstream bugs are actually undefined

grease/src/Grease/Refine.hs
langston-barrett: Good catch, I should update that case to treat these symmetrically.

langston-barrett:
re: multiple bugs: https://github.com/GaloisInc/grease/issues/401

PR #404, opened 2025-10-01
Errors

langston-barrett:
Towards #4.

- Replace some `GreaseException`s with `panic`s
- Add some explicit error enumerations, categorize them (e.g., as user errors) in `Grease.Main`
- Create new, more specific types of exception

### DIFF
--- a/grease-exe/src/Grease/Main.hs
+++ b/grease-exe/src/Grease/Main.hs
@@ -110,6 +110,7 @@ import Grease.Concretize.ToConcretize qualified as ToConc
 import Grease.Diagnostic
 import Grease.Diagnostic.Severity (Severity)
 import Grease.Entrypoint
+import Grease.Error (GreaseException (GreaseException))
 import Grease.ExecutionFeatures (greaseExecFeats)
 import Grease.Heuristic
 import Grease.LLVM qualified as LLVM
--- a/grease-exe/src/Grease/Main.hs
+++ b/grease-exe/src/Grease/Main.hs
@@ -188,6 +188,7 @@ import Lumberjack qualified as LJ
 import Prettyprinter qualified as PP
 import Prettyprinter.Render.Text qualified as PP
 import System.Directory (Permissions, getPermissions)
+import System.Exit as Exit
 import System.FilePath (FilePath)
 import System.IO (IO)
 import Text.LLVM qualified as L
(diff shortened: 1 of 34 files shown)


langston-barrett:
More remains to be done, but for the sake of a reasonably sized diff I'll stop here for now.

grease-exe/src/Grease/Main.hs
```diff
@@ -195,22 +200,43 @@ import What4.ProgramLoc qualified as W4
...
 doLog la diag = LJ.writeLog la (MainDiagnostic diag)
 
+malformedElf :: GreaseLogAction -> PP.Doc Void -> IO a
+malformedElf la doc = do
+  doLog la (Diag.MalformedElf doc)
+  throw (GreaseException (Text.pack (show doc)))
+
+userError :: GreaseLogAction -> PP.Doc Void -> IO a
+userError la doc = do
+  doLog la (Diag.UserError doc)
```
2over12, line 227: sorry is it normal to log an exception then throw an exception with the same text, feels a bit weird since theyll both land in stderr?

grease-exe/src/Grease/Main.hs
```diff
@@ -1209,17 +1251,21 @@ loadAddrOvs ::
...
   IO (AddressOverrides arch)
-loadAddrOvs archCtx halloc memory simOpts = do
+loadAddrOvs la archCtx halloc memory simOpts = do
   mbAddrOvs <- loadAddressOverrides (regStructRepr archCtx) halloc memory (simAddressOverrides simOpts)
   case mbAddrOvs of
-    Left err -> do
-      let msg = PP.renderStrict (PP.layoutPretty PP.defaultLayoutOptions (PP.pretty err))
-      throw (GreaseException ("user error: " <> msg))
+    -- See Note [Explicitly listed errors]
+    Left e@AddrOv.BadAddress{} -> userError la (PP.pretty e)
```
2over12: not sure if it save much but could maybe bind a local function (Pretty e) =>  e -> IO unit while still matching over all of them that closes over la as well

2over12 (approved):
looks great, appreciate it

grease-exe/src/Grease/Main.hs
langston-barrett: Good catch. I was planning on reworking this later, but I might as well do so now. The ideal outcome here IMO is to simply `exit(1)` here instead of throwing an exception. The problem is that the test suite doesn't capture logs that happen before exceptions, which we would want to match on. That's fixable though, and worth fixing.

grease-exe/src/Grease/Main.hs
langston-barrett: Yeah definitely worth it, was just lazy 😄

PR #409, opened 2025-10-03
Errors pt 3

### LINKED ISSUE #4: Structured error messages
Currently, we use a single `GreaseException :: Text -> GreaseException` type to categorize all forms of errors in `grease`. This makes it difficult to selectively catch certain types of exceptions, as the only way to tell them apart is to match on the particular message in the `Text` field. We should consider using a structured error message representation that encodes each type of error as a separate data constructor, similarly to what we do with `Diagnostic`.

langston-barrett:
Builds on #408, fixes #4.


grease/src/Grease/Skip.hs
```diff
@@ -154,12 +154,13 @@ registerSkipOverride ::
   GreaseLogAction ->
   Mem.DataLayout ->
   C.GlobalVar Mem ->
+  GO.CantResolveOverrideCallback sym ext ->
   W4.FunctionName ->
   C.FnHandle args ret ->
   C.OverrideSim p sym ext r args' ret' ()
-registerSkipOverride la dl memVar funcName hdl =
+registerSkipOverride la dl memVar errCb funcName hdl =
   case createSkipOverride la dl memVar funcName (C.handleReturnType hdl) of
```
2over12, line 162: Sorry this is only approximately related to your PR but if im reading this right, skip overrides will be fallen back to only in the case that a user forward delcares a signature and never binds a body for it. I guess this means we can default to just filling in a minimal shape if a user gives us a stub for missing functions or w/e, but does this match behavior we want, seems kinda unintuitive that instead of returning an error if a forward decl isnt bound that we will fill in some symbolic variables? I could be misreading

langston-barrett: That's right! It's by analogy to LLVM where skip overrides are created for functions that are `declare`d but not `define`d. If we didn't do this, it's not clear how we would test the corresponding functionality (perhaps we would have a `--skip` flag or similar?). I agree that it could be misleading. This might be a good option for a warning flag (#413).

Issue #165, opened 2025-05-14
hlint: Add more conventional names for qualified imports


langston-barrett:
In [our `hlint` configuration](https://github.com/GaloisInc/grease/blob/006b19fcff4f524e0ee2c998033e712ee79f6fde/.hlint.yaml#L20), we have a list of conventional names for qualified imports of various modules, e.g., the following line enforces that `Control.Monad` can only be imported qualified as `Monad`:

https://github.com/GaloisInc/grease/blob/006b19fcff4f524e0ee2c998033e712ee79f6fde/.hlint.yaml#L23

This kind of consistency is nice, because it allows developers to know the exact origin of a qualified name at a glance.

We should add some more lines for more modules that we import qualified. Ideally, it would be nice to write a little script that collects the list from `.hlint.yaml`, collects all the import statements from our Haskell sources, and prints a prioritized lists of modules that are frequently imported qualified, but have no entry in `.hlint.yaml`.

PR #417, opened 2025-10-07
Add more conventional names for qualified imports

langston-barrett:
Towards #165.


langston-barrett:
Again, there's plenty more to be done but will leave this here for the sake of diff size + allowing downstream consumers to adjust a bit at a time.

2over12 (approved):

PR #420, opened 2025-10-08
Fail on errors in entrypoint parsing in raw mode

### LINKED ISSUE #419: Uninformative Errors for Address Entrypoints in Raw Mode
https://github.com/GaloisInc/grease/blob/16a315d8988d0c2f59d1a1559fa280495c9b878e/grease-exe/src/Grease/Main.hs#L1879C9-L1879C29 is poorly written. We do a bunch of fallible operations then [mapMaybe](https://github.com/GaloisInc/grease/blob/16a315d8988d0c2f59d1a1559fa280495c9b878e/grease-exe/src/Grease/Main.hs#L1887) over them. This setup swallows errors related to entrypoint parsing. We should just fail if the user supplied a bad entrypoint and give an informative message

2over12:
Closes #419

### DIFF
--- a/grease-exe/src/Grease/Main.hs
+++ b/grease-exe/src/Grease/Main.hs
@@ -1879,15 +1879,28 @@ simulateMacawRaw la memory halloc archCtx simOpts parserHooks =
       mbOrError doc = Maybe.maybe (userError la doc) pure
     -- When loading a binary in raw mode
     -- symbol entrypoints cannot be used
-    let
-      getAddressEntrypoint :: Entrypoint -> IO (Maybe FilePath, Text.Text, MM.MemSegmentOff (MC.RegAddrWidth (MC.ArchReg arch)))
-      getAddressEntrypoint ent = do
-        (txt, fpath) <- case ent of
-          Entrypoint{entrypointLocation = EntrypointAddress x, entrypointStartupOvPath = override} -> pure (x, override)
-          e -> userError la ("An entrypoint was provided that was not an address in raw mode:" PP.<+> PP.viaShow e)
-        intAddr <- mbOrError ("Provided address was not parseable:" PP.<+> PP.pretty txt) (readMaybe $ Text.unpack txt :: Maybe Integer)
-        addr <- mbOrError ("A provided address entrypoint could not be resolved to an addressin memory:" PP.<+> PP.pretty intAddr) (MM.resolveAbsoluteAddr memory $ fromIntegral intAddr)
-        pure (fpath, txt, addr)
+    let getAddressEntrypoint ent = do
+          (txt, fpath) <- case ent of
+            Entrypoint{entrypointLocation = EntrypointAddress x, entrypointStartupOvPath = override} -> pure (x, override)
+            e ->
+              userError
+                la
+                ( "An entrypoint was provided that was not an address in raw mode:"
+                    PP.<+> PP.pretty e
+                )
+          intAddr <-
+            mbOrError
+              ( "Provided address was not parseable:"
+                  PP.<+> PP.pretty txt
+              )
+              (readMaybe $ Text.unpack txt :: Maybe Integer)
+          addr <-
+            mbOrError
+              ( "A provided address entrypoint could not be resolved to an address in memory:"
+                  PP.<+> PP.pretty intAddr
+              )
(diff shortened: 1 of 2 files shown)


grease-exe/src/Grease/Main.hs
```diff
@@ -1874,17 +1874,22 @@ simulateMacawRaw ::
...
-          addr <- MM.resolveAbsoluteAddr memory $ fromIntegral intAddr
-          pure (fpath, txt, addr)
-
-    let entryAddrs :: [(Maybe FilePath, Text.Text, MM.MemSegmentOff (MC.RegAddrWidth (MC.ArchReg arch)))] = Maybe.mapMaybe getAddressEntrypoint entries
+    let
+      getAddressEntrypoint :: Entrypoint -> IO (Maybe FilePath, Text.Text, MM.MemSegmentOff (MC.RegAddrWidth (MC.ArchReg arch)))
+      getAddressEntrypoint ent = do
+        (txt, fpath) <- case ent of
+          Entrypoint{entrypointLocation = EntrypointAddress x, entrypointStartupOvPath = override} -> pure (x, override)
+          e -> userError la ("An entrypoint was provided that was not an address in raw mode:" PP.<+> PP.viaShow e)
```
langston-barrett: Looks like `Entrypoint` has a `Pretty` instance, does this work?
```suggestion
          e -> userError la ("An entrypoint was provided that was not an address in raw mode:" PP.<+> PP.pretty e)
```

grease-exe/src/Grease/Main.hs
```diff
@@ -1874,17 +1874,22 @@ simulateMacawRaw ::
...
+        intAddr <- mbOrError ("Provided address was not parseable:" PP.<+> PP.pretty txt) (readMaybe $ Text.unpack txt :: Maybe Integer)
+        addr <- mbOrError ("A provided address entrypoint could not be resolved to an addressin memory:" PP.<+> PP.pretty intAddr) (MM.resolveAbsoluteAddr memory $ fromIntegral intAddr)
```
langston-barrett: These lines are quite long, perhaps consider `let`-binding the error messages and / or the values.

langston-barrett:
```suggestion
        addr <- mbOrError ("A provided address entrypoint could not be resolved to an address in memory:" PP.<+> PP.pretty intAddr) (MM.resolveAbsoluteAddr memory $ fromIntegral intAddr)
```

grease-exe/src/Grease/Main.hs
```diff
@@ -1874,17 +1874,22 @@ simulateMacawRaw ::
...
+        pure (fpath, txt, addr)
+
+    entryAddrs :: [(Maybe FilePath, Text.Text, MM.MemSegmentOff (MC.RegAddrWidth (MC.ArchReg arch)))] <- forM entries getAddressEntrypoint
```
langston-barrett, line 1905: Is the type signature necessary here? If there's some kind of ambiguity, perhaps just partially type-applying `getAddressEntrypoint @arch` would work?

grease-exe/tests/sanity/pass/simple-static/test.c
```diff
@@ -7,3 +7,7 @@ int test(int *x) { return *x; }
 // all: flags {"--load-base", "0x1000", "--address", "0x1000", "--raw-binary"}
 // arm: go(prog)
 // arm: ok()
+
+// all: flags {"--load-base", "0x1000", "--symbol", "test", "--raw-binary"}
```
langston-barrett:
```suggestion
// arm: flags {"--load-base", "0x1000", "--symbol", "test", "--raw-binary"}
```

grease-exe/tests/sanity/pass/simple-static/test.c
```diff
@@ -7,3 +7,7 @@ int test(int *x) { return *x; }
 // all: flags {"--load-base", "0x1000", "--address", "0x1000", "--raw-binary"}
```
langston-barrett: Looks like `go` is just called for `arm`, might as well do the same for `flags`?
```suggestion
// arm: flags {"--load-base", "0x1000", "--address", "0x1000", "--raw-binary"}
```

Issue #391, opened 2025-09-24
Switch from Macaw's `resolveSymBV` to What4's `concretize`


langston-barrett:
https://github.com/GaloisInc/what4/pull/264 will add new functionality to What4 for concretizing symbolic expressions. We should use this to replace two uses of Macaw's `resolveSymBV` in GREASE:

https://github.com/GaloisInc/grease/blob/9186c27c4f40c8bb5ec5898f4cb0d6e4f24f524d/grease/src/Grease/Macaw/ResolveCall.hs#L267-L282

https://github.com/GaloisInc/grease/blob/9186c27c4f40c8bb5ec5898f4cb0d6e4f24f524d/grease/src/Grease/Macaw/Overrides/Networking.hs#L898

Both uses don't rely on the bounds search behavior, which potentially carries a heavy performance hit.

PR #431, opened 2025-10-20
Preserve thumb mode in resolve call

### LINKED ISSUE #430: GREASE does not persist thumb mode during resolve call
This issue is tracking the current likely cause of some downstream decoding problems. When we call resolve call inside of a function that has PSTATE_T=1 we throw away this context. 

We call discover here https://github.com/GaloisInc/grease/blob/94cb2e57fc030c883772aa12363c7ab7888cb699/grease/src/Grease/Macaw/ResolveCall.hs#L380 then

Discover function in Macaw is called here: https://github.com/GaloisInc/grease/blob/94cb2e57fc030c883772aa12363c7ab7888cb699/grease/src/Grease/Macaw/ResolveCall.hs#L594

 Per https://github.com/GaloisInc/macaw/blob/754422bf2bb901540edce82e6a3616fb1580ae1e/macaw-aarch32/src/Data/Macaw/ARM/Disassemble.hs#L35 we would need to know the mode to get this to disassemble correctly.

We will need to somehow initialize discoverFunction in the right context

2over12:
Closes #430

### DIFF
--- a/grease-aarch32/src/Grease/Macaw/Arch/AArch32.hs
+++ b/grease-aarch32/src/Grease/Macaw/Arch/AArch32.hs
@@ -22,7 +22,6 @@ import Data.Macaw.Memory qualified as MM
 import Data.Macaw.Symbolic qualified as Symbolic
 import Data.Macaw.Symbolic.Concretize qualified as Symbolic
 import Data.Map qualified as Map
-import Data.Maybe (fromJust)
 import Data.Parameterized.Classes (ixF')
 import Data.Parameterized.Context qualified as Ctx
 import Data.Parameterized.NatRepr (knownNat)


grease-aarch32/src/Grease/Macaw/Arch/AArch32.hs
```diff
@@ -86,9 +100,28 @@ armCtx halloc mbReturnAddr stackArgSlots = do
...
+  , bak ~ C.OnlineBackend solver scope st fs
+  , arch ~ ARM.ARM
+  ) =>
+  bak ->
+  Ctx.Assignment (CS.RegValue' sym) (Symbolic.MacawCrucibleRegTypes arch) ->
+  MC.ArchSegmentOff arch ->
+  IO (MC.ArchSegmentOff arch)
+armArchPCFixup bak regs origAddr = do
+  let C.RV (CLM.LLVMPointer _base off) = regs ^. ixF' ARM.Symbolic.Regs.pstateT
+  raddr <- Symbolic.resolveSymBV bak C.knownNat off
```
langston-barrett, line 122: Let's add a TODO linking to #391.

grease-aarch32/src/Grease/Macaw/Arch/AArch32.hs
```diff
@@ -86,8 +98,30 @@ armCtx halloc mbReturnAddr stackArgSlots = do
       , _archOffsetStackPointerPostCall = pure
       , -- assumes AAPCS32 https://github.com/ARM-software/abi-aa/blob/main/aapcs32/aapcs32.rst#parameter-passing
         _archABIParams = Some.Some <$> [ARM.r0, ARM.r1, ARM.r2, ARM.r3]
+      , _archPCFixup = armArchPCFixup
       }
 
+armArchPCFixup ::
```
langston-barrett, line 106: Let's add a Haddock saying why something nontrivial is necessary here, and how incrementing the address by 1 takes care of that.

grease-aarch32/src/Grease/Macaw/Arch/AArch32.hs
```diff
@@ -86,8 +98,30 @@ armCtx halloc mbReturnAddr stackArgSlots = do
...
+  MC.ArchSegmentOff arch ->
+  IO (MC.ArchSegmentOff arch)
+armArchPCFixup bak regs origAddr = do
+  let C.RV (CLM.LLVMPointer _base off) = regs ^. ixF' ARM.Symbolic.Regs.pstateT
+  -- TODO(#391): What4 is adding direct concretize functionality.
+  -- We should replace the use of macaw here.
+  raddr <- Symbolic.resolveSymBV bak C.knownNat off
+  pure $ case WI.asBV raddr of
+    Nothing -> panic "armArchPCFixup" ["PSTATE_T should always be concrete"]
+    Just bv -> if BV.testBit' 0 bv then fromJust $ MM.incSegmentOff origAddr 1 else origAddr
```
langston-barrett: Rather than use `fromJust`, perhaps we could pattern match on the result and `panic` in the case where it fails? That will give a more informative message if this ever does fail.

grease-exe/tests/sanity/pass/thumb-calls/test.c
```diff
@@ -0,0 +1,11 @@
...
+
+// Simple call test to make sure calling thumb mode functions works.
+
+__attribute__((target("thumb"))) int callee(int *x) { return *x; }
+
+__attribute__((target("thumb"))) int test(int *x) { return callee(x); }
+
+// arm: flags {"--address", "0x1008b"}
+// arm: go(prog)
+// arm: uninit_stack()
```
langston-barrett, line 13: Why should this result in an uninitialized stack read?

langston-barrett (commented):
I'm confused about the behavior on the test-case but otherwise looks fine.

grease-aarch32/src/Grease/Macaw/Arch/AArch32.hs
2over12: This should be effectively impossible right? this assumes we found a call to a function, resolved it correctly and it was the last byte of the segment

grease-exe/tests/sanity/pass/thumb-calls/test.c
2over12: It's unclear to me as well... im not super happy about it. I can try to find some time to look into this as well i was mostly testing that it now decoded properly and treating this as a separate issue. We could also open an issue to investigate. Im pretty time bounded for this specific thing

langston-barrett: I'm fine merging, let's just open an issue and have a nubered TODO here.

grease-aarch32/src/Grease/Macaw/Arch/AArch32.hs
langston-barrett: Alright, let's at least add a comment saying as much.

langston-barrett (approved):
Looks good, let's just be sure to squash the commits since several of them don't pass CI.

grease-exe/tests/sanity/pass/thumb-calls/test.c
RyanGlScott: This was spun off into https://github.com/GaloisInc/grease/issues/432.

PR #434, opened 2025-10-21
Bump Crucible submodule, adjust to debugger API changes

langston-barrett:
Along the way, creates an actual personality type for LLVM.

### DIFF
--- a/grease-exe/src/Grease/Main.hs
+++ b/grease-exe/src/Grease/Main.hs
@@ -118,6 +118,7 @@ import Grease.LLVM qualified as LLVM
 import Grease.LLVM.DebugInfo qualified as GLD
 import Grease.LLVM.Overrides qualified as GLO
 import Grease.LLVM.Overrides.SExp qualified as GLOS
+import Grease.LLVM.Personality qualified as GLP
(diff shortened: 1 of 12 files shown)


grease-exe/src/Grease/Main.hs
```diff
@@ -810,7 +808,7 @@ macawInitState ::
   , wptr ~ MC.ArchAddrWidth arch
   , ext ~ Symbolic.MacawExt arch
   , ret ~ Symbolic.ArchRegStruct arch
-  , p ~ GreaseSimulatorState sym arch
+  , p ~ GreaseSimulatorState MDebug.MacawCommand sym arch
```
2over12, line 811: Guessing we will need downstream changes to consuming personalities?

langston-barrett: Are you asking if this is a breaking change to GREASE's API? Indeed it is.

2over12 (approved):
This looks good to me, downstream we should probably open a PR sooner rather than later, unfortunately that will cause some conflicts i imagine but that's on me

grease-exe/src/Grease/Main.hs
2over12: yeah i just mean there's probably going to be a fair number of changes on the downstream stuff to compensate for this i imagine

PR #437, opened 2025-10-22
Allow setting max thread count in Ghidra plugin

2over12:
This setting helps with memory pressure on memory bound setups

### DIFF
--- a/.github/actions/ghidra-install/action.yaml
+++ b/.github/actions/ghidra-install/action.yaml
@@ -21,5 +21,5 @@ runs:
             --output ~/ghidra.zip \
             https://github.com/NationalSecurityAgency/ghidra/releases/download/Ghidra_11.4.2_build/ghidra_11.4.2_PUBLIC_20250826.zip
         cd ~/ && unzip -x ghidra.zip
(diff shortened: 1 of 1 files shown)


ghidra-plugin/src/main/scala/grease/GreaseAnalyzer.scala
```diff
@@ -214,6 +222,14 @@ class GreaseAnalyzer
       null,
       "Uses the image base of this Ghidra binary as the load base when in raw mode (this is the default, disable to use custom image base)"
     )
+
+    options.registerOption(
+      GreaseAnalyzer.MAX_THREAD_OPT,
+      OptionType.INT_TYPE,
+      0,
+      null,
+      "Set the maximum thread count for the GREASE worker. Default is cpu+1"
```
langston-barrett, line 231: I don't see where this default is set, is that part of `GThreadPool`?

2over12: Indeed the GThreadPool if nothing is set will do this

PR #444, opened 2025-10-28
doc: Add some links to general background section from LWN.net

langston-barrett:
I think these are helpful explainers


2over12 (approved):

langston-barrett:
>  I dont think it should be GREASE's docs job to teach how to do binary analysis

> Id almost prefer a "binary analysis at galois" or something page

That's fair, and agreed! It's just that no such page exists. So I view this as a bit of a stop-gap.

PR #447, opened 2025-11-04
Use global tracking MacawArchStateUpdate to run address overrides

### LINKED ISSUE #416: Address overrides cannot access register state from the middle of a basic block
Address overrides access registers as they are at the start of the block because Macaw does not keep the registers variable up to date (the structure is essentially SROAed until the block ends)

Cross refs from Macaw:
https://github.com/GaloisInc/macaw/pull/534#issuecomment-3312332616
https://github.com/GaloisInc/macaw/issues/460#issue-2818884521

2over12:
Fixes #416


grease/src/Grease/Macaw/SimulatorHooks.hs
```diff
@@ -121,6 +130,30 @@ ptrAnd bak x y = do
...
+
+updateStruct ::
+  (C.OrdF (MC.ArchReg arch), sym ~ W4.ExprBuilder t st fs) =>
+  Ctx.Assignment (MC.ArchReg arch) ctx ->
+  Ctx.Assignment (CS.RegValue' sym) (Symbolic.CtxToCrucibleType ctx) ->
+  MapF.MapF (MC.ArchReg arch) (Symbolic.MacawCrucibleValue (CS.RegEntry (W4.ExprBuilder t st fs))) ->
+  Ctx.Assignment (CS.RegValue' sym) (Symbolic.CtxToCrucibleType ctx)
+updateStruct Ctx.Empty Ctx.Empty _ =
+  Ctx.Empty
+updateStruct (rstRegs Ctx.:> currReg) (rstVals Ctx.:> currVal) regToNewVal =
```
2over12, line 140: i feel like i write this type of manually recursive code a lot when we need to transform the ctx type, am I missing something?

grease/src/Grease/Macaw/SimulatorHooks.hs
```diff
@@ -121,6 +130,30 @@ ptrAnd bak x y = do
...
   pure (CLM.LLVMPointer blk off)
 
+macawAssignToCruc ::
+  (forall tp. f tp -> g (Symbolic.ToCrucibleType tp)) ->
+  Ctx.Assignment f ctx ->
+  Ctx.Assignment g (Symbolic.CtxToCrucibleType ctx)
+macawAssignToCruc f a =
+  case a of
+    Ctx.Empty -> Ctx.empty
+    b Ctx.:> x -> macawAssignToCruc f b Ctx.:> f x
```
langston-barrett: This looks like [`fmapFC`](https://hackage-content.haskell.org/package/parameterized-utils-2.1.10.0/docs/Data-Parameterized-TraversableFC.html#v:fmapFC) of `f`.

grease/src/Grease/Macaw/SimulatorHooks.hs
langston-barrett: This case is a bit tricker because we know that `CtxToCrucibleType` doesn't change the length of `ctx` but I'm not sure how easy it would be to convince GHC of that. If both of these were over `ctx` (which you could probably make happen via a `newtype` around `RegValue` that works on Macaw types and applies `ToCrucibleType` inside the `newtype`), then you could use [`zipWith`](https://hackage-content.haskell.org/package/parameterized-utils-2.1.10.0/docs/Data-Parameterized-Context.html#v:zipWith).

grease/src/Grease/Macaw/SimulatorHooks.hs
2over12: isnt it the case that that fmapFC assumes they are indexed by the same type? anyways this is dead code post removing the init struct stuff

langston-barrett: Yep, quite right. Again, solvable with a `newtype` but it's a moot issue now.

2over12:
Dropping back to draft to add a call test

grease-exe/src/Grease/Main.hs
```diff
@@ -857,6 +858,7 @@ macawInitState ::
   IO (CS.ExecState p sym ext (CS.RegEntry sym ret))
 macawInitState la archCtx halloc macawCfgConfig simOpts bak memVar memPtrTable setupHook addrOvs mbCfgAddr entrypointCfgsSsa toConcVar setupMem initFs args = do
   let sym = CB.backendGetSym bak
+  styVar <- C.freshGlobalVar halloc "grease:archRegs" (Symbolic.crucGenRegStructType $ Symbolic.archFunctions (archCtx ^. archVals))
```
langston-barrett: Where does the name `sty` come from? "Struct type"? Perhaps "regs" might be more semantically meaningful, as in `regsVar`?
```suggestion
  let regsTy = Symbolic.crucGenRegStructType $ Symbolic.archFunctions (archCtx ^. archVals)
  styVar <- C.freshGlobalVar halloc "grease:archRegs" regsTy
```

grease-exe/src/Grease/Main.hs
```diff
@@ -886,7 +888,7 @@ macawInitState la archCtx halloc macawCfgConfig simOpts bak memVar memPtrTable s
 
   let globals = GSIO.initFsGlobals initFs
   let initFsOv = GSIO.initFsOverride initFs
-  initState bak la macawExtImpl halloc memVar setupMem globals initFsOv archCtx setupHook addrOvs personality regs fnOvsMap mbStartupOvSsaCfg ssa
+  initState bak la macawExtImpl halloc memVar styVar setupMem globals initFsOv archCtx setupHook addrOvs personality regs fnOvsMap mbStartupOvSsaCfg ssa
```
langston-barrett: Since `initState` also gets `halloc` and `archCtx`, perhaps the initialization of this variable could happen inside that function? I ask because it already has an unfortunate number of parameters, might be nice to avoid another if we can. In contrast, I think we need `memVar` to have already been created use by other code.

grease-exe/tests/sanity/pass/addr-override-call-mod/test.lua
```diff
@@ -0,0 +1,7 @@
+flags {"--symbol", "test"}
+-- Address of the nop after setting rsi == rdi function from `objdump -d`.
+flags {"--addr-override", "0x10001029:tests/sanity/pass/addr-override-call-mod/extra/addr-override.x86_64.cbl"}
+-- We check that we can observe the change in rsi state from the middle of the block after 
+-- a call
```
langston-barrett, line 5: Does this PR fix #416? If so, probably worth mentioning that issue here, and in the PR description.

grease-exe/tests/sanity/pass/addr-override-call-mod/test.S
```diff
@@ -0,0 +1,18 @@
+# test that at the nop we can read from rsi by checking assert rsi == rdi 
+# this tests that we can observe the result of the function call
```
langston-barrett, line 2: ...inside of an address override specifically, right?

grease-exe/tests/Makefile
```diff
@@ -102,6 +102,10 @@ STANDALONE_SHARED_LIBRARY_X64_SOS    = $(addsuffix /test.x64.elf,$(STANDALONE_SH
 STANDALONE_SHARED_LIBRARY_PPC32_SOS  = $(addsuffix /test.ppc32.elf,$(STANDALONE_SHARED_LIBRARY_DIRS))
 
 
+
+ASSEMBLY_STANDALONE_SHARED_LIBRARY_DIRS = sanity/pass/addr-override-internal-reg-mod sanity/pass/addr-override-call-mod
+ASSEMBLY_SHARED_LIBRARY_X64_SOS    = $(addsuffix /extra/test.x64.elf,$(ASSEMBLY_STANDALONE_SHARED_LIBRARY_DIRS))
```
langston-barrett: Odd whitespace here

grease/src/Grease/Macaw/Overrides/Address.hs
```diff
@@ -373,12 +372,13 @@ tryRunAddressOverride ::
...
+  C.GlobalVar (Symbolic.ArchRegStruct arch) ->
   C.CrucibleState p sym (Symbolic.MacawExt arch) rtp blocks r args ->
   C.SomeCFG (Symbolic.MacawExt arch) (Ctx.EmptyCtx Ctx.::> Symbolic.ArchRegStruct arch) C.UnitType ->
   IO ()
-tryRunAddressOverride archCtx memVar crucState cfg = do
-  let archFns = archCtx Lens.^. archVals . Lens.to Symbolic.archFunctions
-  case DMSR.simStateRegs archFns crucState of
+tryRunAddressOverride archCtx memVar archRegs crucState cfg = do
+  let mbRegs = C.lookupGlobal archRegs (crucState Lens.^. C.stateGlobals)
+  case mbRegs of
```
langston-barrett, line 381: Perhaps just inline? Up to you.

grease/src/Grease/Macaw/SimulatorHooks.hs
```diff
@@ -173,6 +198,22 @@ extensionExec archCtx bak la tgtOvs memVar baseExt stmt crucState = do
             , show baddr
             , show iaddr
             ]
+    Symbolic.MacawArchStateUpdate _ updates ->
+      let nstate =
```
langston-barrett: Is this short for "new state"? Let's use `camelCase` consistently and call it `nState` (or perhaps even `newState`).

grease/src/Grease/Macaw/SimulatorHooks.hs
```diff
@@ -173,6 +198,22 @@ extensionExec archCtx bak la tgtOvs memVar baseExt stmt crucState = do
...
+                            let regs = CS.lookupGlobal archStruct globs
+                                regAssign = Symbolic.crucGenRegAssignment $ Symbolic.archFunctions (archCtx ^. archVals)
+                             in CS.insertGlobal
+                                  archStruct
+                                  ( case regs of
+                                      Just rs -> updateStruct regAssign rs updates
+                                      Nothing -> panic "extensionExec" ["The register struct global should be initialized during initState."]
+                                  )
+                                  globs
+                        )
```
langston-barrett: Maybe worth giving a name to this lambda in the same `let`, just above `nstate` to dedent it a bunch.

langston-barrett (approved):
Looks good, mostly stylistic / readability comments. Feel free to take them or leave them.

One thing that might be nice would be a test involving a Haskell override of a libc function. `getenv` might be a good choice, I think it always returns `NULL`. Would we be able to read the return value in `rax`?

grease-exe/src/Grease/Main.hs
2over12: This requires concretizing initState to be IO instead of MonadIO m, i think this is fine as afaik we never take advantage of the fact that the monad is generic so maybe we should do that?

PR #448, opened 2025-11-05
ci: Enable assertions for Crucible-LLVM, GREASE, What4

### LINKED ISSUE #435: tests: Enable assertions for Crucible-LLVM, What4
Both Crucible-LLVM and What4 use `Control.Exception.assert`. Cabal compiles code with `-O1` by default, which disables these assertions. We should enable them in our test-suite. See https://github.com/GaloisInc/crucible/pull/1585 for an example.

langston-barrett:
Fixes #435.

### DIFF
--- a/doc/dev/dev.md
+++ b/doc/dev/dev.md
@@ -27,6 +27,11 @@ cabal build exe:grease
 
 See [Test suite](tests.md) for how to run the tests.
 
+To build a binary suitable for releases, run the following before `cabal build`:
+```sh
+cat cabal/release.cabal.project >> cabal.project.local
+```
+
 ## Docker
 
 GREASE also offers a nightly Docker image that gets built after each commit to


grease-exe/tests/Main.hs
```diff
@@ -274,4 +295,4 @@ main = do
         , "x86"
         ]
   tests <- mapM discoverTests (map ("tests" </>) dirs)
-  T.defaultMain $ T.testGroup "Tests" (shapeTests : tests)
+  T.defaultMain $ T.testGroup "Tests" (assertTests : shapeTests : tests)
```
2over12, line 298: If im reading this right we are now going to fail the test suite if you build GREASE default? is this a good idea? maybe document in the dev docs that you should copy assert.cabal.project into the local cabal project unless im misreading

langston-barrett: > If im reading this right we are now going to fail the test suite if you build GREASE default?

Right.

> is this a good idea?

Good question - I think we want:

1. Assertions checked when running tests, in CI at least but probably also locally
2. Assurance that assertions haven't been disabled by Cabal/CI configuration changes
3. The ability to disable assertions, e.g., for "release" builds

I agree that it would be nice if `cabal test` passed with no further configuration.

Other options are:

- Have a [Cabal flag](https://cabal.readthedocs.io/en/stable/cabal-project-description-file.html#cfg-field-flags) that enables assertions. It could be on by default, and then we could document in the dev docs how to turn it off for release builds
- Add `-fno-ignore-asserts` to the Cabal files and require a `cabal.project` file with `-fignore-asserts` to override them for release builds

> maybe document in the dev docs

Definitely a good idea regardless, will do for whichever option we like.

2over12: I dont have a strong preference here really... i really would love something like a default developer profile that we set up enabling assertions and then a release profile that we would pass as a switch for release builds (ie. how i would do things with dune), but i dont know how to do such ergonomics with cabal

langston-barrett: Yeah, Cargo also has a notion of profile, and enables these kinds of debug assertions in the "dev" profile and disables them in the "release" profile. We could probably set something similar up using Cabal flags, and just have the dev profile on by default.

langston-barrett: > We could probably set something similar up using Cabal flags, and just have the dev profile on by default.

Ah, thinking about this a bit more, I don't actually think this is possible. We specifically want to enable assertions in our *dependencies*, especially Crucible-LLVM and What4. It's only possible to set flags for dependencies in `cabal.project` files. Using flags is only possible in `.cabal` files, [not in `cabal.project` files](https://cabal.readthedocs.io/en/stable/cabal-project-description-file.html#conditionals-and-imports). So I think that the current solution is the only one that really works. The one change we could consider is putting the assertion machinery in [our top-level `cabal.project`](https://github.com/GaloisInc/grease/blob/d6584e3971cb567837eff67089f4defdf4c14c29/cabal.project) so that the tests would pass with no further intervention, and we could document that you may want to override those stanzas when making a release build.

langston-barrett: I've flipped the default so that assertions are enabled by `cabal.project`. Now, tests will pass by default when cloning and `cabal test`ing. I've included `cabal/release.cabal.project` to override this and disable assertions (and also enable `-O2`), and added that to the Dockerfile (which is our only "release-like" platform at the moment). I didn't include documentation on this since it's a default. Let me know if you'd like to see docs.

cabal/release.cabal.project
```diff
@@ -0,0 +1,9 @@
+-- Options suitable for release builds of GREASE.
+--
```
2over12, line 2: might be worth putting this in dev docs

2over12 (approved):

cabal/release.cabal.project
langston-barrett: Good call, done!

PR #449, opened 2025-11-05
Bump asl-translator, macaw submodules to fix pop/ldm semantics

### LINKED ISSUE #432: GREASE uninitialized stack read in thumb test case
In the [thumb sanity test case](https://github.com/GaloisInc/grease/blob/c32b7810e07221b44e6fdd2291147087688342b6/grease-exe/tests/sanity/pass/thumb-calls/test.c#L11) there is an uninit stack read in pop that is expected to be initialized

RyanGlScott:
This bumps submodules to bring in the following changes:

* https://github.com/GaloisInc/asl-translator/pull/58
* https://github.com/GaloisInc/macaw/pull/547

These fix the `macaw-aarch32` semantics for instructions like `pop` and `ldm` to avoid performing unnecessary memory reads, thereby preventing the issue that was causing the `thumb-calls` test case to incorrectly be flagged as performing an uninitialized stack read. As such, this fixes #432.

Note that `thumb-calls` still doesn't pass `grease` even with these changes, but this is because `macaw`'s code discovery is not smart enough to reckon with a function that ends with a `pop {pc, ...}` statement. Fixing that problem on the `macaw` side would be non-trivial, so I have opted to accept this.

### DIFF
--- a/grease-exe/tests/Makefile
+++ b/grease-exe/tests/Makefile
@@ -119,8 +119,19 @@ SHARED_LIBARY_DWARFV5_X64_SOS = $(addsuffix /test.x64.elf,$(SHARED_LIBARY_DWARFV
 RAW_BINARY_DIRS = sanity/pass/simple-static
 RAW_BINARY_EXES = $(addsuffix /test.armv7l.elf,$(RAW_BINARY_DIRS))
 
-
-STATIC_DIRS = $(filter-out $(DYNAMIC_DIRS) $(DYNAMIC_PIE_DIRS) $(STRIPPED_STATIC_DIRS) $(ASSEMBLY_ARM_DIRS) $(ASSEMBLY_STANDALONE_SHARED_LIBRARY_DIRS) $(STATIC_WITH_LIBS_DIRS) $(STACK_PROTECTOR_DIRS) $(SHARED_LIBRARY_DIRS) $(STANDALONE_SHARED_LIBRARY_DIRS) $(SHARED_LIBARY_DWARFV4_DIRS) $(SHARED_LIBARY_DWARFV5_DIRS) $(RAW_BINARY_DIRS), $(DIRS))
+NON_STATIC_DIRS = $(DYNAMIC_DIRS) \
+                  $(DYNAMIC_PIE_DIRS) \
+                  $(STRIPPED_STATIC_DIRS) \
+                  $(ASSEMBLY_ARM_DIRS) \
+                  $(ASSEMBLY_STANDALONE_SHARED_LIBRARY_DIRS) \
+                  $(STATIC_WITH_LIBS_DIRS) \
+                  $(STACK_PROTECTOR_DIRS) \
+                  $(SHARED_LIBRARY_DIRS) \
+                  $(STANDALONE_SHARED_LIBRARY_DIRS) \
+                  $(SHARED_LIBARY_DWARFV4_DIRS) \
+                  $(SHARED_LIBARY_DWARFV5_DIRS) \
+                  $(RAW_BINARY_DIRS)
+STATIC_DIRS = $(filter-out $(NON_STATIC_DIRS), $(DIRS))
 STATIC_ARM_EXES   = $(addsuffix /test.armv7l.elf,$(STATIC_DIRS))
 # See discussion on !288 for why libpng-cve-2018-13785 appears here.
 STATIC_X64_EXES   = $(addsuffix /test.x64.elf,$(STATIC_DIRS) refine/neg/libpng-cve-2018-13785)


grease-exe/tests/sanity/pass/push-pop/test.S
```diff
@@ -0,0 +1,17 @@
...
+test:
+  push {r7}
+  pop {r7}
+  mov r0, #1
+  bx      lr
+
+.global _start
+.syntax unified
+_start:
+  bl      test
```
langston-barrett: nbd, but a bit of odd spacing on the `bx` and `bl` compared to the other instructions

grease-exe/tests/sanity/pass/thumb-calls/test.c
```diff
@@ -10,4 +10,7 @@ __attribute__((target("thumb"))) int test(int *x) { return callee(x); }
 
 // arm: flags {"--address", "0x1008b"}
 // arm: go(prog)
-// arm: uninit_stack()
+
+// TODO: Ideally, Macaw's code discovery would be smart enough to figure this
+// out, but it currently isn't.
```
langston-barrett, line 15: Should we file a Macaw issue and refer to it here?

grease-exe/tests/Makefile
```diff
@@ -117,7 +120,7 @@ RAW_BINARY_DIRS = sanity/pass/simple-static
 RAW_BINARY_EXES = $(addsuffix /test.armv7l.elf,$(RAW_BINARY_DIRS))
 
 
-STATIC_DIRS = $(filter-out $(DYNAMIC_DIRS) $(DYNAMIC_PIE_DIRS) $(STRIPPED_STATIC_DIRS) $(ASSEMBLY_STANDALONE_SHARED_LIBRARY_DIRS) $(STATIC_WITH_LIBS_DIRS) $(STACK_PROTECTOR_DIRS) $(SHARED_LIBRARY_DIRS) $(STANDALONE_SHARED_LIBRARY_DIRS) $(SHARED_LIBARY_DWARFV4_DIRS) $(SHARED_LIBARY_DWARFV5_DIRS) $(RAW_BINARY_DIRS), $(DIRS))
+STATIC_DIRS = $(filter-out $(DYNAMIC_DIRS) $(DYNAMIC_PIE_DIRS) $(STRIPPED_STATIC_DIRS) $(ASSEMBLY_ARM_DIRS) $(ASSEMBLY_STANDALONE_SHARED_LIBRARY_DIRS) $(STATIC_WITH_LIBS_DIRS) $(STACK_PROTECTOR_DIRS) $(SHARED_LIBRARY_DIRS) $(STANDALONE_SHARED_LIBRARY_DIRS) $(SHARED_LIBARY_DWARFV4_DIRS) $(SHARED_LIBARY_DWARFV5_DIRS) $(RAW_BINARY_DIRS), $(DIRS))
```
langston-barrett: This is getting pretty long, perhaps consider breaking it up with `\`?

grease-exe/tests/sanity/pass/push-pop/test.S
RyanGlScott: I've changed the contents of this file to:

```
.global test
.syntax unified
.type	test, %function
.thumb
test:
  push  {r7}
  pop   {r7}
  mov   r0, #1
  bx    lr

.global _start
.syntax unified
_start:
  bl    test
```

Which more closely resembles the disassembled output that `objdump -d` produces.

grease-exe/tests/Makefile
RyanGlScott: Done in https://github.com/GaloisInc/grease/pull/449/commits/62255efd165fa4d201daf1b14a5100dfde813544.

grease-exe/tests/sanity/pass/thumb-calls/test.c
RyanGlScott: Truth be told, there are quite a few places in the `grease-exe` test suite that check for `Could not identify block`-like error messages, and they are all accompanied by similar TODOs. I like the idea of filing an issue to say "please make Macaw's code discovery smarter", but I would like to make sure that we phrase it in a way where there is something actionable that we can do. (In general, I suspect that Macaw's code discovery will never be perfect, so certain programs may be beyond our reach.)

When I talked to @danmatichuk about Macaw's code discovery failing on this specific test case, he noted that he encountered similar issues in `pate`, and that he had to employ some advanced workarounds. IIRC, he essentially needed to implement his own version of `macaw-refinement` so that he could leverage SMT solvers to resolve odd corner cases. Perhaps we could file an issue about porting this machinery to Macaw as a first-class feature? I'd need to consult with @danmatichuk to figure out the specifics of what this would entail, however. (For instance, I don't know where this code lives in `pate`.)

langston-barrett: Fair enough. I'm fine with merging as-is if this isn't a particularly clear ask.

PR #455, opened 2025-11-07
Dump covered instruction in consumable format

2over12:
Adds the capability to dump covered instructions and a section map in preparation to display coverage

### DIFF
--- a/grease/src/Grease/Macaw/SimulatorHooks.hs
+++ b/grease/src/Grease/Macaw/SimulatorHooks.hs
@@ -169,7 +169,8 @@ extensionExec ::
   CS.GlobalVar (MC.ArchRegStruct arch) ->
   CS.ExtensionImpl p sym (Symbolic.MacawExt arch) ->
   Symbolic.MacawEvalStmtFunc (C.StmtExtension (Symbolic.MacawExt arch)) p sym (Symbolic.MacawExt arch)
-extensionExec archCtx bak la (ExecutingAddressAction executeInsnAct) tgtOvs memVar archStruct baseExt stmt crucState = do
+extensionExec archCtx bak la insnAct tgtOvs memVar archStruct baseExt stmt crucState = do
+  let ExecutingAddressAction executeInsnAct = insnAct
   let sym = CB.backendGetSym bak
   case stmt of
     Symbolic.PtrAnd _w (CS.RegEntry _ x) (CS.RegEntry _ y) -> do


grease-cli/src/Grease/Cli.hs
```diff
@@ -452,6 +453,20 @@ simOpts = do
...
+            <> Opt.metavar "FILE"
+            <> Opt.help "Dump the addresses of executed instructions"
+      )
+  simDumpSectionMap <-
+    Opt.optional
+      ( Opt.strOption $
+          Opt.long "dump-sections"
+            <> Opt.metavar "FILE"
+            <> Opt.help "Dump a mapping from section indexes to addresses"
+      )
```
langston-barrett: Are these options that downstream projects might want to use in their CLIs? If so, might make sense to move these to the top level and export them, possibly wrapping them together in a struct with a single parser as is done with some other option groups.

grease-exe/src/Grease/Main.hs
```diff
@@ -961,6 +964,21 @@ macawRefineOnce la archCtx simOpts halloc macawCfgConfig memPtrTable setupHook a
                     )
                 ]
 
+addressCallBack ::
+  MM.MemWidth (MC.RegAddrWidth (MC.ArchReg arch)) =>
+  Proxy
+    arch ->
```
langston-barrett: Feel free to ignore, but this could go on one line

grease-exe/src/Grease/Main.hs
```diff
@@ -1229,32 +1250,40 @@ simulateMacawCfgs ::
...
-                      fromMaybe 0 $
-                        MML.loadOffset $
-                          mcLoadOptions macawCfgConfig
-                  }
-          pure (entry, result)
+  withMaybeFile (simDumpCoverage simOpts) WriteMode $ \covHandle -> do
+    withMaybeFile (simDumpSectionMap simOpts) WriteMode $ \secMapHandle -> do
+      let mem = mcMemory macawCfgConfig
+      let secsJson = Load.dumpSections mem
+      Maybe.maybe (pure ()) (\hand -> Text.IO.hPutStrLn hand (renderJSON secsJson)) secMapHandle
```
langston-barrett: Given the `Maybe.maybe` here, it seems like `withMaybeFile` is perhaps overly complex for what this code does. Perhaps:
```haskell
case simDumpSectionMap simOpts of
  Nothing -> pure ()
  Just path -> do
    let mem = mcMemory macawCfgConfig
    let secsJson = Load.dumpSections mem
    withFile path WriteMode $ \hand -> Text.IO.hPutStrLn hand (renderJSON secsJson)
```
This could also go above the `withMaybeFile` for `covHandle`, since it doesn't need access to `covHandle`.

grease-exe/src/Grease/Main.hs
```diff
@@ -1229,32 +1250,40 @@ simulateMacawCfgs ::
...
+                    let url = "https://galoisinc.github.io/grease/sexp-progs.html"
+                    userError la (PP.pretty err <> "\n" <> "For more information, see " <> url)
+              case checkMacawEntrypointCfgsSignatures archCtx entrypointCfgs of
+                -- See Note [Explicitly listed errors]
+                Left err@GME.BadArgs{} -> usrErr err
+                Left err@GME.BadRet{} -> usrErr err
+                Right ok -> pure ok
+            let
+              addressHandle :: ExecutingAddressAction arch
+              addressHandle = ExecutingAddressAction $ Maybe.maybe (\_ -> pure ()) (addressCallBack @arch Proxy) covHandle
```
langston-barrett: This thing is referred to as an "action" in `ExecutingAddressAction`, a "callback" in `addressCallBack`, and a "handle" in `addressHandle`. No biggie, but might be a bit more readable to standardize?

grease-exe/src/Grease/Main.hs
```diff
@@ -961,6 +964,21 @@ macawRefineOnce la archCtx simOpts halloc macawCfgConfig memPtrTable setupHook a
...
 
+addressCallBack ::
+  MM.MemWidth (MC.RegAddrWidth (MC.ArchReg arch)) =>
+  Proxy
+    arch ->
+  Handle ->
+  MM.MemSegmentOff (MC.RegAddrWidth (MC.ArchReg arch)) ->
+  IO ()
+addressCallBack _ hdl addr =
+  let js = renderJSON $ Load.memSegOffToSectionMemAddr addr
```
langston-barrett: Perhaps we should just export a function `MemSegmentOff -> JSON` from `Load`, rather than exporting `memSegOffToSectionMemAddr` + `SectionMemAddr`?

grease/src/Grease/Macaw/Load.hs
```diff
@@ -427,6 +433,61 @@ resolveCoreDumpEntrypointAddress la loadOpts mem binaryHeaderInfo symMap coreDum
...
+-- That is an offset from a relocatable region
+data SectionMemAddr = SectionMemAddr
+  { secInfoRegionIndex :: Int
+  , secInfoRegionOffset :: Word64
+  }
+  deriving (Generic, Show)
+
+instance ToJSON SectionMemAddr
+
+memSegOffToSectionMemAddr :: (MM.MemWidth w) => MM.MemSegmentOff w -> SectionMemAddr
```
langston-barrett: nit: extra parens

grease/src/Grease/Macaw/Load.hs
```diff
@@ -427,6 +433,61 @@ resolveCoreDumpEntrypointAddress la loadOpts mem binaryHeaderInfo symMap coreDum
...
+-- JSON
+newtype SectionDump = SectionDump [SectionInfo] deriving (Generic, Show)
+
+instance ToJSON SectionDump
+
+-- | Dumps the section map of memory that is a map from SectionIndex->memSegOff in a
+-- format that can be rendered as JSON. Each section index is mapped to a json representation of a
+-- 'MM.MemAddr' that contains the base region of the address along with the offset.
+dumpSections :: MM.MemWidth w => MM.Memory w -> SectionDump
+dumpSections mem =
```
langston-barrett, line 449: Similar question here - Perhaps instead of exporting `SectionDump` and `dumpSections`, which feel like implementation details, we could just export a function `Memory -> JSON` (or `Memory -> ByteString` or whatever)?

grease-cli/src/Grease/Cli.hs
```diff
@@ -452,6 +453,20 @@ simOpts = do
   simFsOpts <- fsOptsParser
   simInitPrecondOpts <- initPrecondOptsParser
   simBoundsOpts <- boundsOptsParser
+  simDumpCoverage <-
+    Opt.optional
+      ( Opt.strOption $
+          Opt.long "dump-coverage"
+            <> Opt.metavar "FILE"
+            <> Opt.help "Dump the addresses of executed instructions"
+      )
```
langston-barrett: Is there any point to using `--dump-coverage` without `--dump-sections`? If not, perhaps we should error (or even just warn) in that case?

grease/src/Grease/Macaw/Load.hs
```diff
@@ -427,6 +433,61 @@ resolveCoreDumpEntrypointAddress la loadOpts mem binaryHeaderInfo symMap coreDum
...
+        ( \(secIdx, memSegOff) ->
+            let
+              seg = memSegOffToSectionMemAddr memSegOff
+             in
+              SectionInfo
+                { secInfoSectionIndex = fromIntegral secIdx
+                , secInfoSectionAddr =
+                    seg
+                }
+        )
```
langston-barrett, line 463: Perhaps:
```suggestion
        ( \(secIdx, memSegOff) ->
            let seg = memSegOffToSectionMemAddr memSegOff
             in SectionInfo
                { secInfoSectionIndex = fromIntegral secIdx
                , secInfoSectionAddr = seg
                }
        )
```

grease/src/Grease/Macaw/Load.hs
```diff
@@ -427,6 +433,61 @@ resolveCoreDumpEntrypointAddress la loadOpts mem binaryHeaderInfo symMap coreDum
...
+        (Map.toList secs)
+   in
+    SectionDump sinfo
```
langston-barrett:
```suggestion
   in SectionDump sinfo
```

grease/src/Grease/Macaw/Load.hs
```diff
@@ -427,6 +433,61 @@ resolveCoreDumpEntrypointAddress la loadOpts mem binaryHeaderInfo symMap coreDum
...
+
+-- | Dumps the section map of memory that is a map from SectionIndex->memSegOff in a
+-- format that can be rendered as JSON. Each section index is mapped to a json representation of a
+-- 'MM.MemAddr' that contains the base region of the address along with the offset.
+dumpSections :: MM.MemWidth w => MM.Memory w -> SectionDump
+dumpSections mem =
+  let
+    secs = MM.memSectionIndexMap mem
+    sinfo =
+      fmap
```
langston-barrett: I find it a bit clearer to use less-polymorphic functions when possible:
```suggestion
      map
```

grease/src/Grease/Macaw/SimulatorHooks.hs
```diff
@@ -145,6 +147,9 @@ updateStruct (rstRegs Ctx.:> currReg) (rstVals Ctx.:> currVal) regToNewVal =
             CS.RV val
    in updateStruct rstRegs rstVals regToNewVal Ctx.:> newVal
 
+-- | An IO action that observes the starting address of an instruction before it executes.
+newtype ExecutingAddressAction arch = ExecutingAddressAction (MM.MemSegmentOff (MC.RegAddrWidth (MC.ArchReg arch)) -> IO ())
```
langston-barrett:
```suggestion
newtype ExecutingAddressAction arch
  = ExecutingAddressAction (MM.MemSegmentOff (MC.RegAddrWidth (MC.ArchReg arch)) -> IO ())
```

langston-barrett (commented):
Let me see if I understand. These new flags are generally expected to be used together, and the point is to export a list of addresses of instrucitons that GREASE executes so that such a list could be imported into and displayed in a UI like Ghidra.

The two flags added dump to different files at different times:

1. `--dump-sections` walks over the Macaw `Memory` and dumps a map `SectionIndex -> BaseAddress`
2. `--dump-coverage` fires on each instruction executed and writes a pair `(SectionIndex, Offset)`.

Clients are expected to combine this information by looking up the `SectionIndex` from the instruction in the section map, and adding together the `BaseAddress+Offset` to get the actual address of the instruction (And this is an address in the ELF file? Or an address in GREASE's virtual memory, so clients will also need to subtract GREASE's load base address?).

If the above description is accurate, then could we perhaps simplify by directly dumping addresses in the format that clients want to consume, i.e., doing the lookup `SectionIndex -> BaseAddress` entirely within GREASE?

(By the way, all of the above sort of information would be a good kind of thing to include in a PR description.)

langston-barrett:
I see. That's reasonably complicated, it would be nice to document it somewhere (perhaps in a `Note`).

grease/src/Grease/Macaw/Load.hs
```diff
@@ -427,6 +431,40 @@ resolveCoreDumpEntrypointAddress la loadOpts mem binaryHeaderInfo symMap coreDum
             prStatusPcWithOffset
       pure nearestEntrypointAddr
 
+memSegOffToSectionMemAddr :: MM.MemWidth w => MM.MemSegmentOff w -> Aeson.Value
+memSegOffToSectionMemAddr memSegOff =
```
langston-barrett: I think there used to be a type `SectionMemAddr` which is where the name of this function comes from, perhaps now rename to just `memSegOffToJson`?

grease-exe/src/Grease/Main.hs
```diff
@@ -1229,32 +1249,39 @@ simulateMacawCfgs ::
...
-                  , batchLoadOffset =
-                      fromMaybe 0 $
-                        MML.loadOffset $
-                          mcLoadOptions macawCfgConfig
-                  }
-          pure (entry, result)
+  withMaybeFile (simDumpCoverage simOpts) WriteMode $ \mbCovHandle -> do
+    let mem = mcMemory macawCfgConfig
+    let secsJson = Load.dumpSections mem
+    Maybe.maybe (pure ()) (\covHandle -> Text.IO.hPutStrLn covHandle (renderJSON secsJson)) mbCovHandle
```
langston-barrett: This is a minor style thing so feel free to ignore, but I would generally write this with an explicit `case` on `mbCovHandle`.

grease-cli/src/Grease/Cli.hs
```diff
@@ -345,6 +347,18 @@ symStdinParser =
...
+simDumpCoverageParser :: Opt.Parser (Maybe FilePath)
+simDumpCoverageParser =
+  Opt.optional
+    ( Opt.strOption $
+        Opt.long "dump-coverage"
+          <> Opt.metavar "FILE"
+          <> Opt.help
+            ( "Dump the addresses of executed instructions. These addresses are relative to a section"
+                ++ "index which is the first JSON array appended to the file"
+            )
```
langston-barrett, line 359: Would this be a valid [JSON lines](https://jsonlines.org/) file at that point? Perhaps worth saying so, if so.

I like the simplification of a single file, by the way!

grease/src/Grease/Macaw/Load.hs
```diff
@@ -13,11 +13,15 @@ module Grease.Macaw.Load (
   LoadedProgram (..),
   LoadError (..),
   load,
+  dumpSections,
+  memSegOffToSectionMemAddr,
```
langston-barrett: It's not super clear to me that these belong in this module... but I'm not really sure where else we might put them 🤷

grease/src/Grease/Macaw/SimulatorHooks.hs
```diff
@@ -156,12 +162,14 @@ extensionExec ::
...
   GreaseLogAction ->
+  -- | Action for logging executed instructions
+  ExecutingAddressAction arch ->
   AddressOverrides arch ->
   CS.GlobalVar CLM.Mem ->
   CS.GlobalVar (MC.ArchRegStruct arch) ->
   CS.ExtensionImpl p sym (Symbolic.MacawExt arch) ->
   Symbolic.MacawEvalStmtFunc (C.StmtExtension (Symbolic.MacawExt arch)) p sym (Symbolic.MacawExt arch)
-extensionExec archCtx bak la tgtOvs memVar archStruct baseExt stmt crucState = do
+extensionExec archCtx bak la (ExecutingAddressAction executeInsnAct) tgtOvs memVar archStruct baseExt stmt crucState = do
```
langston-barrett: Here's a little trick for avoiding super long parameter lines:
```suggestion
extensionExec archCtx bak la insnAct tgtOvs memVar archStruct baseExt stmt crucState = do
  let ExecutingAddressAction executeInsnAct = insnAct
```

PR #457, opened 2025-11-14
Provide a Ninja-based runner for linters

langston-barrett:
Add a simple Python script to generate a [Ninja](https://ninja-build.org/) build configuration. The build configuration runs our various linters in the same configuration as CI. This makes linting both fully incremental and fully parallel, and suitable for running on every save in an editor or in a pre-commit hook.

Also:

- Find and fix some places where CI was not linting, but should have been
- Move an out-of-place section on `git blame` in the docs

TODO:

- [x] Describe how to set up a pre-commit hook that runs this script

### DIFF
--- a/scripts/lint/lint.py
+++ b/scripts/lint/lint.py
@@ -87,6 +87,10 @@ def txt(path: str):
 
 
 def hs():
+    cb = ls_files("*.cabal")
+    for path in cb:
+        txt(path)
+
     hs = ls_files("*.hs")
     for path in hs:
         if "elf-edit-core-dump" in path:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -36,6 +36,7 @@ jobs:
       - name: Check for trailing whitespace
         run: >
           python3 scripts/lint/whitespace.py
+          $(git ls-files '*.cabal')
           $(git ls-files '*.hs')
           $(git ls-files '*.md')
           $(git ls-files '*.py')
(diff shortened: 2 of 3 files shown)


doc/function-calls.md
```diff
@@ -6,8 +6,9 @@ GREASE's algorithm for handling function calls is as follows:
   - If `--error-symbolic-fun-calls` was provided, an error is generated.
   - Otherwise, the call is skipped.
 - Otherwise, the virtual address is resolved to an address in the binary. If
-  this fails, the call is skipped. <!-- TODO(#137): Make this an error by
-  default -->
+  this fails:
+  - If `--skip-invalid-call-addrs` was provided, the call is skipped.
+  - Otherwise, an error is generated.
```
langston-barrett, line 11: Our TODO numbering system and stale TODO checker really pays dividends, IMO 😁

scripts/lint/lint.py
```diff
@@ -0,0 +1,142 @@
+#!/usr/bin/env python
```
2over12, line 1: I would say overall i find this script a bit hard to read but get the gist, bit hard to tell what the text manipulation is doing etc but probably fine since this is best effort

2over12 (approved):

scripts/lint/lint.py
langston-barrett: Anything in particular I could do to to help? Add more comments, an quick primer on Ninja syntax, logging?

langston-barrett: Let's discuss further on #459!

Issue #470, opened 2025-11-20
Print errors in constructing dwarf shapes


2over12:
We should make `fromDwarfInfo` collect a log of parsing errors, and individual shapes return an Either so we know why a shape failed to parse

2over12:
Indeed, another instance of coarse grained parsing of linked issues

PR #471, opened 2025-11-20
Construct dwarf shapes through qualified and typedefs

2over12:
Change extract type to extract the relevant type rather than a typedef or qualified type. We should also fix https://github.com/GaloisInc/grease/issues/470 so we can write a regression test for this (e.g. then we could check that there is no DWARF parsing error in the log)


grease/src/Grease/Macaw/Dwarf.hs
langston-barrett, line 228: The new line is quite long, IMO this would be more readable and less repetitive with a single equation with a `\case` over the `TypeApp` on the RHS.

PR #487, opened 2025-12-19
DWARF: Add debug logging for dwarf

2over12:
Adds debug output, decided to pass a logging action down up to individual shape parsing. Closes #470

PR #488, opened 2025-12-19
DWARF: Add parsing shapes from DWARF arrays

### LINKED ISSUE #263: DWARF Shapes: Support Arrays
Building arrays requires determining the number of elements. In dwarf this number can either be declared through DW_AT_upper_bound and DW_AT_lower_bound which are DWARF expressions or DW_AT_count. Neither of these attributes are easy to get with the current state of dwarf parsing in macaw. Downstream issues exist to address these 

- https://github.com/GaloisInc/macaw/issues/507
- https://github.com/GaloisInc/macaw/issues/508

2over12:
Closes #263. Currently only supports count declared arrays. I suspect upper bound is used quite rarely

Blocked on #487 and https://github.com/GaloisInc/macaw/pull/553

### DIFF
--- a/grease/src/Grease/Macaw/Dwarf.hs
+++ b/grease/src/Grease/Macaw/Dwarf.hs
@@ -119,7 +119,7 @@ loadDwarfPreconditions gla targetAddr memory tyUnrollBound argNames initShapes e
         hoistMaybe $ either (const Nothing) Just repl
     )
 
-extractType :: Subprogram -> MDwarf.TypeRef -> Either String MDwarf.TypeApp
+extractType :: Subprogram -> MDwarf.TypeRef -> Either DwarfDiagnostic.DwarfShapeParsingError MDwarf.TypeApp
 extractType sprog ref =
   extractTypeInternal ref
     >>= ( \case
@@ -128,12 +128,14 @@ extractType sprog ref =
             x -> Right x
         )
  where
-  extractTypeInternal :: MDwarf.TypeRef -> Either String MDwarf.TypeApp
+  extractTypeInternal :: MDwarf.TypeRef -> Either DwarfDiagnostic.DwarfShapeParsingError MDwarf.TypeApp
   extractTypeInternal vrTy =
     let mp = MDwarf.subTypeMap sprog
      in case Map.lookup vrTy mp of
-          Nothing -> Left $ "No type for typeref: " ++ show vrTy
-          Just (a, _) -> a
+          Nothing -> Left $ DwarfDiagnostic.UnexpectedDWARFForm $ "No type for typeref: " ++ show vrTy
+          Just (a, _) -> case a of
+            Left x -> Left $ DwarfDiagnostic.UnexpectedDWARFForm x
+            Right x -> Right x
 
 constructPtrTarget ::
   CLM.HasPtrWidth w =>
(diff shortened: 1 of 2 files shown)


grease/src/Grease/Macaw/Dwarf.hs
```diff
@@ -164,13 +164,27 @@ constructPtrTarget tyUnrollBound sprog visitCount tyApp =
   shapeOfTyApp :: CLM.HasPtrWidth w => MDwarf.TypeRef -> Either String (Seq.Seq (MemShape w NoTag))
   shapeOfTyApp x = shapeSeq =<< extractType sprog =<< pure x
 
+  numOfRange :: MDwarf.SubrangeBounds -> Either String Int
+  numOfRange =
+    \case
+      MDwarf.SubrangeCount val -> Right $ fromIntegral val
+      _ -> Left "Array parser currently only supports DW_AT_count and does not support DW_AT_upperbound"
```
langston-barrett: Not sure I really understood your comment about this on the last PR. Still feels to me like this would be easy to convert to a datatype with a `Pretty` instance.

grease/src/Grease/Macaw/Dwarf.hs
```diff
@@ -164,13 +164,27 @@ constructPtrTarget tyUnrollBound sprog visitCount tyApp =
...
-  shapeSeq (MDwarf.ArrayType _elTy _ub) = Left "Array types currently unsupported due to GREASE's DWARF parser, see https://github.com/GaloisInc/grease/issues/263"
+  shapeSeq (MDwarf.ArrayType elTy ub) =
+    do
+      let onlySubrange = if List.length ub == 1 then Maybe.listToMaybe ub else Nothing
+      ubRange <- case onlySubrange of
+        Just x -> Right x
+        Nothing -> Left "Expected only one DW_AT_subrange_type for DW_TAG_array_type"
+      num <- numOfRange (MDwarf.subrangeUpperBound ubRange)
+      tyShape <- shapeOfTyApp elTy
+      pure $ join $ Seq.replicate num tyShape
```
langston-barrett, line 189: What monad is this `join` in? My mental type-checker is off here, I wouldn't expect it to be needed at all.

2over12: this is in the monad `Seq` to flatten the repetition, couldnt find a more straightforward thing like concat

grease/src/Grease/Macaw/Dwarf.hs
2over12: the original comment was in reference to this api in DWARF https://galoisinc.github.io/macaw/Data-Macaw-Dwarf.html#t:AbsType where we bubble up Either Strings as errors from the internal dwarf library which happens in a couple places, so the point being we keep it as Either String a, but ive converted to wrapping these in a generic error

grease/src/Grease/Macaw/Dwarf.hs
langston-barrett: You might be looking for `mconcat`?

grease/src/Grease/Macaw/Dwarf.hs
langston-barrett: I see, looks great thanks!

grease/src/Grease/Macaw/Dwarf.hs
RyanGlScott: Note that `mconcat` requires the outermost type to be a list, so while `mconcat` would work for `[Seq a]`, it would not work for `Seq (Seq a)` (which I think is the type involved in the code here). An even more general version of `mconcat` is [`fold`](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Data-Foldable.html#v:fold) from the `Foldable` class, which should work for `Seq (Seq a)`.

PR #492, opened 2025-12-28
Experiment with assertion language

2over12:
This PR experiments with adding separation logic assertions to GREASE. The initial design missed some finicky issues that I'm trying to work through here. No action required, just tracking some design/implementation decisions, troubles, and progress. The goal here is a datastructure that both is easy to parse into and easy to operationlize in precondition setup.

1. Parsing assertions independently from an initialized symbolic memory and arguments 
2. Handling non base values (e.g. pointers)

### Parsing Assertions Independently from an Initialized set of Arguments
I think it's ideal if we can parse an assertion without initializing blocks before hand or setting up an initial rid/arg etc. We also (probably) want to reuse What4 expressions to make it easier to expose new syntax in assertions as needed (we can just add the former and bind it to a What4 expression rather than creating some internal syntax, then translating to what4 down the road). To do this we need to support substituting blocks for blockids and argument values for symbolic expressions once they are created.

This is the purpose of the bindings type:
```haskell
data Bindings sym argTys = Bindings
  { blocks :: Map.Map (WI.SymNat sym) (WI.SymNat sym)
  , args :: Ctx.Assignment (CS.RegValue' sym) argTys
  }
```
Which annotates expressions with an  assignment of RegValues. The idea would be that an expressions would be parsed with an arg binding that is entirely fresh variables and then a substitution is applied by setting args == some new assignment, where elements are pairwise set equal e.g. for base types just assume(symexpr1 == symexpr2). The idea with blocks is to do similar and map a SymNat as it exists to some other SymNat, but i think this current representation is not ideal because we are relying on subtle invariants of what populates that map (e.g. the value of the map should always be a constant sym nat so that we can setup equivalence classes that get bound to a fresh doMalloc).

An additional observation is Bindings should probably be moved to a top-level structure since we dont need separate bindings on each expression we can just parse in a context of a single assignment and maybe a monad of state to collect block IDs as we go or some such, so the result would be:
```haskell
data BoundAssertion sym wptr argTys s = {
assertBinding :: Bindings sym argTys,
assert :: Assertion sym wptr argTys s 
}
```

An additional observation is that bindings on abstract labels appear useful but may not be at least for now. It would seem being able to write down `#v == (ite #b $l1 $l2)` would be useful for handling conditional pointers, however, we dont have a plan for how to operationalize `uninit_cell(#v, 2)` because we cannot initialize a muxed cell so we will reject this, therefore we cannot do anything with a muxed label at the moment so perhaps we can leave off block substitutions and only support literal labels in predicates.

### Handling Pointer Values
The initial design also glossed over some complexity with forming, destructuring and storing pointers. We would like to talk about the label and offset of an argument (e.g. rdi) but we'd also like to form new pointer values e.g. make `#v == rdi+5` e.g. `#v == ($lx, 5)`. We need to explicitly create a LLVMPointer internally which cannot be part of symexprs etc. In particular we will need to form the Crucible type of pointer to store it in init_cell. One option would be to have syntax init_cell_ptr and init_cell which either take bvs or pointers and are then operationalized to create the correct types internally. Instead to unify these into a single predicate the idea is to create an opaque type `Ptr` that can be constructed or destructed and then converted to a memory value:
```
data MemVal sym wptr argTys s
  = MemPtr (PtrVar argTys wptr s)
  | forall w. MemBv {memBvWidth :: Param.NatRepr w, memBvExpr :: BindableExpr sym (WI.BaseBVType w) argTys}
```
A PtrVar is a reference to an LLVMPointer value stored in the environment (either a bound existential or argument).

Which we can pass to the LLVM memmodel to store.

`Ptr`'s can only be made via the predicate `is_ptr` (by creating an existential upon which the predicate is applied) or by already being in the environment (e.g. `typeof(rdi) == Ptr`). 

`is_ptr(Ptr, Label, Bv ptr_width) // (the bound pointer, the abstract label, the offset)` can both destruct and construct pointers:
```
is_ptr(rdi, $l1, #o) // destructs rdi, operationalized as taking the base of base(rdi) == block_id($l1) /\ offset(rdi) == #o
is_ptr(#x, $l2, #o2) // operationalized as taking base(#x) == block_id($l2) /\ offset(#x) == #o2 but we need to initialize a fresh ptr for the existential (either apriori by knowing which existentials are pointers or on the fly)
```

I think the correct procedure here is probably to traverse the Assertion and create pointers prior to initialization.

Then in our binding map for rdi  and pointer existential we have fresh base/offset pointers that we can pass to InitCell to do this in syntax we have:
```
init_cell($l2, #o2, mem_ptr(#x))
```
Which is operationalized by looking up #x in the map of pointers and constructing that pointer as a Crucible RegValue.

### Design Fixes/Notes from Review
1. We should lift bindings to the top level
2. I think we should unify and generalize bound existential pointers and arguments via a MapF from Nonce to C.RegValue where we populate ptr arguments  from existential with fresh nonces, and  the Nonces for arguments could have a map like Nonce->Ctx.Index argty potentially

### DIFF
--- a/grease-exe/src/Grease/Main.hs
+++ b/grease-exe/src/Grease/Main.hs
@@ -143,6 +143,7 @@ import Grease.Panic (Grease, panic)
 import Grease.Pretty (prettyPtrFnMap)
 import Grease.Profiler.Feature (greaseProfilerFeature)
 import Grease.Refine
+import Grease.Refine qualified as Refine
 import Grease.Requirement
 import Grease.Setup
 import Grease.Setup qualified as Setup
@@ -913,14 +914,15 @@ withMaybeFile (Just pth) imd f = withFile pth imd (f . Just)
 
 data PreconditionEnvironment sym bak ext argTys where
   PreconditionEnvironment ::
-    PP.Pretty fmted =>
-    precond ->
-    (precond -> fmted) ->
-    [RefineHeuristic sym bak ext argTys precond] ->
+    (PP.Pretty fmted, pty ~ precond ext tag argTys) =>
+    pty ->
+    (pty -> fmted) ->
+    [RefineHeuristic sym bak ext argTys pty] ->
+    Refine.SetupFn precond ext tag argTys ->
     PreconditionEnvironment sym bak ext argTys
 
 macawInitPreconditionEnv ::
-  forall sym bak arch scope st fm precond solver.
+  forall sym bak arch scope st fm solver.
   ( CB.IsSymBackend sym bak
   , sym ~ W4.ExprBuilder scope st (W4.Flags fm)
   , bak ~ CB.OnlineBackend solver scope st (W4.Flags fm)
@@ -958,7 +960,7 @@ macawInitPreconditionEnv la _ bak archCtx simOpts macawCfgConfig argNames mbCfgA
     let opts = simInitPrecondOpts simOpts
     initArgShapes <- getMacawInitArgShapes la bak archCtx opts macawCfgConfig argNames mbCfgAddr
     let fmt = ShapePrint.PrintableShapes addrWidth argNames
-    pure $ PreconditionEnvironment initArgShapes fmt heuristics
+    pure $ PreconditionEnvironment initArgShapes fmt heuristics Setup.setup
 
 simulateMacawCfg ::
   forall sym bak arch solver scope st fm.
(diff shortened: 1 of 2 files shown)


grease-exe/src/Grease/Main.hs
```diff
@@ -908,6 +912,56 @@ withMaybeFile :: Maybe FilePath -> IOMode -> (Maybe Handle -> IO a) -> IO a
...
 withMaybeFile (Just pth) imd f = withFile pth imd (f . Just)
 
+data PreconditionEnvironment sym bak ext argTys where
+  PreconditionEnvironment ::
+    (PP.Pretty fmted, pty ~ precond ext tag argTys) =>
+    pty ->
+    (pty -> fmted) ->
+    [RefineHeuristic sym bak ext argTys pty] ->
+    Refine.SetupFn precond ext tag argTys ->
+    PreconditionEnvironment sym bak ext argTys
```
langston-barrett, line 922: Offline we discussed the idea of having a single `data` that does the existential quantification over `precond` wrap another struct that has accessors for these fields.

grease/src/Grease/Assertion/Assrt.hs
```diff
@@ -0,0 +1,80 @@
...
+
+-- current problem we have a bit of a weird situation where to interact with the mem model we have non
+-- symbolic stuff we'd like to do. That is we want to form a ptr(x,a) but this can only be a top, level thing
+-- and is only consumeable in the store position...
+-- imagine if we have a = ptr(l,o)... what does that mean... my view maybe.. is we could in theory do
+-- init_cell_ptr() but then this is restricting ptrs, perhaps we need to unify exprs and ptr exprs
+-- the trouble is making a bindable regValue is... annoying
+data MemVal sym wptr argTys s
+  = MemPtr (PtrVar argTys wptr s)
+  | forall w. MemBv {memBvWidth :: Param.NatRepr w, memBvExpr :: WI.SymBV sym w}
```
langston-barrett, line 54: FWIW `SymBV`s store their width: https://hackage-content.haskell.org/package/what4-1.7.3/docs/What4-Expr-Builder.html#v:bvWidth

Issue #476, opened 2025-12-02
Perform async reporting of ExecResults during DFS mode


2over12:
DFS mode often wont terminate or will timeout even though we found an interesting bug on one path. We only get the initial logging output related to these terminal states that occurs in `execAndRefine`. If we hit a timeout the combiner will drop the rest of the results, indeed if on some other path we find some cant refine situation or something we will drop a bug in this case as well. This situation where DFS results are dependent on other paths feels a bit wrong to me. If you want to do OX proving honestly I think you should just run SSE? If you are running DFS you kinda want the independent state results in my opinion. 

One option here would be maybe to have more explicit async logging of each of these states (e.g. with concretized info as these go by) or fully bubble up a set of results

langston-barrett:
Ah, I'll bet that the problem is that we are pattern-matching on constructors of `AbortedResult` here:

https://github.com/GaloisInc/grease/blob/eb3836428310e0e00e26de48b712aecd276d8604/grease/src/Grease/Refine.hs#L500

This case is likely matched by the DFS path, but missed by the SSE path where the result is probably an `AbortedBranch`. We should fix this inconsistency.

PR #494, opened 2025-12-30
Fix discrepancy between DFS / SSE mode results

### LINKED ISSUE #493: Discrepancy between DFS + SSE
Split off from #476, which is about *changing* how we report results in DFS mode. This is just about a discrepancy between DFS / SSE mode.

> Here is a test-case that demonstrates the issue.
> ```
> (declare @abort () Unit)
> 
> (defun @test ((p (Ptr 64))) Unit
>   (start start:
>     (let off (ptr-offset 64 p))
>     (branch (equal? off (bv 64 0)) if: else:))
>   (defblock if:
>     (funcall @abort)
>     (return ()))
>   (defblock else:
>     (jump else:)))
> 
> ```
> ```
> cabal run exe:grease -- test.llvm.cbl --path-strategy sse
> 
> No entry points specified, analyzing all known functions.
> Finished analyzing 'test'. Possible bug(s):
> 
> At test.llvm.cbl:17:5:
> Resource exhausted: reached maximum number of loop iterations (32)
> ...
> 
> At test.llvm.cbl:14:5:
> Call to abort
> ...
> ```
> vs.
> ```
> cabal run exe:grease -- test.llvm.cbl --path-strategy dfs
> 
> No entry points specified, analyzing all known functions.
> Path ended at test.llvm.cbl:14:5 with result possible bug
> Path ended at test.llvm.cbl:17:5 with result resource exhausted
> Finished analyzing 'test'. Resource exhausted: reached maximum number of loop iterations (32)
> ```
> The DFS+SSE result should be the same, and/or the DFS mode should report per-path results individually. 

 _Originally posted by @langston-barrett in [#476](https://github.com/GaloisInc/grease/issues/476#issuecomment-3700540308)_

langston-barrett:
Fixes #493.

### DIFF
--- a/grease-exe/tests/refine/neg/combine-results.llvm.cbl
+++ b/grease-exe/tests/refine/neg/combine-results.llvm.cbl
@@ -4,7 +4,7 @@
 ; a possible bug and the other cannot be refined (in this case, due to the loop
 ; bound).
 ;
-; Regression test for #476.
+; Regression test for #493.
 
 (declare @abort () Unit)
 
@@ -21,11 +21,10 @@
 ;; flags {"--symbol", "test"}
 ;; flags {"--path-strategy", "sse"}
 ;; go(prog)
-;; could_not_infer()
+;; check_not "Possible bug(s):"
+;; check 'reached maximum number of loop iterations'
 
 ;; flags {"--symbol", "test"}
 ;; flags {"--path-strategy", "dfs"}
 ;; go(prog)
-; TODO(#476):
-; could_not_infer()
 ;; check 'reached maximum number of loop iterations'
--- a/grease/src/Grease/Refine.hs
+++ b/grease/src/Grease/Refine.hs
@@ -82,7 +82,7 @@ module Grease.Refine (
   findPredAnnotations,
 ) where
 
-import Control.Applicative (pure)
+import Control.Applicative (pure, (<|>))
 import Control.Exception.Safe qualified as X
 import Control.Lens ((^.))
 import Control.Monad qualified as Monad
@@ -480,6 +480,10 @@ processExecResult =
       Just (Exhausted msg)
     CS.AbortedResult _ (CS.AbortedExec (CB.AssertionFailure (C.SimError _loc (C.Unsupported _cs feat))) _gp) ->
       Just (Unsupported feat)
+    CS.AbortedResult ctx (CS.AbortedBranch _loc _pred r1 r2) ->
+      -- Taking the first 'CantRefine' is consistent with 'consumer'
+      processExecResult (CS.AbortedResult ctx r1)
+        <|> processExecResult (CS.AbortedResult ctx r2)
     _ -> Nothing
 
 execAndRefine ::


grease-exe/tests/refine/neg/combine-results.llvm.cbl
```diff
@@ -0,0 +1,29 @@
...
+  (defblock if:
+    (funcall @abort)
+    (return ()))
+  (defblock else:
+    (jump else:)))
+
+;; flags {"--symbol", "test"}
+;; flags {"--path-strategy", "sse"}
+;; go(prog)
+;; check 'reached maximum number of loop iterations'
```
2over12, line 25: Sorry am I misreading this test, per: https://github.com/GaloisInc/grease/issues/476#issuecomment-3700540308 they both already outputted the maximum number of loop iterations error just in different ways?

grease/src/Grease/Refine.hs
```diff
@@ -463,6 +463,29 @@ proveAndRefine bak execResult la bbMap refineData goals = do
...
+      Just (Exit (Just 0))
+    CS.AbortedResult _ (CS.AbortedExit (Exit.ExitFailure code) _) ->
+      Just (Exit (Just code))
+    CS.AbortedResult _ (CS.AbortedExec (CB.EarlyExit _loc) _gp) ->
+      Just (Exit Nothing)
+    CS.AbortedResult _ (CS.AbortedExec (CB.AssertionFailure (C.SimError _loc (C.ResourceExhausted msg))) _gp) ->
+      Just (Exhausted msg)
+    CS.AbortedResult _ (CS.AbortedExec (CB.AssertionFailure (C.SimError _loc (C.Unsupported _cs feat))) _gp) ->
+      Just (Unsupported feat)
+    CS.AbortedResult ctx (CS.AbortedBranch _loc _pred r1 r2) ->
```
2over12, line 483: If im reading this right the goal here is to take SSE's abort branch in this case and perform the same thing DFS/ the proof combiner would do and report the first cant refine, is that correct?

grease-exe/tests/refine/neg/combine-results.llvm.cbl
langston-barrett: Oh whoops, no you're right 😅 Will add a `check_not` or something so that this test would fail with the bug.

grease/src/Grease/Refine.hs
langston-barrett: Yep!

grease-exe/tests/refine/neg/combine-results.llvm.cbl
```diff
@@ -0,0 +1,31 @@
...
+    (return ()))
+  (defblock else:
+    (jump else:)))
+
+;; flags {"--symbol", "test"}
+;; flags {"--path-strategy", "sse"}
+;; go(prog)
+;; check_not "Possible bug(s):"
+;; check 'reached maximum number of loop iterations'
+;; check_not "Possible bug(s):"
```
2over12: should this be after the go(prog) doesnt this check the sse result twice?

langston-barrett: The SSE result was the one that reported `Possible bug`. This line appears twice because we want to make sure that this doesn't appear before nor after `reached maximum...`, although now that I think about it just checking first is sufficient.

2over12: i thought the semantics of check_not was that it was a global check anyways? (i might be misremembering)

2over12: either way im set

langston-barrett: It checks the remainder of the stream from where you are at the moment.

2over12: yeah i guess my point is it never makes sense to call check not twice afaik? because you already know after that point that there is nothing there, otherwise you need something like check_not_until

langston-barrett: Right, that's why I removed the second check 😄

PR #495, opened 2026-01-02
cli: Add `--skip` option to skip function calls

### LINKED ISSUE #482: cli: `--skip` flag
It would be nice to support a flag to skip execution of certain functions. This would likely be pretty easy.

langston-barrett:
Fixes #482.

Needs:

- [x] Implementation for binaries
  - [x] Test skipping a defined function
- [x] Implementation for LLVM bitcode programs
  - [x] Test skipping a `@define`d function
  - [x] Test skipping a `@declare`d function with a user-override
  - [x] Test skipping a `@declare`d function with a built-in override
- [x] Implementation for LLVM S-expression programs
  - [x] Test skipping a `@define`d function
  - [x] Test skipping a `@declare`d function with a user-override
  - [x] Test skipping a `@declare`d function with a built-in override
- [x] Implementation for Macaw S-expression programs
  - [x] Test skipping a `@define`d function
  - [x] Test skipping a `@declare`d function with a user-override
  - [x] Test skipping a `@declare`d function with a built-in override
- [x] Docs for new test-suite directory
- [x] Docs for flag in function call reference

### DIFF
--- a/grease-exe/tests/feat/skip/skip-bogus.llvm.cbl
+++ b/grease-exe/tests/feat/skip/skip-bogus.llvm.cbl
@@ -1,0 +1,12 @@
+; Copyright (c) Galois, Inc. 2025
+
+; Test --skip with a bogus function name in LLVM S-expression files.
+
+(defun @test () Unit
+  (start start:
+    (return ())))
+
+;; flags {"--skip", "bogus"}
+;; flags {"--symbol", "test"}
+;; go(prog)
+;; ok()
--- a/grease-exe/tests/feat/skip/skip-bogus.x64.cbl
+++ b/grease-exe/tests/feat/skip/skip-bogus.x64.cbl
@@ -1,0 +1,12 @@
+; Copyright (c) Galois, Inc. 2025
+
+; Test --skip with a bogus function name in Macaw S-expression files.
+
+(defun @test ((regs X86Regs)) X86Regs
+  (start start:
+    (return regs)))
+
+;; flags {"--skip", "bogus"}
+;; flags {"--symbol", "test"}
+;; go(prog)
+;; ok()


grease/src/Grease/Macaw/ResolveCall.hs
```diff
@@ -334,37 +353,40 @@ lookupFunctionHandleResult bak la halloc arch memory symMap pltStubs dynFunMap f
...
+                      , st
+                      )
+                  -- ...otherwise, skip the PLT call entirely.
+                  Nothing ->
+                    -- TODO(#182): Option to make this an error
+                    pure
+                      ( SkippedFunctionCall $
+                          PltNoOverride funcAddrOff pltStubName
+                      , st
+                      )
```
langston-barrett, line 389: Note for reviewers: this is unchanged, just indented.

grease-exe/tests/feat/skip/skip-override.x64.cbl
```diff
@@ -0,0 +1,23 @@
...
+;; go(prog)
+;; must_fail()
+
+;; flags {"--skip", "bug"}
+;; flags {"--overrides", "tests/feat/skip/bug.aux.x64.cbl"}
+;; flags {"--symbol", "test"}
+;; go(prog)
+; TODO:
+; ok()
+;; must_fail()
```
langston-barrett: Not really sure what I'm doing wrong here... This yields
```
:1:0: error: in _start
Global symbol not allocated
Details:
  Global symbol "bug" has no associated allocation
```
I'm a bit tempted to just push this off to a follow-up issue, especially considering it seems like an obscure case (if you want to skip a function you're overriding, you can simply not provide the override in the first place, this test exists mostly to establish the precedence between these two mechanisms).

langston-barrett: Ah, the problem there was that I forgot to pass `--overrides`. That's a pretty poor error message... https://github.com/GaloisInc/grease/issues/497

langston-barrett: I can see the skip override being created and bound... and it doesn't seem to matter whether it is bound [before the first override](https://github.com/GaloisInc/grease/blob/a6210f3f86a98acb01ab66ae06091748b080e569/grease/src/Grease/Macaw/SetupHook.hs#L204) or after [the last one](https://github.com/GaloisInc/grease/blob/a6210f3f86a98acb01ab66ae06091748b080e569/grease/src/Grease/Macaw/SetupHook.hs#L211), or both...

langston-barrett: [`CS.bindFnHandle`](https://github.com/GaloisInc/crucible/blob/1c98666066d289404c795f125e5b1804d8f7dc1a/crucible/src/Lang/Crucible/Simulator/OverrideSim.hs#L256) eventually boils down to a `MapF.insert`, so the last binding should win... So why doesn't it?

langston-barrett: Fixed by also processing forward declarations.

2over12 (approved):

PR #503, opened 2026-01-07
tests: Add some tests for AND of pointers

langston-barrett:


2over12 (approved):
LGTM is there some background on why we are adding pointer-and forms that im missing? Is this more of a completeness thing

PR #505, opened 2026-01-08
DWARF defaults test case

### LINKED ISSUE #504: Test case for "default shape for failed pointers and field members"
#502 landed without a test case, but it could use one.

2over12:
Fixes #504


grease-exe/tests/sanity/pass/dwarf-default-for-field-members-and-pointers/test.c
```diff
@@ -0,0 +1,22 @@
...
+  char z;
+} y;
+
+struct bar {
+  int (*somefn)(int);
+  int x;
+  y *blah;
+};
+
+int foo(int x, int *y, struct bar *z) { return x + *y + z->x; }
```
langston-barrett: Why not just

```suggestion
int test(int x, int *y, struct bar *z) { return z->x; }
```

grease-exe/tests/sanity/pass/dwarf-default-for-field-members-and-pointers/test.c
```diff
@@ -0,0 +1,22 @@
...
+
+// all: flags {"--symbol", "foo", "--use-debug-info-types", "--no-heuristics"}
```
langston-barrett: In most of our tests, the main function is named `test`:

```suggestion
// all: flags {"--symbol", "test", "--use-debug-info-types", "--no-heuristics"}
```

langston-barrett (approved):

PR #510, opened 2026-01-08
Bump `llvm-pretty{,bc-parser}`, `crucible` submodules for LLVM 20

### LINKED ISSUE #509: Update `llvm-pretty{,-bc-parser}` submodules
They recently gained support for LLVM 19 and 20, and had some corresponding interface changes. We should upgrade.

langston-barrett:
Fixes #509.

### DIFF
--- a/grease/src/Grease/LLVM/DebugInfo.hs
+++ b/grease/src/Grease/LLVM/DebugInfo.hs
@@ -65,7 +65,10 @@ minTypeSize :: CLM.HasPtrWidth w => LDU.Info -> Bytes
 minTypeSize =
   \case
     LDU.ArrInfo _elemTy -> 0 -- arrays can be empty
-    LDU.BaseType _nm dibt -> Bytes.bitsToBytes (L.dibtSize dibt)
+    LDU.BaseType _nm dibt ->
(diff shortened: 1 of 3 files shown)


grease/src/Grease/LLVM/DebugInfo.hs
```diff
@@ -65,7 +65,10 @@ minTypeSize :: CLM.HasPtrWidth w => LDU.Info -> Bytes
 minTypeSize =
   \case
     LDU.ArrInfo _elemTy -> 0 -- arrays can be empty
-    LDU.BaseType _nm dibt -> Bytes.bitsToBytes (L.dibtSize dibt)
+    LDU.BaseType _nm dibt ->
+      case L.dibtSize dibt of
```
2over12, line 69: No reason to block really but maybe some sort of new llvm-20 bitcode test that touches this path or something wouldn't be bad

2over12 (approved):

grease/src/Grease/LLVM/DebugInfo.hs
langston-barrett: Yes, good point! #511.

PR #513, opened 2026-01-09
llvm: Only create allocations for overrides when necessary

### LINKED ISSUE #508: bug: `Missing implementation for 'malloc'`
[test.zip](https://github.com/user-attachments/files/24508736/test.zip)
```
cabal run exe:grease -- libxml2*.bc --symbol htmlParseAttValue --use-debug-info-types
```
```
...
Heuristic: grow and initialize memory referenced by: %0[7]
Heuristic: grow and initialize memory referenced by: %0[7]
Heuristic: grow and initialize memory referenced by: %0[7]
Missing implementation for 'malloc'
Finished analyzing 'htmlParseAttValue'. Missing implementation for 'malloc'
```
No idea how this could happen, `malloc` has a built-in override.

Also happens with symbols:

- `xz_decomp`
- `xmlParsePEReference`
- `htmlParseAttValue`

... and others

langston-barrett:
If an allocation already exists, just reuse it.

Fixes #508.

Feels like we should be able to economize further, but deleting the `forM_` over the declarations and unconditionally using `do_register...` both failed. Could use more investigation, but this is a step in the right direction.

### DIFF
--- a/grease/src/Grease/LLVM/Overrides.hs
+++ b/grease/src/Grease/LLVM/Overrides.hs
@@ -190,9 +190,12 @@ registerLLVMOverrides la builtinOvs userOvs skipFuns bak llvmCtx fs defns decls 
     -- part is necessary.
     CS.modifyGlobal mvar $ \mem ->
       case Map.lookup symb (CLM.memImplGlobalMap mem) of
+        -- Functions from LLVM modules will already have associated allocations,
+        -- don't create new ones
         Just (CLM.SomePointer ptr)
           | CLM.PtrWidth <- CLM.ptrWidth ptr ->
               pure (ptr, mem)
+        -- Functions from S-expression files need allocations
         _ -> liftIO (CLLVM.registerFunPtr bak mem name symb aliases)
 
   -- Note the order here: we register "skip" overrides *first*, so that built-in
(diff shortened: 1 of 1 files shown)


grease/src/Grease/LLVM/Overrides.hs
```diff
@@ -181,12 +181,17 @@ registerLLVMOverrides la builtinOvs userOvs skipFuns bak llvmCtx fs defns decls
   let allDecls = decls List.++ basicDecls List.++ fwdDeclLLVMDecls
   Foldable.forM_ allDecls $ \decl -> do
     doLog la (Diag.FoundDeclare decl)
-    let L.Symbol name = L.decName decl
+    let symb = L.decName decl
+    let L.Symbol name = symb
     let aliases = []
     -- See the module comment on "Lang.Crucible.LLVM.Functions" for why this
```
2over12, line 189: I know this isnt your fault but with the code under this being modified a bit this could use dome documentation. My current understanding based on follow up is that globals already exist due to previous initialization so we need to ensure to bind them

grease/src/Grease/LLVM/Overrides.hs
```diff
@@ -252,8 +257,16 @@ registerLLVMOverrides la builtinOvs userOvs skipFuns bak llvmCtx fs defns decls
...
-    CLLVM.alloc_and_register_override bak llvmCtx ov []
+  registerOv (CLLVM.SomeLLVMOverride ov) = do
+    let decl = CLLVM.llvmOverride_declare ov
+    let symb = L.decName decl
+    let mvar = CLLVM.llvmMemVar llvmCtx
+    mem <- CS.readGlobal mvar
+    case Map.lookup symb (CLM.memImplGlobalMap mem) of
+      Just (CLM.SomePointer ptr)
+        | CLM.PtrWidth <- CLM.ptrWidth ptr ->
+            CLLVM.do_register_llvm_override llvmCtx ov
```
2over12, line 275: It might be nice to have an explanation of the conditions under which a funptr is already allocated. Best I can tell this will happen if it is the case that an override targets a declaration that is used to initialize some global? Then that const global already has an fptr assigned and we dont end up binding it

langston-barrett: Good call. Functions should generally have allocations if they have `define`s/`declare`s in an LLVM module. `define`s/`declare`s from an S-expresion file will generally not. I think in Crucible-LLVM speak, function allocations are const globals. Will add a comment saying this.

PR #514, opened 2026-01-09
Fix misleading faux-aliasing during shape printing 

langston-barrett:


grease/src/Grease/Setup.hs
```diff
@@ -168,11 +173,12 @@ setupPtrMem la bak dl nm sel tgt@(PtrTarget bid _) =
           case Map.lookup bid' resMap of
             Just memoizeRes -> pure (setupResPtr memoizeRes, setupResTgt memoizeRes)
             Nothing -> do
-              (ptr, rtgt) <- unseenFallback
+              (ptr, rTgt) <- unseenFallback
               s <- get
-              let newmap = Map.insert bid' (SetupRes{setupResPtr = ptr, setupResTgt = rtgt}) resMap
-              _ <- put (s{_setupRes = newmap})
-              pure (ptr, rtgt)
+              let res = SetupRes{setupResPtr = ptr, setupResTgt = rTgt}
```
2over12, line 177: No worries but would a lense make this a bit cleaner? Something like zooming on the setupRes and using a setter

grease/src/Grease/Shape/Print.hs
```diff
@@ -56,14 +56,25 @@ import Prelude hiding (print)
...
   { allocs :: IntMap (PP.Doc Void)
+  , aliases :: IntMap Int
+  -- ^ A map from 'BlockId's (from 'PtrShape.PtrTarget's) to keys of 'allocs'
   , maxKey :: {-# UNPACK #-} !Int
   }
 
 allocNext :: Allocs -> (BlockId, Allocs)
 allocNext as = let k = maxKey as in (BlockId k, as{maxKey = k + 1})
 
+aliasNext :: BlockId -> Allocs -> (BlockId, Allocs)
```
2over12, line 70: Maybe a doc here, i guess the point is either fresh or use the block id already allocated

grease/src/Grease/Shape/Print.hs
```diff
@@ -56,14 +56,25 @@ import Prelude hiding (print)
...
+aliasNext (BlockId bid) as =
+  case IntMap.lookup bid (aliases as) of
+    Just bid' -> (BlockId bid', as)
+    Nothing ->
+      let (bid'@(BlockId bidVal), as') = allocNext as
```
2over12, line 75: maybe a getter here could flatten this a bit

grease/src/Grease/Shape/Print.hs
```diff
@@ -56,14 +56,25 @@ import Prelude hiding (print)
...
+       in let as'' = as{aliases = IntMap.insert bid bidVal (aliases as')}
```
2over12: similar maybe a setter

grease/src/Grease/Shape/Print.hs
langston-barrett: Not sure that's a huge win, it's kind of just moving destructuring out of the pattern into a later call. Since the pattern isn't *that* complicated and is the only thing happening on this line, I think I'll keep this as-is for now.

grease/src/Grease/Setup.hs
langston-barrett: Agreed, rewritten using a lens!

grease/src/Grease/Shape/Print.hs
langston-barrett: Extracted a helper into a `where`-clause.

PR #518, opened 2026-01-13
Add conservative shapes

2over12:
Adds a debug mode for shapes in DWARF that replaces composites with initialized bytes and only unrolls top-level pointers (secondary pointers are replaced with initialized bytes). This setup allows null pointers

### DIFF
--- a/grease/src/Grease/Macaw/Dwarf.hs
+++ b/grease/src/Grease/Macaw/Dwarf.hs
@@ -160,7 +160,7 @@ constructPtrTarget ::
   MDwarf.TypeApp ->
   UseConservativeDebugShapes ->
   IO (PtrTarget w NoTag)
-constructPtrTarget gla tyUnrollBound sprog visitCount tyApp (UseConservativeDebugShapes shouldBeConservative) =
+constructPtrTarget gla tyUnrollBound sprog visitCount tyApp isConservative =
   ptrTarget Nothing
     <$> liftIO
       ( do
@@ -172,6 +172,7 @@ constructPtrTarget gla tyUnrollBound sprog visitCount tyApp (UseConservativeDebu
             Right x -> pure x
       )
  where
+  UseConservativeDebugShapes shouldBeConservative = isConservative
   ishape w = except $ Right $ Seq.singleton $ Initialized NoTag (toBytes w)
   padding :: Integral a => a -> MemShape w NoTag
   padding w = Uninitialized (toBytes w)


grease/src/Grease/Macaw/Dwarf.hs
```diff
@@ -110,14 +116,17 @@ loadDwarfPreconditions ::
   Elf.ElfHeaderInfo (MC.ArchAddrWidth arch) ->
   -- | Architecture-specific information, needed for ABI info
   ArchContext arch ->
+  -- | Wether pointers should be instantiated conservatively (that is just substituting a size)
```
langston-barrett:
```suggestion
  -- | Whether pointers should be instantiated conservatively (that is just substituting a size)
```

grease/src/Grease/Macaw/Dwarf.hs
```diff
@@ -110,14 +116,17 @@ loadDwarfPreconditions ::
...
+  -- or form the precise type when possible.
```
langston-barrett:
```suggestion
  -- or from the precise type when possible.
```

Really, I might just expand the Haddock on `ShouldUseConservativeDebugShapes` and leave off the comment here.

grease/src/Grease/Macaw/Dwarf.hs
```diff
@@ -57,6 +59,10 @@ import Lang.Crucible.LLVM.MemModel qualified as CLM
 import Lang.Crucible.Types qualified as CT
 import Lumberjack qualified as LJ
 
+-- | Wether to use conservative shapes that do not initialize pointers (allowing for null)
+-- unless they are base pointers
+newtype ShouldUseConservativeDebugShapes = ShouldUseConservativeDebugShapes Bool
```
langston-barrett: A little shorter perhaps:
```suggestion
newtype UseConservativeDebugShapes = UseConservativeDebugShapes Bool
```

grease/src/Grease/Macaw/Dwarf.hs
```diff
@@ -149,8 +158,9 @@ constructPtrTarget ::
   Subprogram ->
   VisitCount ->
   MDwarf.TypeApp ->
+  ShouldUseConservativeDebugShapes ->
   IO (PtrTarget w NoTag)
-constructPtrTarget gla tyUnrollBound sprog visitCount tyApp =
+constructPtrTarget gla tyUnrollBound sprog visitCount tyApp (ShouldUseConservativeDebugShapes shouldBeConservative) =
```
langston-barrett: Long line, perhaps match this in a `let`

grease/src/Grease/Macaw/Shapes.hs
```diff
@@ -78,12 +79,15 @@ macawInitArgShapes la bak archCtx opts parsed elf memory argNames mbCfgAddr = do
...
+            shouldBeConservative
+  let getDwarfArgs shouldBeConservative = do
+        v <- MaybeT.runMaybeT $ getDwarfArgsMb shouldBeConservative
         pure $ Maybe.fromMaybe initArgs0 v
-      else pure initArgs0
+  dwarfedArgs <-
+    case shouldUseDwarf of
+      GO.NoDebugInfoShapes -> pure initArgs0
+      GO.ConservativeDebugInfoShapes -> getDwarfArgs (Dwarf.ShouldUseConservativeDebugShapes True)
+      GO.PreciseDebugInfoShapes -> getDwarfArgs (Dwarf.ShouldUseConservativeDebugShapes False)
```
langston-barrett: nbd, but feels like perhaps `shouldUseDwarf` could be a `Maybe` of an enumeration where these are the two options, and we could eliminate `ShouldUseConservativeDebugShapes` entirely and just match on that enum. Might be a little simpler.

2over12: Im currently leaving this as is for the parser if we have strong feelings about changing it I can

PR #522, opened 2026-01-16
doc: Fix typo re: debug info

langston-barrett:


2over12 (approved):

langston-barrett:
Ah, CI isn't running bc this is a doc-only change... This is a known problem with GH where it's impossible to have required status checks that are only required when they are run :/ We should probably just disable the path-based ignore workflow triggers.

PR #525, opened 2026-01-16
llvm: Adapt to removal of LLVM declarations from LLVM overrides

langston-barrett:
Testing and preparing for https://github.com/GaloisInc/crucible/pull/1705.

Changes:

- S-expression programs now must use the actual declared type of the override. This is generally an improvement, see our test cases.

### DIFF
--- a/grease/src/Grease/LLVM/Overrides.hs
+++ b/grease/src/Grease/LLVM/Overrides.hs
@@ -215,6 +215,11 @@ registerLLVMOverrides la builtinOvs userOvs skipFuns bak llvmCtx fs defns decls 
   -- The Crucible-LLVM "ABI" only has `LLVMPointer`s, not e.g., raw bitvectors.
   -- Lowering turns non-pointer arguments and return values into pointers. See
   -- Lang.Crucible.LLVM.Intrinsics.Cast for more details.
+  --
+  -- We need to register *both* the unlowered and lowered overrides because
+  -- users can e.g., analyze an LLVM bitcode module with a user-provided
+  -- S-expression override that calls a built-in override via its unlowered
+  -- signature.
   let loweredOvs = Cast.lowerOverrideTemplate <$> builtinOvs
   let ovs = Foldable.toList (loweredOvs <> builtinOvs)
   builtinDeclOvs <-


grease/src/Grease/LLVM/Overrides/Diagnostic.hs
```diff
@@ -9,38 +12,71 @@ module Grease.LLVM.Overrides.Diagnostic (
...
+          PP.pretty ("<unknown type: " <> Symb.symbolRepr symbRepr <> ">")
+        Just Refl ->
+          case Ctx.viewAssign tyCtx of
+            Ctx.AssignExtend (Ctx.viewAssign -> Ctx.AssignEmpty) (BVRepr w) ->
+              "(Ptr" PP.<+> PP.viaShow w <> ")"
+            -- These are impossible by the definition of LLVMPointerImpl
+            Ctx.AssignEmpty ->
+              panic "ppType" ["Impossible: LLVMPointerType empty context"]
+            Ctx.AssignExtend _ _ ->
+              panic "ppType" ["Impossible: LLVMPointerType ill-formed context"]
```
langston-barrett: https://github.com/GaloisInc/crucible/pull/1707

grease-exe/tests/ux/llvm-override-weird-type.lua
```diff
@@ -1,6 +1,6 @@
 -- Test the UX of our error messages for unsupported types in LLVM overrides
```
langston-barrett: This comment needs an update at least

grease/src/Grease/LLVM/Overrides.hs
```diff
@@ -64,75 +65,72 @@ doLog la diag = LJ.writeLog la (LLVMOverridesDiagnostic diag)
...
+      CS.bindFnHandle hdl (CS.UseOverride ov)
+    _ ->
+      fail $
+        unlines $
+          [ "Bad signature in `declare`"
+          , " name: " ++ show fNm
+          , "  `declare` args: " ++ show hdlArgs
+          , "  override args:  " ++ show llArgs
+          , "  `declare` ret:  " ++ show hdlRet
+          , "  override ret:   " ++ show llRet
```
langston-barrett: These should all use `ppTypeRepr`

grease/src/Grease/LLVM/Overrides.hs
```diff
@@ -206,8 +203,20 @@ registerLLVMOverrides la builtinOvs userOvs skipFuns bak llvmCtx fs defns decls
...
+
+  -- For S-expression files, we register the overrides as-is.
+  let builtinOvsList = Foldable.toList builtinOvs
+  builtinBasicDeclOvs <-
+    CLI.register_llvm_overrides_ llvmCtx builtinOvsList basicDecls
+  builtinFwdDeclOvs <-
+    CLI.register_llvm_overrides_ llvmCtx builtinOvsList fwdDeclDecls
+  -- For LLVM modules, we need to also try the "lowered" overrides.
+  let loweredOvs = Cast.lowerOverrideTemplate <$> builtinOvs
+  let ovs = Foldable.toList (loweredOvs <> builtinOvs)
```
RyanGlScott, line 224: Can you explain a little more about why you need to register both the original overrides and the lowered ones? Is this to maintain compatibility with the old convention? (If so, why do all of the `.cbl` files in `grease-exe/tests/llvm-bc` need to have their type signatures updated?)

RyanGlScott (commented):
Looks generally reasonable, but one question inline.

grease/src/Grease/LLVM/Overrides.hs
langston-barrett: Yeah, good q. For S-expression programs, we register overrides at their sort of "native", "unlowered" types - types that involve `(Bitvector 32)` arguments where the C header would have an `int`. But the Crucible-LLVM "ABI" only has `(Ptr N)`s for arguments and return values. So we also need to try those here. Will add comment.

langston-barrett: Let me know what you think!

RyanGlScott: I understand the need to register the lowered overrides, given that `crucible-llvm`'s ABI only supports `Ptr`s. My question more about why we need to register both the lowered _and_ the original overrides. If I understand correctly, this means that every function being overridden will have two (potentially distinct) types, either of which can be used? I don't think that's necessarily wrong, but it is a bit odd.

Either way, I wonder if the conventions here should be documented under `doc` somewhere.

langston-barrett: I believe that this is because we can have analysis that mixes S-expression files with bitcode modules. For example, you might be analyzing an LLVM bitcode module, but have an S-expression override that itself calls a built-in override via it's non-lowered signature.

RyanGlScott: Fair enough.

PR #536, opened 2026-02-15
Make offsets optional after Setup

2over12:
The idea of this is to prepare for #535  and split off some of the work. We want preconditions to be able to support setting up symbolic offsets. We want offsets to not be represented by concrete values after setup. The idea on how to do this is to make offsets in pointers Maybe Offset and Maybe Offset will be Nothing after setup. We then make concretize print from the actual symbolic value rather than the concrete offset. 

This will allow SL preconds to instead use a precond type that after setup gives pointers an arbitrary symbolic offset, and then concretization will retrieve whatever value this could be.

Alternative representations:
This setup isnt the best because the type doesnt specify what state the shape is in wrt to where the offset is stored. An alternative I could imagine is changing how NoTags work. Essentially Shape NotTag needs to result in a concrete offset. We could maybe have NoTag of llvm pointers end up being concrete offsets and then fix the input tag type for setup but that's a bit of a tricky refactor. Only thing i came up with here is a secondary type parameter that is like the "offset tag" and for NoTag we pass that as something that ignores the type and produces a concrete offset, for the rest we use the normal tag

Made with a fair bit of claude code

### DIFF
--- a/grease/src/Grease/Shape/Print.hs
+++ b/grease/src/Grease/Shape/Print.hs
@@ -245,10 +245,6 @@ printPtrWithOffset getOffset =
       let offset = getOffset tag mOffset
       printBlockOffset blk offset
 
--- | Ignores @tag@s.
-printPtr :: PtrShape ext w tag ty -> Printer w (PP.Doc ann)
-printPtr = printPtrWithOffset (\_ mOffset -> fromMaybe (PtrShape.Offset 0) mOffset)
-
 printBv :: NatRepr w' -> Printer w (PP.Doc ann)
 printBv w' = do
   -- TODO: assert that width % 8 == 0?


2over12:
Thinking about this a bit more I might opt for a slightly different design. Essentially abstract over a ptrShapeData type parameter that is either empty or contains the offset and the target. This will allow us to remove both the target and offset safely after Setup. We need this to allow muxing pointers against preconditions in SL.

Removing the target will require the concretize printer to be able to access a map from block->target so that's going to have to be a fairly big refactor where Setup now collects a map BlockID->Setup Shape. But that would be the idea

2over12:
> Working on such a refactor right now that uses a family over the ptr data... quite a lot of work. If we stay with something runtime dependent then we at least need to also make an ADT under MemShape as well So something like
> 
> ```
> data PtrData
> | Prec (Offset,PtrTarget)
> | NoData
> ```
> 
> and use that directly inside MemShape and PtrShape. This would be easier but we lose out on type safety

Making an attempt at this style of setup on branch ian/transform-pointer-data

PR #537, opened 2026-02-17
Make specification pointer data optional in shape

2over12:
The point of this PR is to make the Shape structure a viable output for other specification setups. The intention here is that SL specifications will implement a setup function that will return shapes with the 'NoData tag just like the concrete specifications do. This setup allows us to reuse more of grease than the original attempt at abstracting over preconditions. 

I have also changed concretize so we do full concretization before returning cdata (that is we call concretize in makeConcretizeData) so that we return a 'Precond version of the argshapes. This setup makes printing and comparison to the original precond easier (altho interestingshapes will have to change in the future somehow since we will no longer have a precond to compare to).

There is remaining work that needs to happen in a follow up PR which is also getting rid of the concrete assumptions about memshapes in printing. That is currently memshapes are flat sequences of initialized values. Ideally we would move these to a map from offset -> cell with the only assumption being that they are disjoint. That setup will allow the SL precondition to create cells with "gaps".

### DIFF
--- a/screach/src/Screach.hs
+++ b/screach/src/Screach.hs
@@ -1608,7 +1608,7 @@ analyzeCfg conf sla gla halloc macawCfgConfig archCtx mbEhi setupHook rtLoc exec
 
 verifyReachable ::
   ( p ~ SP.ScreachSimulatorState p0 sym bak ext (CS.RegEntry sym ret) arch t tys w
-  , TFC.TraversableFC (Shape.ExtShape ext Shape.Precond)
+  , TFC.TraversableFC (Shape.ExtShape ext 'Shape.Precond)
   , CCE.IsSyntaxExtension ext
   , GU.OnlineSolverAndBackend solver sym bak t st fm
   , 16 CT.<= w


grease-exe/tests/llvm/test-conc.llvm.cbl
```diff
@@ -0,0 +1,11 @@
+;; flags {"--symbol", "test",  "--initial-precondition", "tests/llvm/extra/out-of-bound-shape.txt", "--no-heuristics"}
```
langston-barrett, line 3: This test could use a comment saying what's being tested and why

grease-exe/tests/Shape.hs
```diff
@@ -106,27 +106,27 @@ eqPtrShape s1 s2 =
       case testEquality w w' of
         Maybe.Nothing -> False
         Maybe.Just Equality.Refl -> bv == bv'
-    (PtrShape.ShapePtr NoTag off1 tgt1, PtrShape.ShapePtr NoTag off2 tgt2) ->
-      off1 == off2 && eqPtrTarget tgt1 tgt2
+    (PtrShape.ShapePtr NoTag (PtrShape.PrecondPtrData mOffset1 tgt1), PtrShape.ShapePtr NoTag (PtrShape.PrecondPtrData mOffset2 tgt2)) ->
```
langston-barrett, line 109: Long line...

langston-barrett:
Did you also consider moving this data into the `tag`? How would that approach differ from this one with modes?

grease/src/Grease/Shape.hs
```diff
@@ -193,54 +194,47 @@ instance (MC.PrettyF tag, PrettyExt ext tag) => MC.PrettyF (Shape ext tag) where
...
-      ShapeBool tag -> ShapeBool <$> f tag
-      ShapeFloat tag fi -> ShapeFloat <$> f tag <*> pure fi
-      ShapeStruct tag fields ->
-        ShapeStruct <$> f tag <*> traverseFC (traverseFC f) fields
-      ShapeUnit tag -> ShapeUnit <$> f tag
-      ShapeExt ext -> ShapeExt <$> traverseFC f ext
+-- Note: FunctorFC/FoldableFC/TraversableFC instances cannot be provided for Shape
+-- because after adding the ptrData parameter, Shape ext has kind
+-- (C.CrucibleType -> Type) -> PtrDataMode -> C.CrucibleType -> Type, but these type
+-- classes expect (k -> *) -> k' -> *. Use traverseShapeWithType instead.
```
langston-barrett: This seems a bit unfortunate - perhaps mode should come first? I know that's a horrible change to make, but Claude Haiku might be able to help.

grease/src/Grease/Shape.hs
```diff
@@ -193,54 +194,47 @@ instance (MC.PrettyF tag, PrettyExt ext tag) => MC.PrettyF (Shape ext tag) where
...
-    ShapeExt ext -> ShapeExt <$> traversePtrShapeWithType f ext
+    ShapeExt ext -> ShapeExt <$> traversePtrShapeWithType mode f ext
 
 tagWithType ::
   CLM.HasPtrWidth wptr =>
-  (ExtShape ext ~ PtrShape ext wptr) =>
-  Shape ext tag t ->
-  Shape ext C.TypeRepr t
-tagWithType = runIdentity . traverseShapeWithType (\typeRepr _tag -> Identity typeRepr)
+  KnownPtrMode ptrData =>
```
langston-barrett, line 244: Might be more flexible to just take a `ModeRepr`?

grease/src/Grease/Shape/Pointer.hs
```diff
@@ -85,6 +88,45 @@ newtype ExtraStackSlots = ExtraStackSlots {getExtraStackSlots :: Int}
   -- See Note [Derive Read/Show instances with the newtype strategy]
   deriving newtype (Enum, Eq, Integral, Num, Ord, Read, Real, Show)
 
+-- | Mode indicating what pointer metadata is present
+data PtrDataMode = Precond | NoData
```
langston-barrett, line 97: This could use motivation: why do we have these modes?

grease/src/Grease/Shape/Pointer.hs
```diff
@@ -157,59 +202,67 @@ instance MC.PrettyF tag => PP.Pretty (MemShape wptr tag) where
...
-  traverseF f =
-    \case
-      Uninitialized bs -> pure (Uninitialized bs)
-      Initialized tag bs -> Initialized <$> f tag <*> pure bs
-      Pointer tag off tgt -> Pointer <$> f tag <*> pure off <*> TF.traverseF f tgt
-      Exactly bs -> Exactly <$> traverse (TF.traverseF f) bs
+-- Note: FunctorF/FoldableF/TraversableF instances cannot be provided for MemShape
+-- because after adding the ptrData parameter, MemShape wptr has kind
+-- (C.CrucibleType -> Type) -> PtrDataMode -> Type, but these type classes
+-- expect (k -> *) -> *. Use traverseMemShapeWithType instead.
```
langston-barrett: Same comment, to me this indicates that we should consider a different ordering for the type variables.

grease/src/Grease/Shape/Pointer.hs
```diff
@@ -282,44 +335,43 @@ shape, otherwise the first observed 'PtrTarget's shape will win.
...
-
-instance TF.FoldableF (PtrTarget wptr) where
-  foldMapF = TF.foldMapFDefault
-
-instance TF.TraversableF (PtrTarget wptr) where
-  traverseF f (PtrTarget bid tgt) = PtrTarget bid <$> traverse (TF.traverseF f) tgt
+-- Note: FunctorF/FoldableF/TraversableF instances cannot be provided for PtrTarget
+-- because after adding the ptrData parameter, PtrTarget wptr has kind
+-- (C.CrucibleType -> Type) -> PtrDataMode -> Type, but these type classes
+-- expect (k -> *) -> *. Use traversePtrTargetWithType instead.
```
langston-barrett: Same comment

grease/src/Grease/Shape/Pointer.hs
```diff
@@ -504,67 +557,78 @@ instance MC.PrettyF tag => MC.PrettyF (PtrShape ext w tag) where
...
+            PrecondPtrData offset tgt ->
+              PP.pretty tgt PP.<> "+" PP.<> PP.viaShow offset PP.<> ppTag tag
+        NoDataRepr ->
+          case dat of
+            NoPtrData -> "ptr" PP.<> ppTag tag PP.<> ":removed"
+
+-- Note: FunctorFC/FoldableFC/TraversableFC instances cannot be provided for PtrShape
+-- because after adding the ptrData parameter, PtrShape ext w has kind
+-- (C.CrucibleType -> Type) -> PtrDataMode -> C.CrucibleType -> Type, but these type
+-- classes expect (k -> *) -> k' -> *. Use traversePtrShapeWithType instead.
```
langston-barrett: Same comment

grease/src/Grease/Shape/Pointer.hs
```diff
@@ -791,19 +855,21 @@ bytesToPointers ::
...
     -- Need to keep dereferencing as specified by the path
     (CursorExt (DereferencePtr @_ @_ @ts' idx rest), PtrTarget _ ms) ->
       case ms Seq.!? idx of
-        Just (Pointer tag' off subTgt) -> do
+        Just (Pointer tag' ptrData) -> do
           C.Refl <- Right $ lastCons (Proxy @(CLM.LLVMPointerType w)) (Proxy @ts')
-          subTgt' <- bytesToPointers proxy tag rest subTgt
-          Right (ptrTarget Nothing (Seq.update idx (Pointer tag' off subTgt') ms))
+          case ptrData of
+            PrecondPtrData offset subTgt -> do
```
langston-barrett: Minor style nit, but perhaps just match this pattern above in the `Pointer` pattern?

grease/src/Grease/Shape/Pointer.hs
```diff
@@ -821,21 +887,23 @@ modifyPtrTarget ::
...
     -- Need to keep dereferencing as specified by the path
     (CursorExt (DereferencePtr @_ @_ @ts' idx rest), PtrTarget _ ms) ->
       case ms Seq.!? idx of
-        Just (Pointer tag off subTgt) -> do
+        Just (Pointer tag ptrData) -> do
           C.Refl <- pure $ lastCons (Proxy @(CLM.LLVMPointerType w)) (Proxy @ts')
-          subTgt' <- modifyPtrTarget proxy modify rest subTgt
-          Right (ptrTarget Nothing (Seq.update idx (Pointer tag off subTgt') ms))
+          case ptrData of
+            PrecondPtrData offset subTgt -> do
```
langston-barrett: Same comment

grease/src/Grease/Shape/Print.hs
```diff
@@ -180,69 +183,114 @@ printNamedShapes names =
...
-  ExtShape ext ~ PtrShape ext w =>
-  Shape ext tag ty ->
-  Printer w (PP.Doc ann)
-printShape =
+-- The offset extraction function is used for 'ShapePtr' to determine what offset
+-- to print. It receives the tag and the PtrData.
+printShapeWithOffset ::
+  (ExtShape ext tag ptrData ~ PtrShape ext w tag ptrData, PtrShape.KnownPtrMode ptrData) =>
+  -- | Function to extract offset for ShapePtr
+  (tag (CLM.LLVMPointerType w) -> PtrShape.PtrData ptrData w tag -> PtrShape.Offset) ->
```
langston-barrett: Perhaps this should just only work with the `'Precond` mode? Would this parameter be necessary in that case? (I definitely might be missing why we need this to work with multiple modes)

grease/src/Grease/Shape/Print.hs
```diff
@@ -180,69 +183,114 @@ printNamedShapes names =
...
+              blk <- printerAlloc (printTgt tgt) bid
+              let offset = getOffset tag dat
+              printBlockOffset blk offset
+        PtrShape.NoDataRepr ->
+          case dat of
+            PtrShape.NoPtrData -> do
+              -- Simplified: this case should not occur after refactoring
+              -- Keep for backward compatibility with any legacy code
+              let offset = getOffset tag dat
+              pure ("<?+" PP.<> PP.viaShow offset PP.<> ">")
```
langston-barrett: Is this a Claude-ism? 😄 If it doesn't occur, let's `panic`.

grease/src/Grease/Shape/Print.hs
```diff
@@ -299,13 +347,23 @@ bytesToInt :: CLB.Bytes -> Int
 bytesToInt = integerToInt . CLB.bytesToInteger
 
 -- | Ignores @tag@s.
-printMemShape :: PtrShape.MemShape w tag -> Printer w (PP.Doc Void)
+printMemShape :: forall ptrData w tag. PtrShape.KnownPtrMode ptrData => PtrShape.MemShape w tag ptrData -> Printer w tag (PP.Doc Void)
```
langston-barrett: Let's split this onto multiple lines - maybe ask Claude to do this for all lines that became longer in this change?

grease/src/Grease/Shape/Print.hs
```diff
@@ -299,13 +347,23 @@ bytesToInt :: CLB.Bytes -> Int
...
+        case dat of
+          PtrShape.PrecondPtrData offset tgt -> do
+            let bid = PtrShape.ptrTargetBlock tgt
+            blk <- printerAlloc (printTgt tgt) bid
+            printBlockOffset blk offset
+      PtrShape.NoDataRepr ->
+        case dat of
+          PtrShape.NoPtrData ->
+            -- Simplified: should not occur after refactoring
+            pure "<?+?>"
```
langston-barrett: Same comment

grease/src/Grease/Concretize.hs
```diff
@@ -185,6 +196,42 @@ makeConcretizedData bak groundEvalFn minfo initState extra = do
...
+ where
+  -- Helper to concretize a PtrTarget
+  concretizePtrTarget ::
+    (forall tp. C.TypeRepr tp -> CS.RegValue' sym tp -> IO (Conc.ConcRV' sym tp)) ->
+    ShapePtr.PtrTarget wptr (CS.RegValue' sym) 'ShapePtr.NoData ->
+    IO (ShapePtr.PtrTarget wptr (Conc.ConcRV' sym) 'ShapePtr.NoData)
+  concretizePtrTarget concRV (ShapePtr.PtrTarget bid ms) =
+    ShapePtr.PtrTarget bid <$> traverse (concretizeMemShape concRV) ms
+
+  -- Helper to concretize a MemShape
```
langston-barrett: Another Claude-ism? These comments seem unnecessary. I've had success asking Claude to remove comments that simply say what the code is doing, while leaving comments that say why it does it that way.

grease/src/Grease/Concretize.hs
```diff
@@ -185,6 +196,42 @@ makeConcretizedData bak groundEvalFn minfo initState extra = do
...
+        <$> concRV (C.VectorRepr (CLM.LLVMPointerRepr (knownNat @8))) tag
+        <*> pure bytes
+    ShapePtr.Pointer tag ShapePtr.NoPtrData ->
+      ShapePtr.Pointer
+        <$> concRV (CLM.LLVMPointerRepr ?ptrWidth) tag
+        <*> pure ShapePtr.NoPtrData
+    ShapePtr.Exactly bytes ->
+      ShapePtr.Exactly <$> traverse (concretizeTaggedByte concRV) bytes
+
+  -- Helper to concretize a TaggedByte
```
langston-barrett: Same comment

grease/src/Grease/Concretize.hs
```diff
@@ -185,6 +196,42 @@ makeConcretizedData bak groundEvalFn minfo initState extra = do
       , concMem = ConcMem cMem
       , concErr = cErr
       }
+ where
+  -- Helper to concretize a PtrTarget
```
langston-barrett: See below comment

grease/src/Grease/Setup.hs
```diff
@@ -232,6 +237,20 @@ setupPtr la bak layout nm sel target = do
...
+      case WI.asNat blockNat of
+        Just blockNum -> do
+          -- Convert PtrTarget from Precond to NoData mode
+          let targetNoData = removePtrTargetData (PtrTarget bid ms')
+          allocMap <- use setupAllocMap
+          setupAllocMap .= Map.insert blockNum targetNoData allocMap
+        Nothing ->
+          -- Block number is symbolic, which shouldn't happen for malloc'd pointers
+          -- but we'll just skip storing in the map in this case
+          pure ()
```
langston-barrett: Let's perhaps `panic` here?

grease/src/Grease/Setup.hs
```diff
@@ -454,6 +476,31 @@ setupShape la bak layout nm tRepr sel s = do
...
+--
+-- Converts a PtrTarget from 'Precond mode to 'NoData mode, discarding pointer data from all MemShapes.
+removePtrTargetData ::
+  forall w tag.
+  CLM.HasPtrWidth w =>
+  ShapePtr.PtrTarget w tag 'ShapePtr.Precond ->
+  ShapePtr.PtrTarget w tag 'ShapePtr.NoData
+removePtrTargetData (ShapePtr.PtrTarget bid ms) =
+  ShapePtr.PtrTarget bid (fmap removeMemShapePtrData ms)
+
```
langston-barrett, line 502: Why do we want to do this? (I trust there's a good reason, it's just not yet clear to me)

langston-barrett: Ah, I think the PR description answers this question:

> The intention here is that SL specifications will implement a setup function that will return shapes with the 'NoData tag just like the concrete specifications do.

Perhaps we should refer to a GH issue in some of these comments?

grease/src/Grease/Shape/Concretize.hs
```diff
@@ -8,65 +8,100 @@
...
+          blockNum = CLMP.concBlock ptr
+       in if blockNum == 0
             then PtrShape.ShapePtrBVLit tag (CLMP.concWidth ptr) (CLMP.concOffset ptr)
-            else PtrShape.ShapePtr tag off (concPtrTarget tgt)
+            else case Map.lookup (fromInteger blockNum) allocMap of
+              Just target ->
+                let concTarget = concPtrTarget allocMap target
+                    offsetBV = CLMP.concOffset ptr
+                    offsetBytes = BV.asUnsigned offsetBV
+                    offset = PtrShape.Offset (CLB.toBytes offsetBytes)
```
langston-barrett: This block appears to be duplicated - perhaps a helper function?

grease/src/Grease/Shape/Parse.hs
```diff
@@ -98,21 +98,21 @@ lookupAlloc blk as = IntMap.lookup (getBlockId blk) (getAllocs as)
...
   \case
     Uninitialized bs -> Either.Right (PtrShape.Uninitialized bs)
     Initialized bs -> Either.Right (PtrShape.Initialized NoTag bs)
     Exactly wds ->
       Either.Right (PtrShape.Exactly (List.map (PtrShape.TaggedByte NoTag) wds))
     Pointer blk off ->
-      PtrShape.Pointer NoTag off
-        Functor.<$> case lookupAlloc blk allocs of
+      Either.Right . PtrShape.Pointer NoTag . PtrShape.PrecondPtrData off
+        =<< case lookupAlloc blk allocs of
```
langston-barrett: Isn't `pure . f =<< x` equivalent to `f <$> x`?

grease/src/Grease/Shape/Parse.hs
```diff
@@ -121,34 +121,35 @@ type Parse :: Type -> Nat -> Type
...
 
 type instance Shape.ExtShape (Parse ext w) = ParsePtrShape w
 
 -- | Like 'PtrShape.PtrShape', except that 'PtrShape.PtrTarget' is replaced
--- by 'BlockId'.
-type ParsePtrShape :: Nat -> (C.CrucibleType -> Type) -> C.CrucibleType -> Type
-data ParsePtrShape w tag t where
+-- by 'BlockId'. Always uses 'Precond mode during parsing.
+type ParsePtrShape :: Nat -> (C.CrucibleType -> Type) -> PtrShape.PtrDataMode -> C.CrucibleType -> Type
+data ParsePtrShape w tag ptrData t where
```
langston-barrett: If `ptrData` is always `'Precond`, do we need to parameterize over it?

grease/src/Grease/Shape/Pointer.hs
```diff
@@ -85,6 +88,45 @@ newtype ExtraStackSlots = ExtraStackSlots {getExtraStackSlots :: Int}
...
+  = PrecondPtrData
+  { precondOffset :: Offset
+  , precondTarget :: PtrTarget w tag 'Precond
+  }
+
+data PtrModeRepr (tp :: PtrDataMode) where
+  PrecondRepr :: PtrModeRepr 'Precond
+  NoDataRepr :: PtrModeRepr 'NoData
+
+-- | Typeclass for types with a known PtrDataMode at compile time
```
langston-barrett:
```suggestion
-- | Typeclass for types with a known 'PtrDataMode' at compile time
```

grease/src/Grease/Shape/Pointer.hs
```diff
@@ -85,6 +88,45 @@ newtype ExtraStackSlots = ExtraStackSlots {getExtraStackSlots :: Int}
   -- See Note [Derive Read/Show instances with the newtype strategy]
   deriving newtype (Enum, Eq, Integral, Num, Ord, Read, Real, Show)
 
+-- | Mode indicating what pointer metadata is present
+data PtrDataMode = Precond | NoData
+
+-- | Pointer data that may or may not be present
+data family PtrData (mode :: PtrDataMode) (w :: Natural) (tag :: C.CrucibleType -> Type)
```
langston-barrett, line 100: Let's use SAKS here for readability

grease/src/Grease/Shape/Pointer.hs
```diff
@@ -85,6 +88,45 @@ newtype ExtraStackSlots = ExtraStackSlots {getExtraStackSlots :: Int}
...
+
+-- | Precondition mode: contains offset and target
+data instance PtrData 'Precond w tag
+  = PrecondPtrData
+  { precondOffset :: Offset
+  , precondTarget :: PtrTarget w tag 'Precond
+  }
+
+data PtrModeRepr (tp :: PtrDataMode) where
```
langston-barrett: Let's use SAKS here for consistency

grease/src/Grease/Shape/Pointer.hs
```diff
@@ -85,6 +88,45 @@ newtype ExtraStackSlots = ExtraStackSlots {getExtraStackSlots :: Int}
...
+  PrecondRepr :: PtrModeRepr 'Precond
+  NoDataRepr :: PtrModeRepr 'NoData
+
+-- | Typeclass for types with a known PtrDataMode at compile time
+class KnownPtrMode (mode :: PtrDataMode) where
```
langston-barrett: Let's use SAKS here for consistency

grease/src/Grease/Shape/Pointer.hs
```diff
@@ -157,59 +202,67 @@ instance MC.PrettyF tag => PP.Pretty (MemShape wptr tag) where
...
         <$> f (CLM.LLVMPointerRepr ?ptrWidth) tag
-        <*> pure off
-        <*> traversePtrTargetWithType f tgt
+        <*> case mode of
+          PrecondRepr ->
+            case ptrData of
+              PrecondPtrData offset tgt -> PrecondPtrData offset <$> traversePtrTargetWithType mode f tgt
+          NoDataRepr ->
+            case ptrData of
+              NoPtrData -> pure NoPtrData
```
langston-barrett: Could perhaps consider `case (mode, ptrData) of` to collapse this a bit

grease/src/Grease/Shape/Pointer.hs
```diff
@@ -504,67 +557,78 @@ instance MC.PrettyF tag => MC.PrettyF (PtrShape ext w tag) where
...
-      ShapePtr tag off tgt ->
-        ShapePtr <$> f tag <*> pure off <*> TF.traverseF f tgt
-
-ptrShapeType :: CLM.HasPtrWidth w => PtrShape ext w tag t -> C.TypeRepr t
+      ShapePtr tag ptrData ->
+        prettyPtrData tag ptrData
+   where
+    prettyPtrData :: tag (CLM.LLVMPointerType w) -> PtrData ptrData w tag -> PP.Doc ann
+    prettyPtrData tag dat =
+      case knownPtrMode @ptrData of
```
langston-barrett: Same comment re: collapsing these `case`s

grease/src/Grease/Shape/Pointer.hs
```diff
@@ -504,67 +557,78 @@ instance MC.PrettyF tag => MC.PrettyF (PtrShape ext w tag) where
...
       ShapePtrBVLit <$> f (CLM.LLVMPointerRepr w) tag <*> pure w <*> pure bv
-    ShapePtr tag off tgt ->
+    ShapePtr tag ptrData ->
       ShapePtr
         <$> f (CLM.LLVMPointerRepr ?ptrWidth) tag
-        <*> pure off
-        <*> traversePtrTargetWithType f tgt
+        <*> traversePtrDataHelper f ptrData
+ where
+  traversePtrDataHelper :: Applicative m => (forall x. C.TypeRepr x -> tag x -> m (tag' x)) -> PtrData mode wptr tag -> m (PtrData mode wptr tag')
```
langston-barrett: Long line...

grease-exe/tests/Shape.hs
2over12: Since we are through the looking glass on code-gen do we just want to enable a column limit in fourmolu?

grease/src/Grease/Shape.hs
2over12: I think this works fine? we will know the ModeRepr at each state?

grease-exe/tests/Shape.hs
langston-barrett: I'm fine with that.

That being said, doing so will not address the underlying concern. The concern is code readability, which requires that we actually take a look through the code and consider how to make it more readable, not just run Claude and then the formatter and call it a day. For example, formolu would probably  happily split a complex expression like
```haskell
let superLongExpr = function (calledWith manyDifferent) args maybeEvenThreeArgs orFour (orFive thatsALot ofArgs)
```
into
```haskell
let superLongExpr =
      function
        (calledWith manyDifferent)
        args
        maybeEvenThreeArgs
        orFour
        (orFive thatsALot ofArgs)
```
when what the situation actually calls for is packaging three of those arguments into a single `data` type and computing the last argument in a separate `let`.

grease/src/Grease/Shape/Parse.hs
2over12: The answer I believe is yes because Shapes must be of a kind that takes a mode to integrate with ExtShape properly

grease/src/Grease/Shape/Concretize.hs
```diff
@@ -8,65 +8,110 @@
...
+import Lang.Crucible.LLVM.Bytes qualified as CLB
 import Lang.Crucible.LLVM.MemModel.Pointer qualified as CLMP
+import Numeric.Natural (Natural)
 
--- | Turns 'PtrShape.Initialized' into 'PtrShape.Exactly', calls 'concPtrTarget'.
+-- | Look up a block in the allocation map and create PrecondPtrData from a concrete pointer
+--
+-- This helper eliminates duplication between concMemShape and concPtrShape by
+-- encapsulating the logic of: looking up the block, concretizing the target,
+-- extracting the offset, and constructing PrecondPtrData.
```
langston-barrett: This is a "historical" comment that looks LLM-generated. Perhaps you could ask an agent to strip out these comments? I myself use a skill for this purpose: https://github.com/langston-barrett/dots/blob/main/files/claude/skills/comments/SKILL.md

grease/src/Grease/Shape/Pointer.hs
```diff
@@ -128,25 +177,28 @@ traverseTaggedByte f (TaggedByte tag val) = TaggedByte <$> f tag <*> pure val
 --
 -- * @wptr@: Width of a pointer, in bits
 -- * @tag@: See 'Grease.Shape.Shape'
-data MemShape wptr tag
+-- * @ptrData@: Mode indicating whether pointer metadata is present
+type MemShape :: Natural -> PtrDataMode -> (C.CrucibleType -> Type) -> Type
+data MemShape wptr ptrData tag
```
langston-barrett, line 182: This might be a stupid question, but here goes: Do we need `MemShape` (and `PtrTarget`) to be parameterized over the `PtrDataMode`? `MemShape` only shows up in `PtrShape` as a field of `ShapePtr`. But if the `ShapePtr` holds `NoPtrData`, then it doesn't hold a `MemShape`. Basically, if a `MemShape` is present in `PtrShape`, then we must have `PtrDataMode ~ Precond`, right?

grease/src/Grease/Shape.hs
```diff
@@ -193,54 +194,92 @@ instance (MC.PrettyF tag, PrettyExt ext tag) => MC.PrettyF (Shape ext tag) where
       ShapeUnit tag -> "unit" PP.<> ppTag tag
       ShapeExt ext -> MC.prettyF ext
 
-instance TFC.TraversableFC (ExtShape ext) => TFC.FunctorFC (Shape ext) where
+-- | Instances for Shape ext 'Precond
```
langston-barrett: Another unhelpful comment

grease/src/Grease/Shape.hs
```diff
@@ -193,54 +194,92 @@ instance (MC.PrettyF tag, PrettyExt ext tag) => MC.PrettyF (Shape ext tag) where
...
+  where
+  traverseFC f = \case
+    ShapeBool tag -> ShapeBool <$> f tag
+    ShapeFloat tag fi -> ShapeFloat <$> f tag <*> pure fi
+    ShapeStruct tag fields ->
+      ShapeStruct <$> f tag <*> TFC.traverseFC (TFC.traverseFC f) fields
+    ShapeUnit tag -> ShapeUnit <$> f tag
+    ShapeExt ext -> ShapeExt <$> TFC.traverseFC f ext
+
+-- | Instances for Shape ext 'NoData
```
langston-barrett: Also unhelpful

grease/src/Grease/Shape.hs
```diff
@@ -193,54 +194,92 @@ instance (MC.PrettyF tag, PrettyExt ext tag) => MC.PrettyF (Shape ext tag) where
...
+  (TFC.TraversableFC (ExtShape ext 'NoData)) =>
+  TFC.TraversableFC (Shape ext 'NoData)
+  where
+  traverseFC f = \case
+    ShapeBool tag -> ShapeBool <$> f tag
+    ShapeFloat tag fi -> ShapeFloat <$> f tag <*> pure fi
+    ShapeStruct tag fields ->
+      ShapeStruct <$> f tag <*> TFC.traverseFC (TFC.traverseFC f) fields
+    ShapeUnit tag -> ShapeUnit <$> f tag
+    ShapeExt ext -> ShapeExt <$> TFC.traverseFC f ext
```
langston-barrett, line 219: Is this a duplicate of the instance for `'Precond`?

grease/src/Grease/Shape/Pointer.hs
2over12: The MemShape holds NoData post setup in the allocation map so we exercise both cases. We use the MemShapes in the map to rebuild the full shape later by concretizing the block and finding the corresponding memshape in the block map

langston-barrett: Gotcha, thanks!

langston-barrett (approved):

PR #544, opened 2026-03-04
Unified scheduler

langston-barrett:
Create a unified scheduler abstraction that can be used in GREASE, Screach, or other tools based on Crucible. Supports SSE, DFS, BFS, and prioritized exploration. Produced largely by an agent.

### DIFF
--- a/grease/src/Grease/Scheduler.hs
+++ b/grease/src/Grease/Scheduler.hs
@@ -392,10 +392,11 @@ withSatisfiabilityCheck considerSat policy =
               CB.addAssumption bak (CB.BranchCondition loc (C.pausedLoc pf) p')
               let ctx = st ^. C.stateTree . C.actContext
               C.ExecutionFeatureNewState <$> runReaderT (C.resumeFrame pf ctx) st
-          CBO.UnsatisfiableContext ->
-            -- Neither branch feasible: let execution continue normally
-            -- (this will likely result in an abort)
-            pure C.ExecutionFeatureNoChange
+          CBO.UnsatisfiableContext -> do
+            mnext <- dequeue wq
+            case mnext of
+              Nothing -> pure C.ExecutionFeatureNoChange
+              Just next -> C.ExecutionFeatureNewState <$> restoreWorkItem next
           CBO.IndeterminateBranchResult ->
             -- Both branches may be feasible: delegate to the underlying policy
             policyOnBranch policy wq p trueFrame falseFrame bt st
--- a/screach/src/Screach/ShortestDistanceScheduler.hs
+++ b/screach/src/Screach/ShortestDistanceScheduler.hs
@@ -14,7 +14,7 @@ import Control.Applicative ((<|>))
 import Control.Lens ((^.))
 import Control.Monad (forM)
 import Control.Monad.Reader (ReaderT (runReaderT))
-import Control.Monad.State (MonadIO (liftIO), StateT, runStateT)
+import Control.Monad.State (MonadIO (liftIO), MonadTrans (lift), StateT, runStateT)
 import Control.Monad.Trans.Maybe (MaybeT (MaybeT, runMaybeT))
 import Data.IORef (IORef)
 import Data.IORef qualified as IORef
@@ -201,7 +201,7 @@ sdsePrioritizationFunction tgtAddr tgtFunction archCtx cg cache distConfig sla g
         do
           (cfg, snode) <- MaybeT $ liftIO $ getExplorationEntry state frame
           let rcall (Dist.FunctionEntry fentry) (Dist.Callsite callsite) = CG.resolveCall cg cache sla gla mem halloc archCtx symMap pltStubs fentry callsite
-          Dist.CallStack cs <- MaybeT $ liftIO $ runWithCachesRef cachesRef (Just <$> callStackFromSimState state)
+          Dist.CallStack cs <- MaybeT $ lift $ Just <$> callStackFromSimState state
           let poppedCS = Maybe.fromMaybe [] $ tailMay cs
           let
             isT ::


screach/src/Screach.hs
```diff
@@ -894,7 +894,7 @@ withPrioritizationFunction conf avoidList archCtx cfgCache sla gla elf halloc re
           Sched.PrioritizationFunction p sym ext rtp
         cgPfunc = case SAL.resolvedTargetAddr resolvedTargetLoc of
           Just targetAddr -> pfuncFromTargetAddr targetAddr
-          _ -> Sched.defaultPrioritizationFunction
+          _ -> (\_ _ -> pure Nothing)
```
langston-barrett: NTS: need to restore this

screach/src/Screach/RefineFeature.hs
```diff
@@ -342,18 +370,33 @@ sdseExecFeatures ::
...
-    , PthSplit.pathSplitFeature
-        sla
-        schedCallback
-        (CBO.considerSatisfiability bak)
-    ]
+  -- | Returns the features and an IORef to retrieve saved target results after execution
+  IO ([C.ExecutionFeature p sym ext (CS.RegEntry sym ret)], IORef [SavedState p sym ext rtp])
+sdseExecFeatures bak sla gla refineReplay initCFG priorityFunc isTargetPred (RftOpt.AllSolutions exploreMore) = do
+  -- Create IORef for collecting target results
+  savedRef <- newIORef []
```
2over12, line 377: Is there no way to keep this in the personality like Scheduler.hs used to? I'd imagine GREASE could also maybe use an isTarget strategy or something to save buggy states to try to address stuff like https://github.com/GaloisInc/grease/issues/476

2over12: Idea here would be to start to bring Grease in line with screach where we can just run executeCrucible and collect the results from scheduler. Would be a nice step towards unifying entrypoints potentially

langston-barrett: Yeah I think this would be a good follow-up!

screach/src/Screach.hs
```diff
@@ -667,7 +667,7 @@ analyzeSyntax conf sla gla halloc archCtx = do
     (GMSH.ExecutingAddressAction (\_ -> pure ()))
     addrOvs
     entryCfgs
-    Sched.defaultPrioritizationFunction
+    (\_ _ -> pure Nothing)
```
RyanGlScott: Is there a reason not to use `defaultPrioritizationFunction` here?

screach/src/Screach/RefineFeature.hs
```diff
@@ -126,47 +129,70 @@ class HasRefinmentState p sym bak t ext aty w | p -> sym ext where
...
+          SavedState
+            { _savedBackendState = bakState
+            , _savedExecState = C.ResultState st
+            }
 
+-- | Take a fresh initial state (from a given refinement) and set it up for replay.
+--
+-- Returning this via 'C.ExecutionFeatureNewState' is not ideal because the
+-- simulator will immediately begin exploring this state, without consulting
+-- the worklist. But the refined state my be further than some other state on
```
RyanGlScott:
```suggestion
-- the worklist. But the refined state may be further than some other state on
```

screach/src/Screach/RefineFeature.hs
```diff
@@ -126,47 +129,70 @@ class HasRefinmentState p sym bak t ext aty w | p -> sym ext where
...
+-- the worklist from the target. Once we hit a branch we will get to visit the
+-- scheduler again but until that point we are forced to run the refinement.
+-- Ideally we would take the refinement and add it to the worklist. Then let
+-- the scheduler pick the best state This is not possible because we cannot
```
RyanGlScott:
```suggestion
-- the scheduler pick the best state. This is not possible because we cannot
```

screach/src/Screach/RefineFeature.hs
```diff
@@ -126,47 +129,70 @@ class HasRefinmentState p sym bak t ext aty w | p -> sym ext where
...
--- all features see it or something but it would be really ugly.
+-- | Saved symbolic execution result with backend state.
+--
+-- When a target is reached during SDSE, the result and backend state are
+-- captured so they can be verified later.
+data SavedState p sym ext rtp
+  = SavedState
+  { _savedBackendState :: CB.AssumptionState sym
+  , _savedExecState :: C.ExecState p sym ext rtp
+  }
```
RyanGlScott, line 145: Do we ever make lenses for these? If not, it feels a bit silly to prefix these field accessors with underscores.

grease/src/Grease/Scheduler.hs
```diff
@@ -0,0 +1,535 @@
...
+              let ctx = st ^. C.stateTree . C.actContext
+              C.ExecutionFeatureNewState <$> runReaderT (C.resumeFrame pf1' ctx) st
+          CBO.NoBranch False ->
+            -- Only the false branch is feasible: continue with it directly
+            C.withBackend (st ^. C.stateContext) $ \bak -> do
+              pnot <- WI.notPred sym p
+              let pf2' = C.forgetPostdomFrame falseFrame
+              CB.addAssumption bak (CB.BranchCondition loc (C.pausedLoc pf2') pnot)
+              let ctx = st ^. C.stateTree . C.actContext
+              C.ExecutionFeatureNewState <$> runReaderT (C.resumeFrame pf2' ctx) st
```
RyanGlScott: The `NoBranch True` and `NoBranch False` cases are nearly the same, but with minor differences in what predicate is used. Consider factoring out the code shared in common?

grease-exe/tests/llvm/uninit-stack-conditional.llvm.cbl
```diff
@@ -22,3 +22,8 @@
 ;; flags {"--path-strategy", "dfs"}
 ;; go(prog)
 ;; uninit_stack()
+
+;; flags {"--symbol", "test"}
+;; flags {"--path-strategy", "dfs"}
```
RyanGlScott, line 27: The `test` function is already tested above using `--path-strategy=dfs`. Did you mean to use `bfs` here?

langston-barrett: Good catch, thanks!

PR #547, opened 2026-03-06
Handle constant offsets

### LINKED ISSUE #3: Growth heuristic should account for the read/write offset
@langston-barrett says:

-----

This program exceeds the iteration limit:
```c
int test(int *x) {
  return x[4096];
}
```
There are two reasonable behaviors on this program, both of which rely on the heuristics taking into account the constant offset:

1. Growing the allocation in one step of the loop
2. Saying that the resulting allocation would be too large, and giving up

langston-barrett:
Fixes #3.

### DIFF
--- a/grease/src/Grease/Heuristic.hs
+++ b/grease/src/Grease/Heuristic.hs
@@ -216,65 +216,66 @@ handleMemErr la sym argNames args err sel =
     -- Note [Initializing empty pointer shapes] in Grease.Setup, this scenario
     -- can actually happen when refining a function pointer argument to a
     -- function.)
-    (AnyMemErrorLLVM (Mem.MemoryError _op (Mem.BadFunctionPointer fle)))
-      | fle /= Mem.RawBitvector ->
+    (AnyMemErrorLLVM (Mem.MemoryError op reason))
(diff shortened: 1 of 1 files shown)


grease/src/Grease/Heuristic.hs
```diff
@@ -206,12 +212,53 @@ handleMemErr la argNames args err sel =
     (AnyMemErrorLLVM (Mem.MemoryError _op (Mem.BadFunctionPointer fle)))
       | fle /= Mem.RawBitvector ->
           pure Unknown
+    AnyMemErrorLLVM (Mem.MemoryError op _reason) -> do
```
RyanGlScott: Style nit pick: I generally prefer to keep nearly-alike cases like these combined so that it's easier to tell at a glance where all the handling for the `AnyMemErrorLLVM` case is. That is, I would prefer writing this like:

```hs
    AnyMemErrorLLVM (Mem.MemoryError op reason)
      | Mem.BadFunctionPointer fle <- reason,
        fle /= Mem.RawBitvector ->
          pure Unknown
      | otherwise -> do
          ...
```

PR #548, opened 2026-03-06
Don't annotate block numbers

### LINKED ISSUE #45: Don't annotate block numbers
@langston-barrett says:

-----

Before analyzing a function, GREASE creates values to put in arguments/registers. GREASE annotates these values to track their provenance, i.e., which part of the arguments/registers they came from:

```hs
-- | Track the provenance of symbolic values using 'W4.SymAnnotation's.
data Annotations sym ext argTys
  = Annotations
    { _baseAnns :: MapF.MapF (W4.SymAnnotation sym) (SomeBaseSelector ext argTys)
    , _blockAnns :: Map (IntegerAnn sym) (Some (SomePtrSelector ext argTys))
    , _offsetAnns :: MapF.MapF (BVAnn sym) (SomePtrSelector ext argTys)
    }
```

This provenance information is later used in the heuristics to refine the precondition should these values appear in a memory error or similar.

For fresh constants, [annotation is a no-op](https://github.com/GaloisInc/what4/pull/245). Each What4 bound variable already carries a `Nonce` that uniquely identifies it, and this is returned as its annotation. In contrast, when annotating pointers, we annotate the block number, which is always a concrete natural (as returned by Crucible's `doMalloc`). I propose we stop annotating pointer block numbers, and instead use the concrete block number *itself* to track provenance.

To see that this would work, consider the requirements for provenance tracking:

1. Each symbolic expression generated during `Setup` should have a unique identifier (currently the annotation). No value generated during symbolic execution should have the same identifier.
2. That unique identifier should be maintained as a subexpression of any expression that uses a `Setup`-generated value during symbolic execution.

These are currently maintained as follows:

1. Each bound variable has a `Nonce`. Any new bound variables will have a fresh, unrelated `Nonce`. Pointer blocks get annotated, attaching a `Nonce` from the same generator used for bound variables (i.e., it is also always fresh).
2. Bound variable expressions are (almost) never simplified away, because it's very hard to rewrite anything involving an unknown constant. Thus, the bound variables remain subexpressions of almost anything that uses them. For pointer blocks, `Annotation` expressions work basically exactly the same way.

Now, consider how these would be maintained under the new scheme of using block numbers themselves to track provenance:

1. Block numbers allocated by `malloc` are always fresh, there is no way for the target program to reuse a previous block number
2. Pointer blocks [are only ever muxed](https://github.com/GaloisInc/crucible/issues/404), they are never manipulated using arithmetic or any other expression formers. Thus, any pointer *p* derived from a `Setup`-created pointer *q* will always contain *q*'s block number as a subexpression of its own.

Perhaps it's not *necessary* to annotate block numbers, but is it worth making this change? I'd argue that it is. Keeping block numbers concrete will:

- likely speed up memory model operations by hitting cases optimized for constants (or muxes of constants)
- simplify the debug output (including memory model printouts)
- reduce our space usage (though likely not by a huge amount)
- speed up refinement by keeping block number annotations in an `IntMap` instead of a `MapF.Map` (also probably not a huge win)

langston-barrett:
Fixes #45.

### DIFF
--- a/grease-exe/tests/refine/pos/global_var/test.c
+++ b/grease-exe/tests/refine/pos/global_var/test.c
@@ -19,4 +19,13 @@ void test(int x) {
   }
 }
 
-// all: ok()
+// arm: ok()
+// x64: ok()
+
+// On PPC32, the memory model produces an ITE over a function-backed array for
+// the global memory state after the first branch. Yices cannot handle this
+// expression type, so the goal is reported as a possible bug rather than being
+// proved. This would work with another solver (e.g., it works with Z3).
+//
+// ppc32: check "we do not support if/then/else expressions at type function-backed array with solver Yices."
+// ppc32: could_not_infer()
--- a/grease-exe/tests/llvm/func-ptr.llvm.cbl
+++ b/grease-exe/tests/llvm/func-ptr.llvm.cbl
@@ -14,7 +14,7 @@
 ;; go(prog)
 ;; check "Goal failed:"
 ;; check "Failed to load function handle"
-;; check "Cannot resolve a symbolic pointer to a function handle"
+;; check "The given pointer could not be resolved to a callable function"
 ;; no_heuristic()
 
 (defun @test ((p (Ptr 64))) Unit
--- a/grease/src/Grease/LLVM/SimulatorHooks.hs
+++ b/grease/src/Grease/LLVM/SimulatorHooks.hs
@@ -27,9 +27,10 @@ import Lang.Crucible.LLVM.Extension qualified as CLLVM
 import Lang.Crucible.LLVM.MemModel qualified as CLM
 import Lang.Crucible.LLVM.MemModel.Pointer qualified as CLMP
 import Lang.Crucible.Simulator qualified as CS
+import Lang.Crucible.Simulator.ExecutionTree (stateGlobals, stateSymInterface)
+import Lang.Crucible.Simulator.GlobalState qualified as CGS
 import Lumberjack qualified as LJ
 import What4.FunctionName qualified as WFN
-import What4.Interface qualified as WI
 
 doLog :: MonadIO m => GreaseLogAction -> Diag.Diagnostic -> m ()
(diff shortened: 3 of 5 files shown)


grease-exe/tests/refine/pos/global_var/test.c
```diff
@@ -19,4 +19,13 @@ void test(int x) {
...
+// arm: ok()
+// x64: ok()
+
+// On PPC32, the memory model produces an ITE over a function-backed array for
+// the global memory state after the first branch. Yices cannot handle this
+// expression type, so the goal is reported as a possible bug rather than being
+// proved.
+//
+// ppc32: check "we do not support if/then/else expressions at type function-backed array with solver Yices."
+// ppc32: could_not_infer()
```
langston-barrett, line 31: This seems bad, but it's actually good: it shows that using unannotated block numbers is actually causing us to hit code paths in the memory model for concrete integers, which should generally be faster.

RyanGlScott: Is this problem specific to Yices? That is, would this be `ok()` if you used a different solver?

langston-barrett: Yes, that's right:
```
cabal run exe:grease -- --symbol test --solver z3 tests/refine/pos/global_var/test.ppc32.elf


Configuration is affected by cabal.project and cabal.project.local at
'/Users/langston/code/grease'.
All goals passed!
Finished analyzing 'test'.
```

RyanGlScott: Huh, interesting. I am a bit surprised that Yices _can_ handle the corresponding proof goal when the block numbers are symbolic.

I don't think we need to hold up this PR this, but I wonder if enabling `useSymbolicArrays` in `what4` would help here (see https://github.com/GaloisInc/what4/issues/157).

langston-barrett: https://github.com/GaloisInc/grease/issues/559

PR #550, opened 2026-03-07
llvm: Overrides for networking

### LINKED ISSUE #387: io: Support socket networking overrides for LLVM programs
#386 adds built-in overrides for some socket-based networking APIs. These overrides are currently unavailable when analyzing LLVM code, but there's no fundamental reason for that. Let's make them available for LLVM, too!

langston-barrett:
Fixes #387.

### DIFF
--- a/grease/src/Grease/LLVM/Personality.hs
+++ b/grease/src/Grease/LLVM/Personality.hs
@@ -34,7 +34,7 @@ data GreaseLLVMPersonality cExt sym t
   , _serverSocketFds :: Map.Map Integer (Some GSN.ServerSocketInfo)
   -- ^ A map from registered socket file descriptors to their corresponding
   -- metadata. See @Note [The networking story]@ in
-  -- "Grease.Macaw.Overrides.Networking".
+  -- "Grease.Overrides.Networking".
   }
 
 makeLenses ''GreaseLLVMPersonality
--- a/grease/src/Grease/LLVM/SetupHook.hs
+++ b/grease/src/Grease/LLVM/SetupHook.hs
@@ -48,6 +48,25 @@ import What4.FunctionName qualified as WFN
 doLog :: MonadIO m => GreaseLogAction -> Diag.Diagnostic -> m ()
 doLog la diag = LJ.writeLog la (LLVMSetupHookDiagnostic diag)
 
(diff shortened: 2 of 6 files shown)


grease/src/Grease/LLVM/SetupHook.hs
```diff
@@ -88,9 +91,13 @@ syntaxSetupHook la ovs skipFuns prog cfgs errCb =
     let dl = CLTC.llvmDataLayout typeCtx
     let mvar = CLT.llvmMemVar llvmCtx
 
-    -- Register built-in and user overrides.
+    -- Register built-in, networking, and user overrides.
+    let networkOvTemplates =
+          fmap
+            (\(CLI.SomeLLVMOverride ov) -> CLI.basic_llvm_override ov)
+            (Seq.fromList (networkLLVMOverrides bak fs mvar))
```
RyanGlScott: It might be worth factoring out this sub-expression into its own definition, as it appears verbatim in `moduleSetupHook` below as well. If we add more custom overrides in the future, I imagine that we'll want to avoid repeating ourselves.

grease/src/Grease/LLVM/Personality.hs
```diff
@@ -28,6 +31,10 @@ data GreaseLLVMPersonality cExt sym t
   , _toConcretize :: CS.GlobalVar ToConc.ToConcretizeType
   -- ^ Values created during runtime to be passed to the concretization
   -- functionality and generally displayed to the user.
+  , _serverSocketFds :: Map.Map Integer (Some GSN.ServerSocketInfo)
+  -- ^ A map from registered socket file descriptors to their corresponding
+  -- metadata. See @Note [The networking story]@ in
+  -- "Grease.Macaw.Overrides.Networking".
```
RyanGlScott: I think it would make more sense for this Note to live in the new `Grease.Overrides.Networking` module, as much of its contents aren't specific to `macaw`.

langston-barrett: Good call!

grease/src/Grease/LLVM/SetupHook.hs
langston-barrett: Done!

PR #553, opened 2026-03-07
More overrides for concretization

### LINKED ISSUE #382: Add built-in overrides for concretization
It would be useful to have built-in overrides that concretize symbolic values. Using https://github.com/GaloisInc/what4/pull/264, we could provide overrides that query an SMT solver to show that a symbolic value is actually constrained to a particular concrete value and return the latter. We may also want to provide overrides that simply return *some* feasible concrete value from a model.

langston-barrett:
Fixes #382.

### DIFF
--- a/doc/sexp.md
+++ b/doc/sexp.md
@@ -27,17 +27,37 @@ additional types and operations, see:
 There are a few overrides that are only available in S-expression files
 (programs or overrides).
 ```
-(declare @conc-bv-8 ((bv (Bitvector 8))) (Vector (Bitvector 8)))
-(declare @conc-bv-16 ((bv (Bitvector 16))) (Vector (Bitvector 16)))
-(declare @conc-bv-32 ((bv (Bitvector 32))) (Vector (Bitvector 32)))
-(declare @conc-bv-64 ((bv (Bitvector 64))) (Vector (Bitvector 64)))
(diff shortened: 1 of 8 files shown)


doc/sexp.md
```diff
@@ -27,17 +27,37 @@ additional types and operations, see:
...
+(declare @conc-ptr-64 ((p (Ptr 64))) (Ptr 64))
+(declare @conc-vector-bv-8 ((v (Vector (Bitvector 8)))) (Vector (Bitvector 8)))
+(declare @conc-vector-bv-16 ((v (Vector (Bitvector 16)))) (Vector (Bitvector 16)))
+(declare @conc-vector-bv-32 ((v (Vector (Bitvector 32)))) (Vector (Bitvector 32)))
+(declare @conc-vector-bv-64 ((v (Vector (Bitvector 64)))) (Vector (Bitvector 64)))
 (declare @fresh-bytes ((name (String Unicode)) (num (Bitvector w))) (Vector (Bitvector 8)))
 ```
 
-### `@conc-bv-*`
+### `@conc-bool`, `@conc-integer`, `@conc-nat`, `@conc-bv-*`
```
RyanGlScott, line 46: And `conc-vector-bv-*`, presumably?

langston-barrett: That's covered below!

PR #565, opened 2026-03-10
cli: Add `--arg-str` and `--arg-sym-str` flags

### LINKED ISSUE #479: cli: `--arg-str`
As a follow-up to #228, we should have a command-line flag for specifying that a register holds a null-terminated string.

langston-barrett:
Adds two new CLI flags for working with null-terminated strings:

- --arg-str ARG:STRING: Initializes argument ARG to a pointer to a concrete null-terminated string with the given content. For example, --arg-str rdi:hello creates a pointer to the bytes 'h','e','l','l','o',0.

- --arg-sym-str ARG:N: Initializes argument ARG to a pointer to a symbolic null-terminated string buffer of N bytes (including the null terminator). This is useful for testing string-handling code with symbolic inputs.

Closes #479

### DIFF
--- a/grease-cli/src/Grease/Cli.hs
+++ b/grease-cli/src/Grease/Cli.hs
@@ -229,6 +229,18 @@ simpleShapesParser = fmap Map.fromList $ Opt.many $ do
             <> Opt.metavar "ARG:N"
             <> Opt.help "initialize argument ARG to an unsigned 64-bit integer"
         )
+    , Opt.option
+        (withArgName Simple.parseArgStr)
+        ( Opt.long "arg-str"
+            <> Opt.metavar "ARG:STRING"
+            <> Opt.help "initialize argument ARG to a pointer to a concrete, null-terminated, UTF-8 string"
+        )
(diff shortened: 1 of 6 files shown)


grease-cli/src/Grease/Cli.hs
```diff
@@ -229,6 +229,18 @@ simpleShapesParser = fmap Map.fromList $ Opt.many $ do
             <> Opt.metavar "ARG:N"
             <> Opt.help "initialize argument ARG to an unsigned 64-bit integer"
         )
+    , Opt.option
+        (withArgName Simple.parseArgStr)
+        ( Opt.long "arg-str"
+            <> Opt.metavar "ARG:STRING"
+            <> Opt.help "initialize argument ARG to a pointer to a concrete null-terminated string"
```
RyanGlScott: It would be worth noting here that `STRING` must be a UTF-8-encoded string (i.e., you can't just pass arbitrary bytes).

RyanGlScott (approved):

PR #580, opened 2026-03-19
Precompute names and types of registers

langston-barrett:
Instead of having the computations of names/types of registers scattered all over the place, precompute them once and reuse them everywhere. Makes things simpler and perhaps marginally more performant.

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -206,4 +206,5 @@
     - { name: Data.Parameterized.TraversableFC, as: TFC }
 
     # grease
+    - { name: Grease.Macaw.Arch, as: Arch }
     - { name: Grease.Scheduler, as: Sched }
--- a/grease-exe/src/Grease/Main.hs
+++ b/grease-exe/src/Grease/Main.hs
@@ -110,7 +110,7 @@ import Grease.LLVM.SetupHook qualified as LLVM (SetupHook, moduleSetupHook, synt
 import Grease.LLVM.SetupHook.Diagnostic qualified as LDiag (Diagnostic (LLVMTranslationWarning))
 import Grease.LLVM.Shapes qualified as GLS
 import Grease.Macaw qualified as GM
-import Grease.Macaw.Arch qualified as GMA
+import Grease.Macaw.Arch qualified as Arch
 import Grease.Macaw.Arch.AArch32 (armCtx)
 import Grease.Macaw.Arch.PPC32 (ppc32Ctx)
 import Grease.Macaw.Arch.PPC64 (ppc64Ctx)
@@ -448,7 +448,7 @@ data MacawCfgConfig arch = MacawCfgConfig
   -- ^ Map of addresses to PLT stub names.
   , mcDynFunMap :: Map.Map WFN.FunctionName (MC.ArchSegmentOff arch)
   -- ^ Map of dynamic function names to their addresses.
-  , mcRelocs :: Map.Map (MM.MemWord (MC.ArchAddrWidth arch)) (GMA.ArchReloc arch)
+  , mcRelocs :: Map.Map (MM.MemWord (MC.ArchAddrWidth arch)) (Arch.ArchReloc arch)
   -- ^ Map of relocation addresses and types.
   , mcMemory :: MC.Memory (MC.RegAddrWidth (MC.ArchReg arch))
   -- ^ The memory layout of the binary.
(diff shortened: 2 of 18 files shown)


grease/src/Grease/Macaw/Arch.hs
```diff
@@ -283,6 +308,29 @@ defaultPCFixup ::
...
+  , C.CtxRepr (Ctx.EmptyCtx Ctx.::> Symbolic.ArchRegStruct arch)
+  )
+mkArchRegInfo avals =
+  let archFns = Symbolic.archFunctions avals
+      regTypes = Symbolic.crucArchRegTypes archFns
+      regStructType = C.StructRepr regTypes
+      regNames =
+        Symbolic.macawAssignToCruc (Const . mkRegName) $
+          Symbolic.crucGenRegAssignment archFns
+      valueNames = fmapFC (\(Const (RegName n)) -> ValueName n) regNames
```
RyanGlScott: If I'm reading this code correctly, the only differences between `regNames` and `valueNames` is the particular choice of newtype applied on top of the underlying `String`. To simplify thing, do you think it would make sense to only store one or the other, and then add a helper function which retrieves a `ValueName` by converting it from a `RegName` (or vice versa)?

grease/src/Grease/Macaw/Arch.hs
```diff
@@ -283,6 +308,29 @@ defaultPCFixup ::
...
+  Symbolic.SymArchConstraints arch =>
+  Symbolic.GenArchVals Symbolic.LLVMMemory arch ->
+  ( C.TypeRepr (Symbolic.ArchRegStruct arch)
+  , Ctx.Assignment C.TypeRepr (Symbolic.MacawCrucibleRegTypes arch)
+  , Ctx.Assignment (Const RegName) (Symbolic.MacawCrucibleRegTypes arch)
+  , Ctx.Assignment ValueName (Symbolic.MacawCrucibleRegTypes arch)
+  , C.CtxRepr (Ctx.EmptyCtx Ctx.::> Symbolic.ArchRegStruct arch)
+  )
+mkArchRegInfo avals =
+  let archFns = Symbolic.archFunctions avals
```
RyanGlScott, line 324: Truth be told, I'm not sure that I see much benefit in storing this. This is just a simple record field access into `avals :: GenArchVals` (i.e., it's cheap to compute), and we already store `avals` in `ArchContext`, so it seems just as direct to retrieve it that way.

langston-barrett: Right, `archFns` aren't stored as a separate field of `ArchContext` if that's what you were thinking - they are just used to compute `crucArchRegTypes`, which is stored and also used to compute `regStructType`. The computation of `crucArchRegTypes` is pretty trivial, so we could definitely make that an accessor. However, I'd like to keep `regStructType`, since it does involve allocation. What do you think?

RyanGlScott: I'm fine with keeping `regStructType` as a separate field, to be clear. It's specifically `archFunctions` that seems odd to me—you could just as well use `archFunctions (archContext ^. archVals)`, which is trivial to compute. (We could even introduce a helper function for going from `ArchContext` to `MacawSymbolicArchFunctions` if that is a common pattern.)

Relatedly, I wondered if we really do need to store the entirety of `GenArchVals` via the `_archVals` field of `ArchContext` now that we're caching more of its subcomponents, but a quick search through the `grease` codebase suggests that it wouldn't be straightforward to remove `_archVals`. As such, it's probably best to keep it.

langston-barrett: AFAICT we're not storing `archFunctions` as a field of `ArchContext`, am I not understanding your concern here?

RyanGlScott: [`archFunctions`](https://github.com/GaloisInc/macaw/blob/73e42d7a539c6f8f3470345043bf5c2481e581fb/symbolic/src/Data/Macaw/Symbolic.hs#L274) is a record field of `GenArchVals`, and we are storing a `GenArchVals` in `ArchContext` by way of the `_archVals` field.

langston-barrett: Right, but `_archVals` has been around for a long time, right? Are you asking for changes to the code here, or just thinking longer term that we might want to get rid of that?

Sorry, I'm not trying to be obtuse, I'm just not sure what to change.

RyanGlScott: Oh, my apologies, I assumed that this `archFn`s value was being stored as an `ArchContext` field (given it was listed next to other things that are destined to become fields). Now that I've read this more closely, that's not at all the case. Apologies for the confusion on my part!

langston-barrett: No worries at all!

grease/src/Grease/Macaw/Arch.hs
langston-barrett: Done!

langston-barrett: I actually couldn't get this working with `coerce`, so I'll just keep the precomputation here if that's alright.

RyanGlScott: What was the issue you encountered?

langston-barrett: I don't exactly recall now, but I've filed an issue to circle back: https://github.com/GaloisInc/grease/issues/590

PR #581, opened 2026-03-19
screach: Minimize data saved for verification

langston-barrett:
The `ExecState` inside `SavedState` was super heavy-weight, and was
unused except to extract the `RefinementResult`! This should be simpler
and considerably less memory-intensive. The saved backend state was
completely unused.

### DIFF
--- a/screach/src/Screach/RefineFeature.hs
+++ b/screach/src/Screach/RefineFeature.hs
@@ -19,13 +19,9 @@ module Screach.RefineFeature (
   refineErrMap,
   HasScreachPersonality,
   sdseExecFeatures,
-
-  -- * Saved target results
-  SavedState (..),
-  createSaveItem,
 ) where
 
-import Control.Lens ((&), (.~), (?~))
+import Control.Lens ((&), (.~), (?~), (^.))
 import Control.Lens qualified as Lens
 import Control.Monad.IO.Class (MonadIO)
 import Data.Data (Proxy (Proxy))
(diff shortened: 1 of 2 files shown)


screach/src/Screach/RefineFeature.hs
```diff
@@ -370,22 +341,22 @@ sdseExecFeatures ::
...
   let satPolicy =
         Sched.withSatisfiabilityCheck (CBO.considerSatisfiability bak) $
           Sched.priorityPolicy priorityFunc
   let resultCallback r = do
         t <- isTargetPred r
         if t
           then do
-            saved <- createSaveItem r
-            modifyIORef' savedRef (saved :)
+            let rst = (C.execResultContext r ^. C.cruciblePersonality . refinementState :: SrchRefineData sym bak scope ext tys w)
```
RyanGlScott: This line is rather long. Consider converting the inline type annotation to a top-level type signature for `rst`?

screach/src/Screach/RefineFeature.hs
```diff
@@ -370,22 +341,22 @@ sdseExecFeatures ::
...
+            case _refineResult rst of
```
RyanGlScott: Since we have a lens, we may as well use it (`rst ^. refineResult`).

langston-barrett: Type inference was broken, hence the signature above and accessor here...

PR #587, opened 2026-03-20
ci: Collect and display deterministic performance metrics

### LINKED ISSUE #584: ci: Collect determinisitic performance numbers
I've been working a lot on GREASE's performance. I usually focus on deterministic performance numbers, like those from [GHC's `-s` RTS flag](https://downloads.haskell.org/ghc/latest/docs/users_guide/runtime_control.html#rts-flag-s-file), or instruction counts from cachegrind. I'd like to collect these in CI for our biggest test (the libpng example) and display them on the job summary page (as is currently done for coverage). Just like that report, all of the logic of collecting this data should happen in an external script. In this case, it's probably worth using Python with type signatures, since we need to scrape the output. The script should only collect textual data (no `hp2ps`), and only deterministic metrics. This will allow us to compare performance improvements from branches to `main`.

langston-barrett:
Fixes #584.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,6 +141,19 @@ jobs:
     - name: Test screach
       run: cabal test pkg:screach
 
+    - name: Collect performance metrics
+      if: ${{ strategy.job-index == 0 }}
+      run: |
+        python3 scripts/perf.py \
+          --package exe:grease \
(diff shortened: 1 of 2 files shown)


langston-barrett:
Or more recently, here: https://github.com/GaloisInc/grease/actions/runs/23358817893?pr=587

Issue #21, opened 2025-03-11
Support simulating shared libraries


RyanGlScott:
`grease` currently supports dynamically linked binaries, but only in the sense that it can simulate the code in the main binary itself. Any attempts to call out to a shared library via a PLT stub, however, will cause `grease` to simply skip over the function. This is fine for many purposes, but there may come a day when we want to run `grease` on the code in the shared libraries themselves. This issue exists to track implementing this idea.

We can take inspiration from how `ambient-verifier` supports shared libraries: load all the code from the main binary plus its shared library dependencies and place it all in a single global address space, where the most significant bits of each address indicate which binary the code comes from. For instance, an address of `0x00000042` indicates that it comes from the main binary, an address of `0x10000042` indicates that it comes from the first shared library dependency, `0x20000042` the second shared library dependency, and so on. This allows `macaw-symbolic` to operate on all the code without too many drastic changes. One additional thing `grease` would need to do is to change the behavior of invoking PLT stubs, as these would need to call the relevant functions in shared libraries instead of skipping.

There are some design questions to answer:

* How should users be expected to specify the locations of shared libraries? `ambient-verifier` requires users to put all of the shared libraries in a single directory and pass the name of the directory as a command-line argument. This avoids any ambiguity about where shared libraries live, but at the expense of requiring users to do more work. As an alternative, we could rely on the ELF information to figure out where the shared libraries live.
* Should `grease` _always_ simulate code in shared libraries? Much of the time, the code in shared libraries such as `libc` is very difficult to symbolically execute, and attempting to do so can frequently lead to infinite loops. Perhaps it might be a better policy to skip simulating shared library code by default unless a user explicitly requests to do so.

PR #589, opened 2026-03-23
tests: xfail tests for supporting shared libraries

langston-barrett:
xfail tests for #21. Since the PR fixing that will likely be large, I thought it might be helpful to have these as a pre-PR, where we specify what we want the behavior to be before implementing.


grease-exe/tests/feat/dyn/shared-lib-as-main.lua
```diff
@@ -0,0 +1,100 @@
+-- Test simulating a shared library directly as the main target (not via an
+-- executable). Verifies that grease can analyze shared libraries with
+-- --shared-lib-dir to resolve their own dependencies.
+
+---- x64 ----
+
+-- Run 1: libfoo.so as main target (no dependencies)
+flags {"--symbol", "foo"}
+go "tests/feat/dyn/libfoo.x64.so"
+ok()
```
langston-barrett, line 10: This test actually has immediate value: we support executing `.so`s, but AFAIK don't have an existing test for it.

RyanGlScott: There is also `tests/sanity/pass/simulate_from_shared_library`, for what it's worth.

langston-barrett: Oh, nice!

Issue #446, opened 2025-10-29
Featurize GREASE


2over12:
It should be possible to bring the design of GREASE inline with downstream consumers. Instead of an extrinsic refinement loop, GREASE can become a feature in the sequence of crucible `executeCrucible`. Doing this opens up the possibility of alternative scheduling etc etc. To maintain GREASE's current behavior we could just replace the state wholesale with a fresh new initial state under the new ArgShape

PR #600, opened 2026-03-25
Common type for shared personality fields

langston-barrett:
Another refactoring to prepare for #446.

Next steps will be:

- Adding LLVM memory var to `Personality`
- Adding `RefineData` to `Personality`
- Moving the map `IORef` into `RefineData`

### DIFF
--- a/grease/src/Grease/Personality.hs
+++ b/grease/src/Grease/Personality.hs
@@ -1,0 +1,71 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FunctionalDependencies #-}
+{-# LANGUAGE TemplateHaskell #-}
+
+-- |
+-- Copyright        : (c) Galois, Inc. 2026
+-- Maintainer       : GREASE Maintainers <grease@galois.com>
+--
+-- The shared parts of all @grease@ personality types. Both the Macaw personality
+-- ('Grease.Macaw.SimulatorState.GreaseSimulatorState') and the LLVM personality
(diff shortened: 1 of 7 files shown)


grease/src/Grease/Personality.hs
```diff
@@ -0,0 +1,71 @@
...
+  Type ->
+  CrucibleType ->
+  Type
+data Personality cExt sym ext ret = Personality
+  { _pDbgContext :: Dbg.Context cExt sym ext ret
+  -- ^ The debugger context for interactive debugging.
+  , _pToConcretize :: CS.GlobalVar ToConc.ToConcretizeType
+  -- ^ Values created during runtime to be passed to the concretization
+  -- functionality and generally displayed to the user.
+  , _pServerSocketFds :: Map.Map Integer (Some GSN.ServerSocketInfo)
```
RyanGlScott, line 57: I imagine that this `Map` will almost always be empty when you first create a `Personality` value, but at the moment, you have to type `Map.empty` as boilerplate code. Perhaps it would be worth adding a function of type `Context cExt sym ext ret -> GlobalVar ToConcreteize -> Personality cExt sym ext ret` that initializes the `_pServerSocketFds` field with `Map.empty`? (Perhaps calling it `mkPersonality`, for symmetry with `mkGreaseSimulatorState`)

RyanGlScott (approved):

grease/src/Grease/Personality.hs
langston-barrett: Will do in a follow-up!

PR #602, opened 2026-03-25
Store the LLVM memory variable in the `Personality`

langston-barrett:
Yet another step towards #446.

### DIFF
--- a/grease/src/Grease/LLVM/SetupHook.hs
+++ b/grease/src/Grease/LLVM/SetupHook.hs
@@ -61,12 +61,11 @@ networkOvTemplates ::
   ) =>
   bak ->
   CLSIO.LLVMFileSystem w ->
-  CS.GlobalVar CLM.Mem ->
   Seq.Seq (CLI.OverrideTemplate p sym LLVM arch)
-networkOvTemplates bak fs mvar =
+networkOvTemplates bak fs =
   fmap
     (\(CLI.SomeLLVMOverride ov) -> CLI.basic_llvm_override ov)
-    (Seq.fromList (networkLLVMOverrides bak fs mvar))
+    (Seq.fromList (networkLLVMOverrides bak fs))
 
 -- | Hook to run before executing a CFG.
 --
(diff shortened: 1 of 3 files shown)


grease/src/Grease/Personality.hs
```diff
@@ -62,6 +74,43 @@ data Personality cExt sym ext ret = Personality
...
+    , _pDbgContext = dbgCtx
+    , _pToConcretize = toConc
+    , _pServerSocketFds = Map.empty
+    }
+
+-- | A class for personality types that contain an LLVM memory model
+-- 'CS.GlobalVar'.
+class HasMemVar p where
+  getMemVar :: p -> CS.GlobalVar CLM.Mem
+
```
RyanGlScott, line 96: I wonder if there should be a superclass relationship between `HasMemVar`, `HasPersonality`, `HasGreaseLLVMPersonality`, and `HasGreaseSimulatorState`. Perhaps:

```hs
class HashMemVar p => HasPersonality p cExt sym ext ret | p -> cExt sym ext ret

class HasPersonality p cExt sym LLVM ret => HasGreaseLLVMPersonality p cExt sym ret | p -> cExt sym ret

class
  ( HasMacawLazySimulatorState p sym (ArchAddrWidth arch)
  , HasServerSocketFds p
  , HasPersonality p cExt sym (MacawExt arch) ret
  ) =>
  HasGreaseSimulatorState p cExt sym arch ret | p -> cExt sym arch ret
```

grease/src/Grease/Macaw.hs
```diff
@@ -516,13 +517,13 @@ initState ::
   , ?memOpts :: CLM.MemOptions
   , HasGreaseSimulatorState p cExt sym arch ret
   , HasToConcretize p
+  , GP.HasMemVar p
```
RyanGlScott: ...if this superclass relationship existed, then it would suffice to just have a `HasGreaseSimulatorState` constraint here instead of needing both `HasGreaseSimulatorState` _and_ `HasMemVar` (the latter of which feels like it ought to be implied by the former).

grease-exe/src/Grease/Main.hs
```diff
@@ -673,6 +673,7 @@ macawMemConfig ::
   , ?lc :: CLTC.TypeContext
   , GMSS.HasGreaseSimulatorState p cExt sym arch ret
   , ToConc.HasToConcretize p
+  , GP.HasMemVar p
```
RyanGlScott: Similarly, the `HasMemVar` constraint here could be made redundant.

grease/src/Grease/Macaw/Overrides.hs
```diff
@@ -209,6 +212,7 @@ mkMacawOverrideMapWithBuiltins ::
   , CLM.HasPtrWidth (MC.ArchAddrWidth arch)
   , Symbolic.SymArchConstraints arch
   , HasGreaseSimulatorState p cExt sym arch ret
+  , GP.HasMemVar p
```
RyanGlScott: Similarly, the `HasMemVar` constraint here could be made redundant.

grease/src/Grease/Macaw/SetupHook.hs
```diff
@@ -64,9 +65,9 @@ newtype SetupHook sym arch
         , CLM.HasLLVMAnn sym
         , HasGreaseSimulatorState p cExt sym arch ret
         , HasToConcretize p
+        , GP.HasMemVar p
```
RyanGlScott: Similarly, the `HasMemVar` constraint here could be made redundant.

grease/src/Grease/LLVM/SetupHook.hs
RyanGlScott: I'd be surprised if this explicit `GlobalVar Mem` argument was still necessary. Could this be replaced with a `HasMemVar p` constraint?

(Similarly for other places where `GlobalVar Mem`s are passed around explicitly.)

langston-barrett: I've removed some more places where it's passed explicitly, but doing *all* of them might take a while. Let's do it gradually!

PR #604, opened 2026-03-26
Store `RefineData` in the personality

langston-barrett:
Another step towards featurization of GREASE (#446).

- Add `refineErrMap` to `RefineData`
- Put `RefineData` in `Personality`
- Reorder some initialization to make this work
- Update refinement in both codebases to use all this

### DIFF
--- a/grease/src/Grease/Macaw/SimulatorState.hs
+++ b/grease/src/Grease/Macaw/SimulatorState.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE ExplicitNamespaces #-}
 {-# LANGUAGE FunctionalDependencies #-}
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE UndecidableInstances #-}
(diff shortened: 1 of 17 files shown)


grease/src/Grease/Macaw/SimulatorState.hs
```diff
@@ -50,17 +52,25 @@ import Stubs.Syscall qualified as Stubs
...
   -- @ret@: the Crucible return type (usually @'Symbolic.ArchRegStruct' arch@)
   CT.CrucibleType ->
+  -- @argTys@: Crucible argument types for the target function
+  CT.Ctx CT.CrucibleType ->
+  -- @wptr@: pointer width
+  Natural ->
   Type
-data GreaseSimulatorState cExt sym arch ret = GreaseSimulatorState
-  { _gssPersonality :: GP.Personality cExt sym (Symbolic.MacawExt arch) ret
+data GreaseSimulatorState sym bak t cExt arch ret argTys wptr = GreaseSimulatorState
```
RyanGlScott, line 72: We're going for a world record in number of type parameters!

langston-barrett: 🚀

Issue #596, opened 2026-03-24
cli: Configurable pointer concretization


langston-barrett:
We currently do not concretize ("resolve") pointers. We should make this behavior configurable with a `--pointer-concretization` flag with the following options:

1. none (current, default) - no pointer concretization
2. unsound - Fast concretization that picks an arbitrary valid value (1 solver query)
3. unique - Sound concretization using Crucible's `uniquelyConcRegValue` (2 queries)
4. resolve - Sound concretization with bound narrowing using macaw's `resolveLLVMPtr`

PR #611, opened 2026-03-27
Overrides for unique concretization

langston-barrett:
I need this for testing #596.

### DIFF
--- a/grease/src/Grease/LLVM.hs
+++ b/grease/src/Grease/LLVM.hs
@@ -34,11 +34,16 @@ import Lang.Crucible.LLVM.Translation qualified as CLT
 import Lang.Crucible.LLVM.TypeContext qualified as CLTC
 import Lang.Crucible.Simulator qualified as CS
 import Lang.Crucible.Simulator.GlobalState qualified as CS
+import What4.Expr.Builder qualified as WEB
+import What4.FloatMode qualified as W4FM
 
 initState ::
-  forall p sym bak arch m st fs argTys retTy solver scope.
+  forall p sym bak arch m st fs fm argTys retTy solver scope.
   ( MonadIO m
   , MonadThrow m
+  , sym ~ WEB.ExprBuilder scope st fs
+  , fs ~ WEB.Flags fm
+  , C.KnownRepr W4FM.FloatModeRepr fm
   , OnlineSolverAndBackend solver sym bak scope st fs
   , 16 C.<= ArchWidth arch
   , ArchWidth arch ~ 64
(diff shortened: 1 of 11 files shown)


grease/src/Grease/LLVM.hs
```diff
@@ -34,11 +34,16 @@ import Lang.Crucible.LLVM.Translation qualified as CLT
...
 
 initState ::
-  forall p sym bak arch m st fs argTys retTy solver scope.
+  forall p sym bak arch m st fs fm argTys retTy solver scope.
   ( MonadIO m
   , MonadThrow m
+  , sym ~ WEB.ExprBuilder scope st fs
+  , fs ~ WEB.Flags fm
+  , C.KnownRepr W4FM.FloatModeRepr fm
   , OnlineSolverAndBackend solver sym bak scope st fs
```
RyanGlScott, line 47: This commit introduces quite a few additional constraints (e.g., `fs ~ Flags fm` and `KnownRepr FloatModeRepr fm`) that must be repeated in many places, and I wonder if we can reduce the amount of repetition. Some observations:

1. More often than not, these constraints appear alongside an `OnlineSolverAndBackend` constraint, suggesting that we may want to include these constraints in the definition of `OnlineSolverAndBackend`.
2. I am a bit surprised that we need a `KnownRepr FloatModeRepr fm` constraint, as `ExprBuilder` already stores a `FloatModeRepr` field anyway (see [here](https://github.com/GaloisInc/what4/blob/2ebbd6873ec1bf681ac7c8913ab20c2495b31d18/what4/src/What4/Expr/Builder.hs#L440-L442)). But now that I look more closely, it appears that this field is internal to `ExprBuilder`, and there is no function that provides access to it. That seems like a wasted opportunity---if we had such a function, then we could just extract the underlying `FloatModeRepr` without needing to pass around an additional constraint.

   In any case, this is something that would need to be changed on the `what4` side, so no need to do anything about this as part of this PR. It may be worth opening a `what4` issue, however.

langston-barrett: > More often than not, these constraints appear alongside an OnlineSolverAndBackend constraint, suggesting that we may want to include these constraints in the definition of OnlineSolverAndBackend.

That's a good idea, but there are some places that won't work, e.g., in stubs' [`mkFunctionOverride`](https://github.com/GaloisInc/stubs/blob/bd481bde583490f2e7faababbca2ac9f9f9a0054/stubs-common/src/Stubs/FunctionOverride.hs#L139-L147). So fully solving that would also require some upstream changes.

langston-barrett: For tracking the What4 change: https://github.com/GaloisInc/grease/issues/614

PR #615, opened 2026-03-27
Create `BinMd`

langston-barrett:
- Share code for computing things like PLT stub maps
- Share definition of empty loaded program
- Group a bunch of related arguments together

### DIFF
--- a/grease/src/Grease/Macaw/Load.hs
+++ b/grease/src/Grease/Macaw/Load.hs
@@ -10,6 +10,8 @@
 -- Copyright        : (c) Galois, Inc. 2024
 -- Maintainer       : GREASE Maintainers <grease@galois.com>
 module Grease.Macaw.Load (
+  BinMd (..),
+  emptyBinMd,
   LoadedProgram (..),
   LoadError (..),
   load,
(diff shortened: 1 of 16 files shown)


grease/src/Grease/Macaw/Load.hs
```diff
@@ -59,21 +68,44 @@ doLog la diag = LJ.writeLog la (LoadDiagnostic diag)
 
 data LoadedProgram arch
   = LoadedProgram
-  { progLoadedBinary :: Loader.LoadedBinary arch (Elf.ElfHeaderInfo (MC.ArchAddrWidth arch))
```
RyanGlScott: The name "`LoadedProgram`" is now a bit of a misnomer, given that it no longer contains the `LoadedBinary` (i.e., the program). Perhaps we should call this something like `LoadedBinaryMetadata` instead?

grease/src/Grease/Macaw/Load.hs
```diff
@@ -59,21 +68,44 @@ doLog la diag = LJ.writeLog la (LoadDiagnostic diag)
...
-  , progLoadOptions :: LC.LoadOptions
+  { progLoadOptions :: LC.LoadOptions
+  -- ^ The load options used to load addresses in the binary.
+  , progMemory :: MC.Memory (MC.RegAddrWidth (MC.ArchReg arch))
+  -- ^ The memory layout of the binary.
```
RyanGlScott: Since we no longer have a `LoadedBinary` field, I wonder if we should also remove the `progMemory` field as well. The `Memory` can trivially be obtained by calling `memoryImage` on the `LoadedBinary`, after all.

grease/src/Grease/Macaw/Load.hs
langston-barrett: Yeah good point. Will fix!

grease/src/Grease/Macaw/Load.hs
```diff
@@ -57,23 +65,52 @@ import What4.FunctionName qualified as WFN
...
   -- ^ The entrypoint addresses after resolving the user-supplied
   -- 'Entrypoint'.
   }
 
+-- | An empty 'BinMd' for use in paths that don't load a binary
+-- (e.g., Crucible S-expression programs).
+emptyBinMd ::
+  Arch.ArchContext arch ->
+  BinMd arch
+emptyBinMd _archCtx =
```
RyanGlScott: Is the `_archCtx` argument necessary anymore? I wouldn't think that it's needed to resolve ambiguous types, given that `arch` is a parameter to a data type (`BinMd`).

PR #617, opened 2026-03-30
Shadow stack

langston-barrett:
An execution feature for catching stack-based control-flow hijacks.


langston-barrett:
Closing until I have time to revisit.

PR #624, opened 2026-04-01
Add test for standalone version of zlib CVE-2022-37434

langston-barrett:


screach/test-data/zlib-cve-2022-37434.c
```diff
@@ -0,0 +1,2075 @@
+// CFLAGS: -O0
+
+// Standalone version of zlib CVE-2022-37434, see "BUG" below.
```
RyanGlScott, line 3: Was the original code obtained from somewhere on the Internet? If so, it might be nice to include it for traceability purposes.

screach/test-data/extra/zlib-cve-2022-37434.cbl
```diff
@@ -0,0 +1,14 @@
+; Check if rdx (memcpy size) has 32-bit sign bit set at 0x394b
+; At 0x394b: mov %rax,%rdx moves the ternary result to rdx (memcpy 3rd argument).
```
RyanGlScott: I'm not sure I understand the significance of the address 0x394b mentioned in this comment, especially since the CI setup uses the address 0x38d6 instead.

langston-barrett: Whoops, that's left over from an erlier compilation! Thanks

screach/test-data/zlib-cve-2022-37434.c
langston-barrett: Great point, will add.

langston-barrett:
Oh shoot, I clicked auto-merge and forgot about it. Ah, well - will work on the `SymSequence` fix right now.

PR #628, opened 2026-04-02
ci: Deterministic performance metrics for source-only benchmarks

langston-barrett:
More CI benching infrastructure to stress-test different parts of Macaw. No binaries to avoid repo bloat.


langston-barrett:
See here for output: https://github.com/GaloisInc/grease/actions/runs/23919880901?pr=628

RyanGlScott (approved):

PR #630, opened 2026-04-02
Fix DFS bug

langston-barrett:


RyanGlScott (approved):
If I'm reading the fix correctly, the bug was also present in the `bfs` path strategy---worth testing `bfs` as well?

langston-barrett:
Good call, done.

PR #633, opened 2026-04-02
Support reachability analysis in GREASE

langston-barrett:
Moved over from Screach for more sharing.

PR #640, opened 2026-04-11
Share more setup code between GREASE and Screach

langston-barrett:
- Factor out `RefineInputs` to shrink argument lists
- Add `setupRefinement` to set up error maps, initial fs, etc.
- Add `RefinementSetup` as the output of the above, fewer and smaller continuations

`setupRefinement` is also exactly the abstraction that will be needed for part of #633, specifically verification of reachability results.

### DIFF
--- a/screach/src/Screach.hs
+++ b/screach/src/Screach.hs
@@ -994,68 +994,88 @@ initCFG (CCC.SomeCFG entryRegSsaCfg) mbLoadedElf =
         let inputs =
               GR.RefinementInputs
                 { GR.refineInputArgNames = argNames
+                , GR.refineInputArgTypes = regTypes
                 , GR.refineInputArgShapes = argShapes
                 , GR.refineInputHeuristics = heuristics
                 , GR.refineInputSolver = Conf.solver conf
                 }
-        GR.setupRefinement
-          gla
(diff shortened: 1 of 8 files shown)


screach/src/Screach.hs
```diff
@@ -988,163 +976,157 @@ initCFG (CCC.SomeCFG entryRegSsaCfg) mbLoadedElf =
               mem
               memModelContents
               relocs
```
RyanGlScott: The large diff makes it hard to tell for sure, but I think this code now looks like:

```hs
          (mem0, ptrTable) <-
            GM.emptyMacawMem
              bak
              archCtx
              mem
              memModelContents
              relocs
          pure (mem0, ptrTable)
```

Which could be simplified to just:

```hs
            GM.emptyMacawMem
              bak
              archCtx
              mem
              memModelContents
              relocs
```

RyanGlScott (approved):

screach/src/Screach.hs
langston-barrett: Yeah, sorry for the huge diff 🙃 Hopefully this is a good step in the direction of simplifying these huge "main" modules.

langston-barrett: And good catch!

PR #642, opened 2026-04-12
screach: Move verification into its own module 

langston-barrett:
Prep for #633.

### DIFF
--- a/screach/src/Screach/Verify/Diagnostic.hs
+++ b/screach/src/Screach/Verify/Diagnostic.hs
@@ -11,16 +11,13 @@ import Grease.Diagnostic.Severity (Severity (Info, Warn))
 import Prettyprinter qualified as PP
 
 data Diagnostic
-  = -- | "Verifying reachability of result N / M"
-    VerifyReachable
+  = VerifyReachable
       -- | Total number of results
       Int
       -- | Current result number
       Int
-  | -- | "Verified reachability"
-    VerifySuccess
-  | -- | "Failed to verify reachability!"
-    VerifyFailure
+  | VerifySuccess
+  | VerifyFailure
   deriving (Show)
 
 instance PP.Pretty Diagnostic where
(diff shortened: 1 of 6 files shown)


screach/src/Screach/Verify/Diagnostic.hs
```diff
@@ -0,0 +1,43 @@
...
+data Diagnostic
+  = -- | "Verifying reachability of result N / M"
+    VerifyReachable
+      -- | Total number of results
+      Int
+      -- | Current result number
+      Int
+  | -- | "Verified reachability"
+    VerifySuccess
+  | -- | "Failed to verify reachability!"
```
RyanGlScott: Haddock nits:


```suggestion
  = -- | \"Verifying reachability of result N / M\"
    VerifyReachable
      -- | Total number of results
      Int
      -- | Current result number
      Int
  | -- | \"Verified reachability\"
    VerifySuccess
  | -- | \"Failed to verify reachability!\"
```

screach/src/Screach/Verify/Diagnostic.hs
```diff
@@ -0,0 +1,43 @@
...
+import Grease.Diagnostic.Severity (Severity (Info, Warn))
+import Prettyprinter qualified as PP
+
+data Diagnostic
+  = -- | "Verifying reachability of result N / M"
+    VerifyReachable
+      -- | Total number of results
+      Int
+      -- | Current result number
+      Int
```
RyanGlScott: This says `N / M`, but is the first `Int` `N` and the second `Int` `M`, or is it the other way around? Worth clarifying (e.g., say `-- | Total number of results (N)` or `-- | Total number of results (M)`).

RyanGlScott (approved):

langston-barrett:
Fixed some of the Haddock nits by removing the docs - there's not really a need for them given that the instance is right there. The fields are still documented.

PR #646, opened 2026-04-14
chore(submodules): Bump submodules for `microlens` changes

langston-barrett:
For faster builds 🤞

### DIFF
--- a/screach/src/Screach/Verify.hs
+++ b/screach/src/Screach/Verify.hs
@@ -64,7 +64,7 @@ verifyReachable la gla bak initShape execFeats cDataList =
     doLog la (Diag.VerifyReachable (length cDataList) no)
     let concArgShapes = Conc.concArgsShapes (Conc.concArgs cData)
         untagArgs = TFC.fmapFC (TFC.fmapFC (const Shape.NoTag)) concArgShapes
-    -- TODO(#640): Incorporate the concretized filesystem.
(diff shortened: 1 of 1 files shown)


RyanGlScott (commented):
Do you plan on actually replacing the `lens` dependencies in `grease`/`screach` with `microlens` at some point? (This bumps the submodules, but as far as I can tell, the `lens` dependency still remains.)

langston-barrett:
> Do you plan on actually replacing the lens dependencies in grease/screach with microlens at some point?

Yes, I'd like to! I would first need to do so along the What4 -> Crucible -> macaw-symbolic dependency chain. Currently I haven't seen any blocker to doing so, and I think the same motivations apply to these codebases as to p-u, elf-edit, etc. GHC 9.14 has been a confirmation of my hypothesis that shrinking our dependency tree will give us more freedom in supporting a wider range of GHCs, as many dependencies of `lens` are not GHC 9.14-ready even four months after its release.

> (This bumps the submodules, but as far as I can tell, the lens dependency still remains.)

Indeed, this PR will make the build graph for GREASE "wider" by allowing the compilation of elf-edit, flexdis86, macaw-base, and macaw-x86 to proceed before `lens` and all of its dependencies are done, but it doesn't *shrink* the graph at all.

It also just serves as a confirmation (via GREASE's test suite) that there are no behavioral changes from the microlens transitions in the deps.

RyanGlScott (approved):

PR #656, opened 2026-04-23
feat(overrides): `@write-byte-vec`

### LINKED ISSUE #654: override: `write-fresh-bytes`
One of the primary purposes of `@fresh-bytes` is to override functions like `recv` or `read`. Most such functions take a buffer as an input, thus requiring users to loop over the vector returned by `@fresh-bytes` and write in the bytes one-by-one, as seen here:

https://github.com/GaloisInc/grease/blob/cfaea05ad9d93de632cf7efc8ee9ae2a7f7c5a26/grease-exe/tests/x86/extra/recv.x64.cbl#L1-L34

This is cumbersome. We should have a `@write-fresh-bytes` that does the write-loop for users.

langston-barrett:
Fixes #654. Especially useful for using `@fresh-bytes` with a low loop bound.

### DIFF
--- a/doc/sexp.md
+++ b/doc/sexp.md
@@ -103,13 +103,15 @@ These overrides check if all elements of the vector have unique concrete values
 across all satisfying models. If all elements are uniquely determined, returns
 the concrete vector; otherwise returns the symbolic vector unchanged.
 
-### `@fresh-bytes`
+### `@fresh-bytes` and `@write-byte-vec`
 
-The bytes created using `fresh-bytes` will be concretized and printed to the
-terminal if GREASE finds a possible bug.
+The bytes created using `fresh-bytes` will be concretized and printed to
+the terminal if GREASE finds a possible bug. In contrast to `@write-bytes`,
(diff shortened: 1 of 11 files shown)


doc/sexp.md
```diff
@@ -103,45 +103,29 @@ These overrides check if all elements of the vector have unique concrete values
 across all satisfying models. If all elements are uniquely determined, returns
 the concrete vector; otherwise returns the symbolic vector unchanged.
 
-### `@fresh-bytes`
+### `@fresh-bytes` and `@write-byte-vec`
 
 The bytes created using `fresh-bytes` will be concretized and printed to the
 terminal if GREASE finds a possible bug.
```
RyanGlScott: I was hoping to see a little more detail on what `write-byte-vec` does. In particular, it would be worth explaining the distinction between `write-byte-vec` and `write-bytes`.

RyanGlScott (approved):

doc/sexp.md
langston-barrett: Added!

PR #661, opened 2026-04-27
tests: Embed CFLAGS in C files for binary tests, too

langston-barrett:
We already did this for LLVM bitcode tests, it's a nice way to reduce the number of tests that require custom Make rules, and to make tests a bit more self-contained.


langston-barrett:
Closing until I have time to revisit.

PR #671, opened 2026-04-29
grease-llvm: Split out LLVM-related functionality into a package

### LINKED ISSUE #665: refactor: Split out `grease-llvm` for LLVM analysis
Currently, the grease library depends on llvm-pretty-bc-parser and crucible-llvm-syntax:

https://github.com/GaloisInc/grease/blob/f42b31c05d9013aa4aba4153d92bf96287b32cfc/grease/grease.cabal#L183

https://github.com/GaloisInc/grease/blob/f42b31c05d9013aa4aba4153d92bf96287b32cfc/grease/grease.cabal#L177

This means that those dependencies must be compiled before starting on the grease library, and before starting on the grease-ISA packages, even though the latter have no real dependency on the former. We should have a new grease-llvm package for LLVM-specific concerns (overrides, debug info parsing, etc.).

In diagram form, we have:

Image

but want:

Image

langston-barrett:
Fixes #665.


RyanGlScott (approved):
A couple of additional places that should be updated:

https://github.com/GaloisInc/grease/blob/f7d99cae0a09494d67968a938f82ea254b2f3c3c/.github/workflows/ci.yml#L224-L233
https://github.com/GaloisInc/grease/blob/f7d99cae0a09494d67968a938f82ea254b2f3c3c/doc/dev/dev.md#L87-L99

langston-barrett:
Good catch, thanks!

PR #674, opened 2026-05-01
screach: Docker build

### LINKED ISSUE #673: docker: Add screach to image

langston-barrett:
Fixes #673


langston-barrett:
Docker build in CI: https://github.com/GaloisInc/grease/actions/runs/25217491953

.github/workflows/docker.yml
```diff
@@ -10,20 +10,22 @@ on:
...
     strategy:
       fail-fast: false
       matrix:
         # Cover both x86-64 and ARM64.
         os: [ubuntu-24.04, ubuntu-24.04-arm]
+        image:
+          - name: grease
+            repo: ghcr.io/galoisinc/grease
+          - name: screach
+            repo: ghcr.io/galoisinc/screach
```
RyanGlScott, line 28: Am I understanding correctly that this will build the `grease` and `screach` targets in separate, parallel jobs? If so, I worry about wasted work here, since the two jobs will share a lot of the same build dependencies. I wonder if it would be better to build both targets in the same job and then push each of the resulting images at the end of the job.

screach/doc/install.md
```diff
@@ -1,8 +1,38 @@
...
+First, fetch the source:
+```sh
+git clone https://github.com/GaloisInc/grease
+cd grease
+git submodule update --init
+```
+Then, build the image:
+```sh
+docker build --target screach . --tag screach:latest
+```
```
RyanGlScott, line 34: A minor thing, but it might be nice to add some newlines around the code blocks here (and elsewhere). Some parts of the docs do this and others don't, but I find it nice to consistently separate code blocks for newlines to make the source easier to read.

.github/workflows/docker.yml
langston-barrett: That's correct. I might leave this for a follow-up issue if that's alright. I need these Docker containers by Friday and I'm working on some Macaw issues for the same deadline.

RyanGlScott: No objections from me!

RyanGlScott (approved):

.github/workflows/docker.yml
langston-barrett: https://github.com/GaloisInc/grease/issues/678

PR #675, opened 2026-05-01
Resolve unconstrained pointers

### LINKED ISSUE #657: macaw: Improve behavior on unconstrained loads
Unconstrained loads (e.g., through initial GREASE register states) can cause the entire binary to be loaded into the SMT array, making any subsequent SMT query (e.g., path-sat) unworkable for large binaries. We should warn when performing a Macaw load through a global pointer with an unconstrained interval abstract value, at least. Probably, we should use Macaw's pointer resolution at that point and hope the path condition narrows the range.

langston-barrett:
Fixes #657

### DIFF
--- a/grease/src/Grease/Macaw.hs
+++ b/grease/src/Grease/Macaw.hs
@@ -32,6 +32,7 @@ import Data.Macaw.Memory qualified as MM
 import Data.Macaw.Memory.ElfLoader qualified as EL
 import Data.Macaw.Symbolic qualified as Symbolic
 import Data.Macaw.Symbolic.Backend qualified as Symbolic
+import Data.Macaw.Symbolic.Concretize qualified as MSC
 import Data.Macaw.Symbolic.Memory qualified as MSM
 import Data.Macaw.Symbolic.Memory.Lazy qualified as Symbolic
 import Data.Macaw.Types qualified as MT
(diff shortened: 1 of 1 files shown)


langston-barrett:
huh:
```
    fig2-sdse:                   FAIL
      Exception: 
      Check failed! Passing checks:
      
      Failing check:
      ❌ no match at <out>:1:1:
        Searching for target function 'vuln' [address 0x10002530]
        Executing frame: ProgramLoc {plFunction = go, plSourceLoc = 0x10002550} running state
        Executing frame: ProgramLoc {plFunction = go, plSourceLoc = 0x10002550} ct transfer
        
        ...
      
      stack trace:
        test-data/fig2-sdse.c:4 in reached
        test-data/fig2-sdse.c:52 in <main>
      
      
      Use -p '/fig2-sdse/' to rerun this test only.
    tgt-ov-interesting:          FAIL
      Exception: 
      Check failed! Passing checks:
      ✔️ match at <out>:1819:1-1819:34:
        Reached target function 'reached'
      
      stack trace:
        tgt-ov-interesting.lua:4 in reached
        tgt-ov-interesting.lua:49 in <main>
      
      
      
      Failing check:
      ❌ no match at <out>:1819:34:
          (unsat)
        Executing frame: ProgramLoc {plFunction = go, plSourceLoc = 0x100025c0} running state
        Executing frame: ProgramLoc {plFunction = go, plSourceLoc = 0x100025c0} ct transfer
```

langston-barrett:
Ah, I think that this fails due to timeouts from dereferencing pointers inside GREASE-synthesized arguments... Not sure how to handle that, needs more design work.

PR #687, opened 2026-05-12
screach: Use the concretized filesystem when verifying reachability

### LINKED ISSUE #639: reachability: Use the concretized filesystem when verifying reachability
We currently use the concretized argument shapes, but not the concretized files on the filesystem.

langston-barrett:
Fixes #639.

### DIFF
--- a/grease/src/Grease/Refine.hs
+++ b/grease/src/Grease/Refine.hs
@@ -571,7 +571,10 @@ setupRefinement ::
   , Cursor.CursorExt ext ~ PtrCursor.Dereference ext wptr
   ) =>
   GreaseLogAction ->
-  Opts.FsOpts ->
+  -- | Filesystem source: use 'Left' to seed from a 'Conc.ConcFs' (e.g., for
+  -- verification replay), or 'Right' to initialize a fresh filesystem from
+  -- 'Opts.FsOpts' (normal analysis).
+  Either Conc.ConcFs Opts.FsOpts ->
   C.Timeout ->
   C.HandleAllocator ->
   bak ->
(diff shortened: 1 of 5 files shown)


grease/src/Grease/Refine.hs
```diff
@@ -572,14 +572,17 @@ setupRefinement ::
   ) =>
   GreaseLogAction ->
   Opts.FsOpts ->
+  -- | Concretized filesystem to use instead of initializing a fresh one from
+  -- 'FsOpts'. Pass 'Nothing' to use 'FsOpts' as normal.
+  Maybe Conc.ConcFs ->
```
RyanGlScott: It's a bit weird that the filesystem configuration is split between two separate arguments (`FsOpts` and `Maybe ConcFs`) like this. Would it make sense to combine them (e.g., `Either ConcFs FsOpts`)?

screach/src/Screach.hs
```diff
@@ -939,6 +939,9 @@ initCFG ::
   GM.SetupHook sym arch ->
   GMSH.ExecutingAddressAction arch ->
   GMOA.AddressOverrides arch ->
+  -- | Concretized filesystem to use instead of initializing a fresh one from
+  -- 'FsOpts'. Pass 'Nothing' to use 'FsOpts' as normal.
```
RyanGlScott: This says "Pass `Nothing` to use `FsOpts` as normal", but I'm not sure where the `FsOpts` materialize in this type signature.

RyanGlScott (approved):

grease/src/Grease/Refine.hs
langston-barrett: Good idea, done!

screach/src/Screach.hs
langston-barrett: Clarified!

PR #693, opened 2026-05-13
Lint with `tagref`

### LINKED ISSUE #567: Use tagref
I've found this (or something like it) quite useful on my other projects for keeping documentation up-to-date. https://github.com/stepchowfun/tagref

langston-barrett:
`tagref` is a tool to check cross-references in code. It can be thought of as a machine-checkable generalization of our GHC-inspired `Note [Note name]` system of references.

This commit introduces a few example uses, we would expect to add many more in the course of development.

Fixes #567.

### DIFF
--- a/.github/ci.sh
+++ b/.github/ci.sh
@@ -10,7 +10,7 @@ BIN=${PWD}/bin
 is_exe() { [[ -x "$1/$2$EXT" ]] || command -v "$2" > /dev/null 2>&1; }
 
 # Create a grease.buildinfo.json file for the benefit of the Docker image.
-# (See Note [grease.buildinfo.json] in grease-cli/src/Grease/Version.hs.)
+# (See [ref:grease_buildinfo_json].)
 #
 # The first argument is the git commit, and the second argument is the git
 # branch name.
--- a/doc/dev/lint.md
+++ b/doc/dev/lint.md
@@ -80,6 +80,34 @@ git ls-files -z --exclude-standard '*.py' | xargs -0 ruff check
 
 See the [Ghidra batch plugin docs](../ghidra-batch-plugin.md).
 
+## tagref
+
+We lint cross-references in code with [tagref].
+
+```sh
+tagref
+```
+
+tagref verifies that every reference refers to an existing tag, and that there
+are no duplicate tags. The following is the syntax of tags and references:
+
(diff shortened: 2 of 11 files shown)


doc/dev/lint.md
```diff
@@ -80,6 +80,34 @@ git ls-files -z --exclude-standard '*.py' | xargs -0 ruff check
...
+## tagref
+
+We lint cross-references in code with [tagref].
+
+```sh
+tagref
+```
+
+tagref verifies that every reference refers to an existing tag, and that there
+are no duplicate tags. The following is the syntax of tags and references:
```
RyanGlScott, line 92: Out of curiosity, does it also check that each tag has at least one reference somewhere in the repo?

.github/ci.sh
```diff
@@ -10,7 +10,7 @@ BIN=${PWD}/bin
 is_exe() { [[ -x "$1/$2$EXT" ]] || command -v "$2" > /dev/null 2>&1; }
 
 # Create a grease.buildinfo.json file for the benefit of the Docker image.
-# (See Note [grease.buildinfo.json] in grease-cli/src/Grease/Version.hs.)
+# (See [ref:grease_buildinfo_json].)
```
RyanGlScott, line 13: A minor thing, but I think I had implicitly assumed the convention that if you read "see Note [...]" somewhere in a file, then that Note would be defined somewhere in the same file unless otherwise stated (such as in this example, where the Note was defined in a different file). This convention is kind of nice because it allows you to easily find most Notes defined in the same file (just use your text editor's search feature).

After moving to `tagref`, we appear not to use this convention anymore. I suppose I can live with that, though. There might even be a way to get jump-to-ref functionality in one's text editor.

RyanGlScott (approved):

doc/dev/lint.md
langston-barrett: I don't think so, I was considering creating a feature request.

.github/ci.sh
langston-barrett: We could keep such a convention (e.g., by just appending a `file:` reference to non-local `ref:`s), but `git grep` is almost as quick as ctrl-f and I'd like to keep overhead low. Many editors also have "grep in project".

PR #696, opened 2026-05-21
deps: Bump macaw for abstract values in CFGs

langston-barrett:


langston-barrett:
This reduces memory usage by 15% for the Screach `fig2-sdse` test!

langston-barrett:
Running this on an internal demo found https://github.com/GaloisInc/macaw/issues/623

grease-cli/src/Grease/Cli.hs
```diff
@@ -169,6 +169,11 @@ boundsOptsParser = Opt.parserOptionGroup "Bounds, limits, and timeouts" $ do
             <> Opt.showDefault
             <> Opt.value GO.defaultSolverTimeout
         )
+  simCheckAbsValues <-
+    Opt.switch
+      ( Opt.long "check-abs-values"
+          <> Opt.help "Perform additional safety checks on static analysis soundness"
```
RyanGlScott: Do you think it is worth adding some developer documentation to describe what exactly this is checking for? It's not very clear what this is doing without further context.

grease/src/Grease/Macaw.hs
```diff
@@ -527,6 +527,9 @@ initState ::
   ArchContext arch ->
   SetupHook sym arch ->
   AddressOverrides arch ->
+  -- | Add proof obligations for narrowing of abstract values (i.e., use
+  -- 'Data.Macaw.Symbolic.MemOps.narrowBVDomainChecked')
+  Bool ->
```
RyanGlScott: We pass around this `Bool` a lot and give it the same Haddocks everywhere, so I wonder if it would be worth making this into a newtype to offer a place to consolidate the Haddocks. (See existing newtypes in `grease:Grease.Options` for prior art.)

RyanGlScott (approved):

langston-barrett:
Waiting on https://github.com/GaloisInc/macaw/pull/624

grease/src/Grease/Macaw.hs
langston-barrett: Good point.

grease-cli/src/Grease/Cli.hs
langston-barrett: Documented in the Haddock for the newtype!

grease/src/Grease/Macaw.hs
langston-barrett: Done!

RyanGlScott (approved):

Issue #717, opened 2026-06-23
feat(ghidra): Test interoperability of Screach with ECFS Ghidra plug-in


langston-barrett:

langston-barrett:
Done in #721.

PR #721, opened 2026-06-30
test(ghidra): Ensure that Screach and ECFS plugins can be used together

### LINKED ISSUE #651: ecfs: Allow specifying shared library functions in ECFS files as targets
`screach` features a `--target-symbol` flag for specifying the name of a function in a binary as a target, but at the moment, it's of limited utility for ECFS files. This is because `--target-symbol` works by looking up function symbols in the binary's `.symtab` and `.dynsym` sections, which is only a subset of the functions contained in an ECFS file. For instance, you might have an ECFS snapshot from a binary which crashed when executing `abort()` in a shared library, but the function symbol for `abort` wouldn't appear directly in the ECFS file's `.dynsym` section.

There are plans to augment ECFS files with additional, per-shared-library `.dynsym` sections that contain each library's function symbols. With this information, we should be able to extend `--target-symbol` to consider _all_ of an ECFS file's symbol tables, which would finally allow us to use `--target-symbol` to specify shared library functions as targets.

When implementing this, we will need to come up with a story for how to handle the case where a single function name is defined in multiple dynamic symbol tables.

langston-barrett:
Fixes #651, #717.

Screenshot 2026-06-30 at 11 10 02 AM

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -209,8 +209,12 @@ jobs:
 
     - name: Build and test Screach Ghidra plugin
       # Reuse the checked-in Gradle wrapper from ghidra-plugin (the screach
-      # plugin has no wrapper of its own).
-      run: ./ghidra-plugin/gradlew -p screach/ghidra buildExtension test
+      # plugin has no wrapper of its own). Pass the screach binary path so the
+      # integration test can find it without relying on PATH.
+      run: |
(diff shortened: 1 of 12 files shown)


thebendavis (approved):
I installed the updates here and successfully reproduced the working screach-on-ECFS example in my Ghidra environment. LGTM, thanks!

PR #723, opened 2026-07-01
test(elf-edit-ecfs): Golden tests vs. upstream `readecfs`

### LINKED ISSUE #722: tests: elf-edit-ecfs
This package should have a test suite. This should consist of golden tests for `readecfs`, compared (at least at generation time) to upstream tooling from ee-ecfs as an oracle.

langston-barrett:
- Add golden tests for `elf-edit-ecfs`'s `readecfs`
- Add a mode where the output is compare to upstream `readecfs` (not run by default in CI, I ran locally)
- Add parsing for a few ECFS features necessary to get the tests passing

The tests perform some normalization (stripping NUL, trailing whitespace.)

Fixes #722.


elf-edit-ecfs/tools/ReadEcfs.hs
```diff
@@ -139,6 +140,34 @@ printRegisters prStatus =
...
+    Left _ -> Nothing
+    Right (_, _, r) -> Just r
+ where
+  getSigInfo :: Get.Get (Int, Word64)
+  getSigInfo = do
+    signo <- fromIntegral <$> Get.getInt32le
+    Get.skip 8 -- si_errno + si_code
+    Get.skip 4 -- padding before si_addr
+    addr <- Get.getWord64le
+    pure (signo, addr)
```
RyanGlScott: Would it make sense to put these `decode*` functions in `Data.ElfEdit.Ecfs` alongside its existing `decode*` functionality?

elf-edit-ecfs/test/golden/readecfs-auxv.out
```diff
@@ -0,0 +1,16 @@
+
+- readecfs output for file ../ghidra-ecfs/src/test/resources/ecfs/test.x64.elf
```
RyanGlScott, line 2: As I understand it, this isn't just the output of calling `readecfs` directly on the binary, but rather something like `readelf -A`. Can you include the relevant flags in these comments so that it will be easier to regenerate these files in the future should `readecfs`'s output change?

RyanGlScott (approved):

elf-edit-ecfs/test/golden/readecfs-auxv.out
langston-barrett: The test driver regenerates them automatically with `--accept`! The flag -> test correspondence is recorded in `cases` in `Golden.hs`.

elf-edit-ecfs/tools/ReadEcfs.hs
langston-barrett: Good call.

PR #730, opened 2026-07-06
ci: Lint with zizmor

langston-barrett:
[zizmor](http://zizmor.sh/) is a static analysis tool for GitHub Actions workflows, focused on security. It is widely used by major OSS projects, we should use it too to shrink our attack surface and secure the artifacts (e.g., Docker containers) we build and ship from CI.


.github/actions/ghidra-install/action.yml
```diff
@@ -1,14 +1,19 @@
...
+outputs:
+  ghidra-install-dir:
+    description: "Path to the extracted Ghidra installation"
+    value: ${{ steps.fetch.outputs.ghidra-install-dir }}
 runs:
   using: "composite"
   steps:
     - name: Setup java
-      uses: actions/setup-java@v4
+      uses: actions/setup-java@c1e323688fd81a25caa38c78aa6df2d33d3e20d9 # v4
```
langston-barrett, line 10: Note for reviewers: zizmor recommends pinning SHAs directly because tags can be modified. Dependabot understands these SHAs and applies the same policies to them that it would do the version number. It will keep them as SHAs when bumping.

https://docs.zizmor.sh/audits/#unpinned-uses

.github/actions/ghidra-install/action.yml
```diff
@@ -1,14 +1,19 @@
 name: "Install Ghidra"
+outputs:
+  ghidra-install-dir:
+    description: "Path to the extracted Ghidra installation"
+    value: ${{ steps.fetch.outputs.ghidra-install-dir }}
```
langston-barrett, line 5: for this and related changes: https://docs.zizmor.sh/audits/#github-env

lun.toml
```diff
@@ -1,5 +1,15 @@
 ignore = ["./deps/**/*", "*.bat", "*.bc", "*.elf", "*.exe"]
 
+# ---------------------------------------------------------
+# github actions
+
+[[tool]]
+name = "zizmor"
+# we don't have a zizmor.yml/zizmor.yaml config file
+configs = []
+# the default only covers *.yml under .github/, but our composite action uses .yaml
```
RyanGlScott: Oh dear, I didn't realize that the choice of file suffix would become load-bearing at some point. I wonder if we should just standardize on the use of either `.yml` or `.yaml` and use it consistently.

.github/actions/ghidra-install/action.yml
RyanGlScott: Does this mean that Dependabot will submit a PR to bump the action's SHA ref any time a commit is pushed upstream? That seems like a lot of additional churn...

langston-barrett: Good question, but no! Basically Dependabot will continue working exactly as it does not, but it will read and writes SHAs instead of tags. So it will only update to upstream *releases* not arbitrary upstream commits.

RyanGlScott: > So it will only update to upstream _releases_ not arbitrary upstream commits.

If that is the case, then I'm not sure I understand the rationale for using SHA refs instead of tag refs. If a bad actor has the ability to push a malicious release, then Dependabot will still recommend updating to it, regardless of whether the malicious release is referenced with a SHA or a tag. What am I missing?

langston-barrett: One case it protects against is attackers who have obtained force-push privleges but not release-creating privleges. That is likely a narrow gap, but it is a gap nonetheless. Another aspect of this has to do with [cooldowns](https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns): you can configure Dependabot to not suggest *immediate* upgrades in case the project was just compromised, in the hopes that such compromises will be recognized and mitigated in a relatively short period of time. There's no analogous cooldown for a force-push'd tag. Finally, there is an explicit/implicit benefit - you can at least track which of your projects *were* affected by a hacked action if the upgrade is tracked in git history, it's much harder to do so if the version was implicitly pinned by a movable tag.

A non-security benefit is hermeticity: even a trusted developer can accidentally push a patch release that unintentionally breaks downstream consumers. It's nice to pin exact, immovable references in CI for this reason (this is why we use `cabal.freeze` files in some repos too, instead of just Cabal version bounds).

RyanGlScott: Your point about force-pushing tags is a good one. It's a bit unfortunate that we've been spending so much effort to use tags everywhere only to have the recommended best practices change to using SHA refs... oh well.

RyanGlScott (approved):

.github/actions/ghidra-install/action.yml
langston-barrett: Yeah, this is an evolving area too. Just last year GitHub announced [immutable releases](https://github.blog/changelog/2025-10-28-immutable-releases-are-now-generally-available/), which I saw the `ruff` action adopted recently. GitHub Actions security is... a work-in-progress.

lun.toml
langston-barrett: Standardized to .yml!

.github/actions/ghidra-install/action.yml
langston-barrett: GitHub Actions is also supposed to add some native support for this kind of thing in 2026: https://github.blog/news-insights/product-news/whats-coming-to-our-github-actions-2026-security-roadmap/

PR #735, opened 2026-07-15
chore(submodules): Bump macaw for LLVM jump table section parsing

langston-barrett:


grease/src/Grease/Macaw/Load.hs
```diff
@@ -83,6 +84,11 @@ data BinMd arch
   -- their corresponding function addresses.
   , binRelocs :: Map.Map (MM.MemWord (MC.ArchAddrWidth arch)) (Arch.ArchReloc arch)
   -- ^ Map of relocation addresses to their types.
+  , binLLVMJumpTableSizes ::
+      Map.Map
+        (MC.ArchSegmentOff arch)
+        (JT.JumpTableSize (MC.ArchAddrWidth arch))
+  -- ^ Map of LLVM-recorded jump-table addresses to their entry counts.
```
RyanGlScott, line 91: It might be worth noting here that `grease`'s `load` function doesn't fill in this information, and that you have to compute this separately.

RyanGlScott (approved):

grease/src/Grease/Macaw/Load.hs
langston-barrett: Oh whoops! I accidentally pushed this commit to this PR when I intended to put it on a new branch here: #737. Will do this there.

langston-barrett:
Closing in favor of #737.

PR #737, opened 2026-07-21
feat(x86): Support PIE with `-emit-jump-table-sizes-section`

langston-barrett:
...via parsing `R_X86_64_RELATIVE` specifically rather than waiting for a general refactor of Macaw APIs for relocation handling as suggested in https://github.com/GaloisInc/macaw/issues/629#issuecomment-4982689067.


langston-barrett:
Ah, Zizmor helpfully catches that this SHA isn't pointed to by the tag in the comment. That's likely because the tag was moved. We'll need to use full version numbers in these comments:
```
warning[ref-version-mismatch]: action's hash pin has mismatched or missing version comment
  --> ./.github/workflows/lint.yml:25:75
   |
25 |       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
   |         ---------------------------------------------------------------   ^^ points to commit 3d3c42e5aac5
   |         |
   |         is pointed to by tag v7.0.0
   |
   = note: audit confidence → High
   = note: this finding has an auto-fix
   = help: audit documentation → https://docs.zizmor.sh/audits/#ref-version-mismatch
```

RyanGlScott (approved):

